PR #4, opened 2022-02-09
[monorepo] fix: update git submodule

Wayonb:
Update the git submodule to use https instead ssh so that an ssh key is not required to clone.


Jaguar0625 (approved):

Jaguar0625 (approved):

Jaguar0625:
Merged in f556545ccf110adf7248afd89027bbef7236b479 .

PR #24, opened 2022-03-04
[optin/puller] task: populate multisig db

Jaguar0625:


optin/puller/workflows/get_account_states.py
```diff
@@ -19,6 +20,40 @@ def get_addresses(filename):
...
+async def populate_multisig(connection, client, addresses):
+	database = MultisigDatabase(connection)
+	database.create_tables()
+
+	limiter = AsyncLimiter(20, 1.0)
+
+	async def get_state(address):
+		async with limiter:
+			account = await client.account(address)
+			database.insert_if_multisig(account)
```
gimre-xymcity, line 49: hmm, this gets _current_ account state, so is this ok?

Jaguar0625: we do not have historical multisigs, so i guess so

gimre-xymcity (dismissed):

PR #25, opened 2022-03-04
 [optin/puller] task: check for already post opted in accounts

Jaguar0625:


optin/puller/workflows/download_postoptin.py
```diff
@@ -29,14 +31,41 @@ def parse_args():
...
+		print(f'{transaction_height} ERROR: {process_result.message}')
+		database.add_error(process_result)
+	else:
+		destination_address = symbol_network.public_key_to_address(process_result.destination_public_key)
+		optin_transaction_infos = await symbol_client.find_optin_transactions(PublicKey(OPTIN_SIGNER_PUBLIC_KEY), destination_address)
+		optin_transaction_info = next((info for info in optin_transaction_infos if process_result.address == info.address), None)
+		if optin_transaction_info:
+			print(f'{transaction_height} SUCCESS (DUPLICATE)')
+			database.add_request(process_result)
+			database.set_request_status(process_result, OptinRequestStatus.DUPLICATE, optin_transaction_info.transaction_hash)
```
Jaguar0625, line 54: not sure if we should add this into completed too, but a bit tricky in that we don't have balances here (balance seeding runs AFTER this script)

gimre-xymcity (approved):

PR #35, opened 2022-03-23
[optin/puller] task: make processing interactive

Jaguar0625:


optin/puller/workflows/payout.py
```diff
@@ -110,40 +73,60 @@ async def send_funds(self, destination_address, amount, deadline, nem_address):
...
+	for address, request_group in unprocessed_requests.items():
+		print()
+
+		processor.is_dry_run = True
+		await processor.process_unprocessed_request(address, request_group, sent_requests, deadline)
+
+		if is_dry_run:
+			continue
+
+		user_input = input('Press Y to send transaction and update databases: ')
```
gimre-xymcity, line 88: hmm, would maybe just mention, it's ok to Ctrl+C here

gimre-xymcity (approved):

optin/puller/workflows/payout.py
Jaguar0625: ctrl+c or any other input (i.e. n) has same effect. ;)

PR #54, opened 2022-04-06
[optin/puller]: completed db changes

gimre-xymcity:
* add postoptin flag - in `optin_id` table (per operation) and label
 * add `nem_transactions` table (given that there are multiple txes per address not sure if it should be singular `nem_transaction` or not)
 * add symbol transaction hash and height (since there can be only one, it's directly in `symbol_destination` table)
 * finally, add `--post` flag to `workflows/populate`
 
 **NOTE** this change does not contain "special" flag - imo if there is no transactions in `nem_transactions` it should be treated as special already...


optin/puller/tests/db/test_CompletedOptinDatabase.py
```diff
@@ -229,9 +337,16 @@ def _create_database_with_json_mappings(connection):
 			entry_2
 		]
 
+		if remove_nem_transactions:
+			for entry in json_data:
+				if 'source' not in entry:
+					continue
+				for source in entry['source']:
```
Jaguar0625, line 344: entry.get('source', []) ?

Jaguar0625 (dismissed):

PR #55, opened 2022-04-06
[optin/puller]: sym optin transactions -> payout transactions

gimre-xymcity:


optin/puller/puller/client/SymbolClient.py
```diff
@@ -84,7 +84,7 @@ async def find_payout_transactions(self, optin_signer_public_key, start_id=None)
 		transactions = await self.get(url_path, 'data')
 		return transactions
 
-	async def find_optin_transactions(self, optin_signer_public_key, address):
+	async def find_payout_transaction(self, optin_signer_public_key, address):
 		"""Finds optin transactions sent to the specified address."""
```
Jaguar0625, line 88: 1. it is returning array, so shouldn't it be plural?
2. update doc?

Jaguar0625 (approved):

PR #71, opened 2022-04-19
[optin/reporting/client] fix: table row height

OlegMakarenko:
### What was the issue?
- Single table row fills the table body height.
image

### What's the fix?
- Keep row height depending on its content.
image

### DIFF
--- a/optin/reporting/client/src/styles/globals.scss
+++ b/optin/reporting/client/src/styles/globals.scss
@@ -46,6 +46,10 @@ p-tabview,
   z-index: 10 !important;
 }
 
+tr {
+  height: 0;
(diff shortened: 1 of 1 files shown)


yilmazbahadir:
Can you try to put `NBVXTM3RDIY2KTGIASAAAV2WZYIQOD4BVMOVKLM7` in the `In Progress` tab search? The listing is okay but when the filter is cleared sizes go wrong (because of the loading div I guess, not sure we should show the `loading more items...` in such a case @cryptoBeliever ?)

image

yilmazbahadir:
already fixed by another commit in the branch

PR #75, opened 2022-04-22
[tools/vanity]: add to build and README

Jaguar0625:


.github/buildConfiguration.yaml
```diff
@@ -1,3 +1,7 @@
 builds:
   - name: Optin Puller
     path: optin/puller
+
+builds:
```
Wayonb, line 5: There should only be one builds property which have a list of projects.

Wayonb (changes requested):

gimre-xymcity (approved):

Wayonb (approved):

PR #76, opened 2022-04-24
[optin/reporting] task: unit test

AnthonyLaw:
### What was the issue?
- zero unit test in current project

### What's the fix?
- add unit test in reporting backend.
- added sinon, supertest dependencies
- added controller unit test
- added utils unit test
- added router unit test


optin/reporting/package.json
```diff
@@ -7,7 +7,7 @@
   "scripts": {
     "lint": "eslint .",
     "lint:fix": "eslint --fix .",
-    "test": "mocha --full-trace --recursive ./test",
+    "test": "mocha --full-trace --recursive ./test --exit",
```
yilmazbahadir, line 10: Why did we need to add the `--exit` option? Ideally we shouldn't need it.
Pls see this => https://stackoverflow.com/a/52143003/3190131

optin/reporting/test/utils/ServerUtils.spec.js
```diff
@@ -0,0 +1,170 @@
...
+			// Assert:
+			expect(result).to.be.equal('680423C0821C83F2F717A8F8F27B19CC174ED4D3A1DC3B0430');
+		});
+
+		it('return hex given string', () => {
+			// Arrange:
+			const byte = '1234';
+
+			// Act:
+			const result = Utils.byteToHexString(byte);
```
yilmazbahadir, line 62: does `byteToHexString`accept strings instead of byte arr? If so should we state it in the function name?

optin/reporting/package.json
AnthonyLaw: hmm, without `--exit` the terminal just hanging there.
I'm not sure what caused the problem, I will double-check on that.

optin/reporting/test/routers/api.spec.js
```diff
@@ -0,0 +1,117 @@
+const CompletedDB = require('../../models/completed');
+const OptinRequestsDB = require('../../models/optinRequests');
+const app = require('../../server');
+const TestUtils = require('../TestUtils');
+const { expect } = require('chai');
+const { stub, restore } = require('sinon');
+const request = require('supertest');
+
```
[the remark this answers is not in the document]
AnthonyLaw: How can I do a test with `2 pages`?
Because I mocking the data from data response.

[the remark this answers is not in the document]
yilmazbahadir: You can set the pageSize to half _perhaps_ ?

AnthonyLaw:
use `git cherry-pick` merge into `init-reporting`

PR #78, opened 2022-04-30
[optin/reporting] task: refactor query and validation

AnthonyLaw:
### What was the issue?
- current SQL query is using concatenation, to prevent SQL injection we need to use bind params.
- That is zero validation in req.query 

### What's the fix?
- rename req query `tz` to `timezone`
- refactor on the SQL query with bind params.
- added express-validator package
- added req query validation and error handling.


optin/reporting/controllers/optinRequestsController.js
```diff
@@ -57,15 +54,15 @@ const controller = {
 			}
 		};
 
-		const status = statusStringToInt(statusFilter.toLowerCase());
+		const statusFilter = statusStringToInt(status.toLowerCase());
```
Jaguar0625, line 57: why did you rename this? if it is single value (int), i think status name is clearer. "filter" typically connotates a function.

optin/reporting/utils/Validation.js
```diff
@@ -0,0 +1,664 @@
+const { query, validationResult } = require('express-validator');
+
+const validMomentTimezones = [
```
Jaguar0625, line 3: are these not defined in some package?

optin/reporting/utils/Validation.js
```diff
@@ -0,0 +1,664 @@
...
+
+const Validation = {
+	validate: method => {
+		const paginationValidation = [
+			query('pageSize', 'pageSize required').exists().isInt(),
+			query('pageNumber', 'pageNumber required').exists().isInt()
+		];
+
+		const sortValidation = [
+			query('sortBy', 'field name must be string')
```
Jaguar0625, line 606: should you limit possible sortBy values?

optin/reporting/utils/Validation.js
```diff
@@ -0,0 +1,664 @@
...
+
+		if (!errors.isEmpty()) {
+			res.status(422).json({ errors: errors.array() });
+			return;
+		}
+
+		next();
+	}
+};
+module.exports = Validation;
```
Jaguar0625, line 664: should you add some unit tests for this?

optin/reporting/utils/Validation.js
AnthonyLaw: Agree, i will add that 👍🏼

optin/reporting/utils/Validation.js
```diff
@@ -0,0 +1,664 @@
...
+			];
+		}
+
+		case 'exportCsv': {
+			return [
+				query('timezone', 'timezone must be string example America/Los_Angeles').exists().isIn(validMomentTimezones)
+			];
+		}
+
+		case 'getOptinRequests': {
```
yilmazbahadir, line 639: I'd move page-specific validations next to `Controller`s rather than a utility file. So `getCompleted` and `getOptinRequests` can go next to (or inside) the controllers and the rest would stay here as shared validationUtility, _if that makes sense_

optin/reporting/controllers/optinRequestsController.js
AnthonyLaw: I rename that because I did destructure params. I will undo this part.

optin/reporting/utils/Validation.js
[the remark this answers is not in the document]
AnthonyLaw: I will remove the list, it actually can find from the package `moment-timezone`.

optin/reporting/utils/Validation.js
AnthonyLaw: The reason I put all validation in 1 place, is because it will be easier to manage IMO.

Some of the validation rules are shared between `getCompleted` and `getOptinRequests`.
if we move to page-specific validations `Controllers`, we may have some duplicated code.

Jaguar0625 (approved):

PR #117, opened 2022-05-23
[optin/reporting] task: backend unit tests

AnthonyLaw:
### What was the issue?
- backend has low test coverage.

### What's the fix?
- minor refactor such as a naming convention, try-catch block, export missing method.
- added `mock` folder store `in_progress.db`, `completed.db`, add `current.version`
- ncrease backend unit test coverage to 100%


Jaguar0625:
please rebase onto dev 🙏

optin/reporting/test/routers/api.spec.js
```diff
@@ -170,4 +170,21 @@ describe('API Route', () => {
...
+		const randomPath = ['/test', '/api/completed/random', '/api/requests/random'];
+
+		it('returns html index page', async () => {
+			/* eslint-disable no-await-in-loop */
+			for (let i = 0; i <= randomPath.length - 1; ++i) {
+				// Act:
+				const response = await request(app).get(randomPath[i]);
+
+				// Assert:
+				expect(response.status).to.equal(200);
```
Jaguar0625, line 185: why is this 200 instead of 404 (not found) ?

optin/reporting/test/controllers/completedController.spec.js
```diff
@@ -63,100 +70,93 @@ describe('completed controller', () => {
 			expect(result.symbolBalance).to.be.eql(mockSymbolSources.map(symbolInfo => symbolInfo.balance));
 		};
 
+		const runBasicQueryTests = async overWriteParams => {
```
Jaguar0625, line 73: overwrite

optin/reporting/test/controllers/completedController.spec.js
```diff
@@ -63,100 +70,93 @@ describe('completed controller', () => {
...
-				address: '688ADC5D31F49F918AC71DCE18E7085944AEAB0AF60F8AD7',
-				balance: 79268194338335,
-				hashes: 'E49B240D76DAE7277089C2BDA66B297A05AE700361ED253ED48435E9AF9B0FE1',
-				height: 1,
-				timestamps: 1615853185
-			}];
-
-			await runBasicCompletedDataTests(mockNemSources, mockSymbolSources);
+			await runBasicCompletedDataTests({
+				mockNemSources: mockCompletedData[0].nemSource,
```
Jaguar0625, line 92: consider using enum or constant or object property instead of indexes as it will make tests easier to read

e.g. 
mockCompletedData.basic.nemSource
mockCompletedData.merge1.nemSource
mockCompletedData.merge2.nemSource
etc

optin/reporting/test/controllers/completedController.spec.js
```diff
@@ -63,100 +70,93 @@ describe('completed controller', () => {
...
+			});
+
+			getCompletedPaginationStub.returns(Promise.resolve([]));
+
+			// Act:
+			await CompletedController.getCompleted(req, res);
+
+			// Assert:
+			const result = res.json.getCall(0).firstArg;
+			expect(result).not.to.haveOwnProperty('error');
```
Jaguar0625, line 87: as you are only verifying there is no error, this is a weak test. it's probably ok for this, but you should use stricter asserts in future.

optin/reporting/test/controllers/optinRequestsController.spec.js
```diff
@@ -41,18 +71,74 @@ describe('optin requests controller', () => {
...
+			// Arrange:
+			const recordSize = 10;
+			const mockDb = TestUtils.mockInProgressDBRecord(1);
+
+			req = TestUtils.mockRequest({
+				...parameter,
+				nemAddress: 'naof6gii33dlok6cbr6shl6iu6gkvdzhbhvvyrfg'
+			});
+
+			await runBasicDataResponseTests(recordSize, mockDb);
```
Jaguar0625, line 105: missing 

// Act + Assert

(multiple tests)

optin/reporting/test/controllers/versionController.spec.js
```diff
@@ -0,0 +1,48 @@
...
+			await runBasicVersionTests({
+				configPath: '../mock',
+				expectedResult: {
+					versionId: '_ykRdiKlyoCZJAi4FPnptUcNL5FeNW9G',
+					lastUpdated: 'Tue, 10 May 2022 13:59:22 GMT'
+				}
+			});
+		});
+
+		it('return version id and last update with empty value if file not found', async () => {
```
Jaguar0625, line 38: is this expected or why is this not an error / exception?

optin/reporting/test/mock/current.version
```diff
@@ -0,0 +1,2 @@
+_ykRdiKlyoCZJAi4FPnptUcNL5FeNW9G
```
Jaguar0625, line 1: can you rename mock folder to resources as it's more consistent with naming elsewhere

optin/reporting/test/models/runTotalRecordTests.js
```diff
@@ -0,0 +1,13 @@
+const { expect } = require('chai');
```
Jaguar0625, line 1: since this is shared set of test(s), please move to test/test folder to not intermix with tests like we do in other places (e.g. sdk/javascript).

Jaguar0625: likewise should move similar run* files

optin/reporting/test/models/runDBNotFoundTests.js
```diff
@@ -0,0 +1,23 @@
...
+			// Arrange:
+			stub(Config, 'getDataStoragePath').returns(path.join(__dirname, '../path/notFound'));
+			refreshDBs();
+
+			await paginationFunc(parameters);
+		} catch (error) {
+			// Assert:
+			expect(error.message).to.be.equal('Database error :SQLITE_CANTOPEN: unable to open database file');
+		}
+	});
```
Jaguar0625, line 20: this test will pass if there is no exception. you can return from catch handler and have additional assert at end of function that always fails.

optin/reporting/test/models/database.spec.js
```diff
@@ -0,0 +1,28 @@
...
+		const { completed } = getDatabase();
+
+		runBasicDatabaseInstanceTests(completed, 'completed');
+	});
+
+	describe('in_progress db', () => {
+		// Arrange + Act:
+		const { in_progress } = getDatabase();
+
+		runBasicDatabaseInstanceTests(in_progress, 'in_progress');
```
Jaguar0625, line 26: since the Act is outside, this is not really running test but asserting, so consider renaming to something like assertOpenDatabaseProperties. additionally, it's more helpful to have assert here in test instead of helper.

optin/reporting/test/models/completed.spec.js
```diff
@@ -0,0 +1,204 @@
...
+const Config = require('../../config');
+const completedDB = require('../../models/completed');
+const { refreshDBs } = require('../../models/database');
+const { expect } = require('chai');
+const { stub, restore } = require('sinon');
+const path = require('path');
+
+const runBasicOptinTypeTests = (types, parameters) => {
+	/* eslint-disable no-await-in-loop */
+	for (let i = 0; i <= types.length - 1; ++i) {
```
Jaguar0625, line 12: can you use forEach here?

optin/reporting/test/models/completed.spec.js
```diff
@@ -0,0 +1,204 @@
...
+
+		it(`returns ${0 === optinType ? 'pre' : 'post'}-optin records`, async () => {
+			// Act:
+			const result = await completedDB.getCompletedPagination({
+				...parameters,
+				optinType
+			});
+
+			// Assert:
+			result.forEach(item => {
```
Jaguar0625, line 24: this is also weak assert, since you're only validating one of many item properties

optin/reporting/test/models/completed.spec.js
```diff
@@ -0,0 +1,204 @@
...
+					expect(source).to.have.property('height');
+					expect(source).to.have.property('timestamps');
+					expect(source).to.have.property('label');
+				});
+				item.symbolDestination.forEach(source => {
+					expect(source).to.have.property('address');
+					expect(source).to.have.property('balance');
+					expect(source).to.have.property('hashes');
+					expect(source).to.have.property('height');
+					expect(source).to.have.property('timestamps');
```
Jaguar0625, line 86: likewise weak because you're only checking property existence but not value

optin/reporting/test/models/completed.spec.js
```diff
@@ -0,0 +1,204 @@
...
+			};
+			const result = await completedDB.getCompletedPagination({
+				...parameters,
+				...sort
+			});
+
+			// Assert:
+			const firstTimestamps = result[0].id;
+			const lastTimestamps = result[result.length - 1].id;
+
```
Jaguar0625, line 105: for sorting tests, you should check predicate holds for ALL items in result, not just first and last

optin/reporting/test/models/completed.spec.js
```diff
@@ -0,0 +1,204 @@
...
+			const nemAddressHex = '680B140819D8B6C734091637DA4B3B854BC2FD30D3FE42B0BA';
+
+			// Act:
+			const result = await completedDB.getCompletedPagination({
+				...parameters,
+				nemAddressHex
+			});
+
+			// Assert:
+			expect(result[0].nemSource[0].address).to.be.equal(nemAddressHex);
```
Jaguar0625, line 160: you should also check the size of result and result[0].nemSource

optin/reporting/test/routers/api.spec.js
AnthonyLaw: yeah because I had added `*` in the route, so technically whatever route, it should hit the HTML page.

what do you think about this?

https://github.com/symbol/product/blob/1dc3f89bb5ae03300b7997cb824c4506b276cd89/optin/reporting/server.js#L20

optin/reporting/test/controllers/completedController.spec.js
AnthonyLaw: I see, that is good idea, i will update that 👍🏼

optin/reporting/test/controllers/completedController.spec.js
```diff
@@ -63,100 +70,93 @@ describe('completed controller', () => {
...
+			getCompletedPaginationStub.throws(new Error('database error'));
+
+			// Act:
+			await CompletedController.getCompleted(req, res);
+
+			// Assert:
+			const { data, error } = res.json.getCall(0).firstArg;
+
+			expect(data).to.be.eql([]);
+			expect(error).to.be.equal('database error');
```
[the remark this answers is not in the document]
AnthonyLaw: what do you mean missing eql and equal ?

`eql` is compared with `array` or `object`.
`equal` is compared with `string` or `number`

optin/reporting/test/controllers/versionController.spec.js
AnthonyLaw: it's expected, that `current.version` is not found, it will return an empty value.

optin/reporting/models/completed.js
```diff
@@ -20,77 +20,83 @@ const completedDB = {
...
+				}
+			);
+
+			return result.map(item => ({
+				...item,
+				nemSource: JSON.parse(item.nemSource),
+				symbolDestination: JSON.parse(item.symbolDestination)
+			}));
+		} catch (error) {
+			throw new Error(`Database error :${error.message}`);
```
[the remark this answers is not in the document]
Jaguar0625: i think original error message is probably fine. if you want you can log and rethrow, but what you're doing will lose the stack of original exception, which would make debugging more difficult.

optin/reporting/test/routers/api.spec.js
Jaguar0625: can you return 404 if file does not exist?

optin/reporting/test/controllers/completedController.spec.js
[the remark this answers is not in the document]
Jaguar0625: can you use `deep.equal` instead? we have been using that in most of our code and it is a bit more descriptive imo.

optin/reporting/test/models/database.spec.js
AnthonyLaw: I see it was created a helper method to share with `in_progress` and `completed` DB instance.

But I think I have a better idea, I will refactor on this

optin/reporting/test/models/completed.spec.js
AnthonyLaw: I did try that, but `forEach` is not working for `await/async`.

https://stackoverflow.com/questions/37576685/using-async-await-with-a-foreach-loop

I can change it to `for … of` if this looks better.

optin/reporting/test/models/runDBNotFoundTests.js
AnthonyLaw: it's like this way?

```js
expect(() => func())).to.throw(Error);
```

optin/reporting/test/models/completed.spec.js
AnthonyLaw: I'm not sure it makes sense, but here is what I thought.

For this test, I want to ensure that `mock` database returns the correct field in the query.
regarding the value, I believe we already have a uint test in here `optin/puller/tests/db`.

like to get your advice on this as well.

optin/reporting/test/controllers/completedController.spec.js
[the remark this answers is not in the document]
AnthonyLaw: it make sense, let me update it 👍🏼

optin/reporting/test/routers/api.spec.js
AnthonyLaw: sure thing, let me add this test.

optin/reporting/test/models/completed.spec.js
Jaguar0625: it's ok, you can leave it

optin/reporting/test/models/completed.spec.js
Jaguar0625: it is generally better to check the values to make sure the correct inputs are passed down and operated on properly

Jaguar0625:
nice job 👍

Jaguar0625 (approved):

PR #137, opened 2022-06-20
[dependency] fix: linting failures due to pylint update

Wayonb:
With the latest pylint version, new issues were found and fix
1. nodewatch/VersionChartBuilder.py:65:16: R1728: Consider using a generator instead
2. vanity/AddressGenerator.py:43:46: W0130: Duplicate value 0 in set (duplicate-value)

### DIFF
--- a/tools/vanity/vanity/AddressGenerator.py
+++ b/tools/vanity/vanity/AddressGenerator.py
@@ -40,6 +40,7 @@ class AddressGenerator:
 			bip32_root_node = bip32.from_mnemonic(mnemonic, '')  # no password
 
 			for account_index in range(0, self.max_wallet_accounts):
+				# pylint:disable=duplicate-value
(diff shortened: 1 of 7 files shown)


tools/vanity/vanity/AddressGenerator.py
```diff
@@ -40,6 +40,7 @@ def _match_all_thread(self, matcher):
 			bip32_root_node = bip32.from_mnemonic(mnemonic, '')  # no password
 
 			for account_index in range(0, self.max_wallet_accounts):
+				# pylint:disable=duplicate-value
```
Jaguar0625, line 43: this should be list 😬

Wayonb: ah...makes more sense.

Jaguar0625 (approved):

PR #177, opened 2022-08-12
[tools/vanity]: tools/vanity is missing .coveragerc file

Jaguar0625:
problem: code coverage numbers aren't being calculated correctly due to missing .coveragerc file
solution: add .coveragerc file in line with other python projects
          add tests for __main__

### DIFF
--- a/tools/vanity/vanity/__main__.py
+++ b/tools/vanity/vanity/__main__.py
@@ -23,7 +23,7 @@ def create_printer(args, network):
 	return printer
 
 
-def main():
+def main(args):
 	parser = argparse.ArgumentParser(
 		prog=None if globals().get('__spec__') is None else f'python -m {__spec__.name.partition(".")[0]}',
 		description='Symbol Vanity Address Generator'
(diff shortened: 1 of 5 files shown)


Wayonb (approved):
Looks good. 👍🏾 

If you want to exclude the one line see https://coverage.readthedocs.io/en/latest/excluding.html

tools/vanity/vanity/__main__.py
```diff
@@ -53,8 +53,9 @@ def main():
...
 	finally:
 		for outfile in printer.outfiles:
-			outfile.close()
+			if '<stdout>' != outfile.name:
+				outfile.close()
 
 
 if '__main__' == __name__:
-	main()
+	main(sys.argv[1:])
```
gimre-xymcity, line 61: hmm, why is this needed?

gimre-xymcity (dismissed):

tools/vanity/vanity/__main__.py
Jaguar0625: because main() now takes a parameter so it can be called by tests (so argparse doesn't operate on global)

PR #236, opened 2022-10-24
[explorer/nodewatch]: various fixes and updates

Jaguar0625:


explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -39,12 +39,16 @@ def nem_summary():  # pylint: disable=unused-variable
...
+
+	@app.route('/api/nem/chart/height')
+	def api_nem_chart_height():  # pylint: disable=unused-variable
 		return nem_routes_facade.json_height_chart_with_metadata()
 
-	@app.route('/nem/height')
-	def nem_height():  # pylint: disable=unused-variable
+	@app.route('/api/nem/height')
+	def api_nem_height():  # pylint: disable=unused-variable
 		return nem_routes_facade.json_height()
```
gimre-xymcity, line 52: should we keep rendered vs non-rendered apis in some kind of separate groups/regions?

explorer/nodewatch/nodewatch/NetworkRepository.py
```diff
@@ -12,19 +12,52 @@
...
 		self.main_address = main_address
+		self.main_public_key = main_public_key
 		self.endpoint = endpoint
 		self.name = name
 		self.version = version
 		self.height = height
 		self.finalized_height = finalized_height
 		self.balance = balance
-		self.has_api = has_api
+		self.roles = roles
```
gimre-xymcity, line 40: not sure if we need to add this now, but I guess adding 'node' key might be handy (for node key link txes)

explorer/nodewatch/nodewatch/NetworkConnector.py
```diff
@@ -62,7 +65,7 @@ async def _update_height(self, session, descriptor):
 
 				if 'finalized_height' in response_json:
 					descriptor.finalized_height = int(response_json['finalized_height'])
-		except (aiohttp.client_exceptions.ClientConnectorError, asyncio.TimeoutError) as ex:
+		except (aiohttp.client_exceptions.ClientConnectorError, aiohttp.client_exceptions.ContentTypeError, asyncio.TimeoutError) as ex:
```
gimre-xymcity, line 68: why there's ContentType exception?

explorer/nodewatch/nodewatch/__init__.py
Jaguar0625: it should be already

/nem
/api/nem
/symbol
/api/symbol

do you mean to swap /api/nem and /symbol ?

explorer/nodewatch/nodewatch/NetworkConnector.py
Jaguar0625: was hitting it yesterday when nodes were stopped/updating but webserver was still running and returning text/html 502 page 😬

explorer/nodewatch/nodewatch/NetworkRepository.py
Jaguar0625: can add it

Jaguar0625: notice this is not in normal peers information
and currently. due to way ingestion script, works it is not being retrieved for peer-only nodes

explorer/nodewatch/nodewatch/__init__.py
gimre-xymcity: I thought about completely separating, as we have:
/nem/harvesters
/nem/nodes
...

/symbol/voters
/symbol/harvesters
...

keep apis completely separate down in
/api/nem/foo
/api/symbol/foo

not sure if good idea, just an idea

explorer/nodewatch/nodewatch/NetworkRepository.py
gimre-xymcity: hmm, you mean peer-only or having peer role? (the latter one would be fine ofc 😁 )

gimre-xymcity (approved):

explorer/nodewatch/nodewatch/NetworkRepository.py
Jaguar0625: peer only

explorer/nodewatch/nodewatch/__init__.py
Jaguar0625: i'd rather separate nem / symbol and then api within that

PR #308, opened 2023-02-15
[explorer/nodewatch]: various bugfixes and enhancements

Jaguar0625:
[explorer/nodewatch]: expand configuration to allow arbitrary testnets
    
     problem: Symbol can have multiple testnets with different generation hash seeds and explorers
    solution: read additional network properties from config and pass down network instead of network name

    [explorer/nodewatch]: api JSON data should have Content-Type application/json
    
     problem: api routes are returning text/html Content-Type
    solution: use jsonify to return correct content type

    [explorer/nodewatch]: make NEM node public key optional
    
     problem: NetworkRepository fails when NEM node only has main public key
    solution: set node public key to None when not present (this matches Symbol handling)


explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -14,14 +17,27 @@ def create_app():
...
 	log.info(f' configured with NETWORK_NAME {network_name}')
+	log.info(f' configured with MIN_HEIGHT_CLUSTER_SIZE {min_cluster_size}')
+	log.info(f' configured with SYMBOL_EXPLORER_ENDPOINT {symbol_explorer_endpoint}')
+	log.info(f' configured with SYMBOL_GENERATION_HASH_SEED {symbol_generation_hash_seed}')
+	log.info(f' configured with NEM_EXPLORER_ENDPOINT {nem_explorer_endpoint}')
+
+	nem_network = NetworkLocator.find_by_name(NemNetwork.NETWORKS, network_name)
+	symbol_network = NetworkLocator.find_by_name(SymbolNetwork.NETWORKS, network_name)
+	if symbol_generation_hash_seed:
+		symbol_network.generation_hash_seed = symbol_generation_hash_seed
```
Wayonb, line 37: The generation hash needs to be converted to Hash256.

Wayonb: https://github.com/symbol/product/blob/48243dd6206c3396b7b07bdfb6b69b01b7a24eec/explorer/nodewatch/nodewatch/NetworkRepository.py#L185

Wayonb (approved):

gimre-xymcity (approved):

PR #323, opened 2023-02-27
[explorer/puller]: init explorer puller

AnthonyLaw:
init new explorer puller sub-repo
- added script and ci
- added nem client request
- added Postgres SQL database setup
- added nemPullerFacade
- added Block Model 
- added workflows script: sync block data from the network
- added unit test


explorer/puller/db/Databases.py
```diff
@@ -0,0 +1,31 @@
...
+
+		self._nem_db_connection = psycopg2.connect(
+			database="nem_db",
+			user='user',
+			password='admin',
+			host='localhost',
+			port= '54320'
+		)
+
+		self.nem = NemDatabase(self._nem_db_connection)
```
gimre-xymcity, line 25: This seems to mimic optin puller db, but this might not be the best idea.
The reason optin db was made this way is that there are multiple databases and in multiple places, we want to access multiple of them.

Here however, I imagine, we'd rather want abstract certain operations and have similar classes for both Nem and Symbol. It really begs to use inheritance in this case.

Consider following:
```py
class DatabaseConnection:
	def __init__(self, connection_string):
		self.connection_string = connection_string

	def __enter__(self):
		print(f'opening connection {self.connection_string}')
		return self

	def __exit__(self, exc_type, exc_value, traceback):
		print(f'closing connection {self.connection_string}')

class NemDatabase(DatabaseConnection):
	def create_tables(self):
		print('creating tables')
```

This looks quite similar, but what differs is usage later:
```py
with NemDatabase('...') as db:
	db.create_tables()
```

moreover, you can later hide the guts behind some kind of Facades (note this is pretty dumb example):
```py
class NemPullerFacade:
  Db = NemDatabase
  Client = NemClient

class SymbolPullerFacade:
  Db = SymbolDatabase
  Client = SymbolClient
  ```
  
  with this in place, you'll later be able to have chain-agnostic usage:
  ```py
      client = facade.Client(...)
      
      with facade.Db() as db:
        db.nem.create_tables()

        db_height = db.get_current_height()
        chain_height = await client.height()
        ...
```

:)

gimre-xymcity (changes requested):

explorer/puller/model/Block.py
```diff
@@ -0,0 +1,55 @@
...
+		self.timestamp = timestamp
+		self.total_fees = total_fees
+		self.total_transactions = total_transactions
+		self.difficulty = difficulty
+		self.block_hash = block_hash
+		self.signer = signer
+
+	@staticmethod
+	def convert_timestamp_to_datetime(timestamp):
+		return (Block.NEM_EPOCH + timedelta(seconds=timestamp)).strftime('%Y-%m-%d %H:%M:%S')
```
gimre-xymcity, line 24: this should happen via facade:
`facade.network.datetime_converter.to_datetime(timestamp)`

gimre-xymcity (changes requested):

explorer/puller/model/Block.py
```diff
@@ -0,0 +1,55 @@
...
+
+	@classmethod
+	def from_nem_block_data(cls, block_data, nem_facade):
+		block = block_data['block']
+		transactions = block_data['txes']
+
+		timestamp = cls.convert_timestamp_to_datetime(block['timeStamp'])
+		total_fees = sum(tx['tx']['fee'] for tx in transactions)
+		difficulty = f'{round(block_data["difficulty"] / (10 ** 14) * 100, 2):.2f}%'
```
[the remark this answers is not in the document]
AnthonyLaw: yeah, that sounds good to me 👍🏼

explorer/puller/facade/NemPullerFacade.py
```diff
@@ -0,0 +1,80 @@
...
+		"""Sync network blocks."""
+
+		# sync network blocks in database
+		while chain_height > db_height:
+
+			blocks = await self.client().get_blocks_after(db_height)
+
+			for block in blocks['data']:
+				save_block = Block.from_nem_block_data(block, self.nem_facade)
+				self.database().insert_block(save_block)
```
gimre-xymcity, line 75: given, that you're not inserting transactions, it _might_ be good idea, to build a cursor (and therefore SQL transaction) around this whole FOR loop, and commit after loop is done - it should be _much_ faster than committing after every block.

explorer/puller/tests/facade/test_NemPullerFacade.py
```diff
@@ -0,0 +1,26 @@
...
+			temp_config_file.flush()
+			config_file = temp_config_file.name
+
+			# Act:
+			facade = NemPullerFacade(node_url, config_file)
+
+			# Assert:
+			self.assertIsInstance(facade.database(), NemDatabase)
+			self.assertIsInstance(facade.client(), NemClient)
+			self.assertEqual(facade.client().endpoint, node_url)
```
gimre-xymcity, line 26: ~~*cough* unfortunately, this should have at least some basic test of `sync_blocks` (same for other untested methods)~~ 

ok, I see you have that below, why separate file though?

explorer/puller/tests/workflows/test_sync_nem_block.py
```diff
@@ -0,0 +1,78 @@
...
+		)
+
+		# Act:
+		await sync_blocks(nem_client_mock, databases_mock, db_height, chain_height, nem_facade)
+
+		# Assert:
+		self.assertEqual(nem_client_mock.get_blocks_after.call_count, 1)
+		self.assertEqual(databases_mock.insert_block.call_count, 2)
+		self.assertEqual(databases_mock.insert_block.call_args_list[0][0][0], expected_block_1)
+		self.assertEqual(databases_mock.insert_block.call_args_list[1][0][0], expected_block_2)
```
gimre-xymcity, line 78: depending on changes around cursor (mentioned earlier) this file might need additional failure test
(In case one of `insert_block()` will fail)

gimre-xymcity (changes requested):
One general comment:
`Block.from_nem_block_data` looks pretty much artificial and it _seems that it should rather be moved into NemPullerFacade
(I mean NEM puller facade "knows" how nem blocks look like, same will go for symbol puller facade)

explorer/puller/db/DatabaseConnection.py
```diff
@@ -0,0 +1,22 @@
...
+
+class DatabaseConnection:
+	def __init__(self, db_config):
+		self.connection = psycopg2.connect(
+			database=db_config['database'],
+			user=db_config['user'],
+			password=db_config['password'],
+			host=db_config['host'],
+			port=db_config['port']
+		)
```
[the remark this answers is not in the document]
AnthonyLaw: nope, let me move into `__enter__`

explorer/puller/client/NemClient.py
```diff
@@ -0,0 +1,36 @@
...
+
+		url_path = 'local/chain/blocks-after'
+		blocks = await self.post(url_path, {'height': height})
+
+		return blocks
+
+	async def get_block(self, height):
+		""""Gets Block data"""
+
+		url_path = 'block/at/public'
```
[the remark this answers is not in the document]
AnthonyLaw: I used this `local/chain/blocks-after`  endpoint because it can get a block list (10 blocks), so I can sync the block faster

I don't know about his endpoint `/local/block/at`, this endpoint is missing in those docs https://nemproject.github.io/

good to know about this endpoint! 👍🏼

explorer/puller/tests/facade/test_NemPullerFacade.py
AnthonyLaw: `sync_blocks` method was in `sync_nem_block.py`, after refactoring it moved to `NemPullerFacade.py`
I will update the unit test.

explorer/puller/facade/NemPullerFacade.py
```diff
@@ -0,0 +1,105 @@
...
+			blocks = await self.client().get_blocks_after(db_height)
+
+			# initialize cursor
+			cursor = self.database().connection.cursor()
+
+			for block in blocks['data']:
+				self._store_block(cursor, block)
+
+			# commit changes
+			self.database().connection.commit()
```
gimre-xymcity, line 100: 👍

explorer/puller/tests/facade/test_NemPullerFacade.py
```diff
@@ -0,0 +1,192 @@
...
+			cursor.execute('SELECT * FROM blocks')
+			results = cursor.fetchall()
+
+			results = self._convert_memoryview_to_bytes(results)
+
+			facade._client.get_blocks_after.assert_called_once_with(1)
+			self.assertEqual(len(results), 2)
+			self.assertEqual(results[0], self.expected_result_0)
+			self.assertEqual(results[1], self.expected_result_1)
+
```
gimre-xymcity, line 167: you should have additional test, where sync_blocks fails.
you can simulate this by first calling sync_blocks(1,2) inside Arrange, and then calling it once more inside Act (i.e. with (2, 5).

the second call should fail (cause you already have height == 2 in db)

the reason for this test is to show that on failure no new blocks are added to db

PR #406, opened 2023-06-07
[explorer/frontend] task: initial setup, project structure

OlegMakarenko:
## Problem
- There is no frontend project for NEM Explorer.

## Solution
- Setup Next.js project.
- Added config for `symbol_crypto_wasm`.
- Added `eslint`, `prettier` and `stylelint`.
- Added `jest` and `react-testing-library`.
- Added the global style configuration and scss variables.
- Added stub pages.
- Added stub components.
- Added stub unit test.


explorer/frontend/components/FieldPrice.jsx
```diff
@@ -0,0 +1,23 @@
...
+
+const FieldPrice = ({ value, change }) => {
+	let changeText;
+	let changeClassName;
+
+	if (change < 0) {
+		changeText = `${change}%`;
+		changeClassName = styles.changeDecrease;
+	} else {
+		changeText = `+${change}%`;
```
gimre-xymcity, line 11: you can cover both branches with something like 
``const changeText = `${Math.sign(change)}${change}%` ``

explorer/frontend/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "dependencies": {
+    "bootstrap": "^5.1.3",
+    "next": "^13.3.1",
+    "prop-types": "^15.8.1",
+    "query-string": "^7.1.1",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-toastify": "^9.1.3",
+    "sass": "^1.49.0",
+    "symbol-crypto-wasm-web": "^0.1.1",
```
gimre-xymcity, line 26: will this be symbol explorer or nem explorer or both?

explorer/frontend/components/FieldPrice.jsx
OlegMakarenko: @gimre-xymcity actually `Math.sign()` returns `-1`, `0` or `1`, so I don't think it could be used here.

PR #443, opened 2023-06-22
[faucet/fronted]: fix dynamic docker run on image

AnthonyLaw:
## What was the issue?
- Faucet frontend docker image is not able to assign dynamic env params, which it needs to separate build different images to run NEM / Symbol website.

## What's the fix?
- instead of building a website inside the images, I move the build logic in the docker run time.
- the docker run will accept env params, and build different website

```
docker run -e NEXT_PUBLIC_BUILD_TARGET="nem" \
    -e NEXT_PUBLIC_META_TITLE="The XEM Faucet" \
    -e NEXT_PUBLIC_META_DESCRIPTION="NEM Faucet" \
    -e NEXT_PUBLIC_AUTH_URL=https://faucet-auth.url.io \
    -e NEXT_PUBLIC_BACKEND_URL=https://faucet-backend.url.io \
    -e NEXT_PUBLIC_AES_SECRET="secret" \
    -p 3000:3000 \
    --name nem-frontend \
   frontend:latest
```

### DIFF
--- a/faucet/frontend/Dockerfile
+++ b/faucet/frontend/Dockerfile
@@ -3,5 +3,8 @@ WORKDIR /app
 COPY . .
 RUN npm install && npm install -g serve
 
+# Build the app to cache the dependencies
+RUN npm run build
+
 EXPOSE 3000
 CMD ["/bin/sh", "-c", "./scripts/docker/startup.sh"]
--- a/faucet/frontend/Dockerfile
+++ b/faucet/frontend/Dockerfile
@@ -1,7 +1,7 @@
 FROM node:lts-alpine AS builder
 WORKDIR /app
 COPY . .
-RUN npm install && npm run build
-RUN npm install -g serve
+RUN npm install && npm install -g serve
+
 EXPOSE 3000
-CMD ["serve", "-s", "build"]
+CMD ["/bin/sh", "-c", "./scripts/docker/startup.sh"]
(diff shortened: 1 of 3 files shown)


faucet/frontend/scripts/docker/startup.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/sh
+
+npm install
```
Wayonb, line 3: can you run the npm install during the docker build?

Wayonb (changes requested):

faucet/frontend/scripts/docker/startup.sh
AnthonyLaw: yes, that makes sense.

faucet/frontend/Dockerfile
```diff
@@ -1,7 +1,10 @@
 FROM node:lts-alpine AS builder
 WORKDIR /app
 COPY . .
-RUN npm install && npm run build
-RUN npm install -g serve
+RUN npm install && npm install -g serve
+
+# Build the app to cache the dependencies
+RUN npm run build
```
Wayonb, line 7: The npm run build can be on the same RUN line as the npm install since the build should not change unless deps change.

Wayonb (approved):
Looks good. 👍🏾

Wayonb (approved):

PR #479, opened 2023-07-07
[Faucet/authenticator] task: add redirectUrl when twitter auth endpoint 

AnthonyLaw:
## What was the issue?
- it required multiple authenticators for different front-end web apps, and it's not deployed-friendly.

## What's the fix?
- added redirect Url in query params, so it can accept different redirect URLs from different frontend web apps.


Jaguar0625 (approved):
👍

faucet/authenticator/test/controllers/index_spec.js
```diff
@@ -34,10 +35,10 @@ describe('twitter controller', () => {
 			generateAuthLinkStub.returns(Promise.resolve(twitterAuthResponse));
 
 			// Act:
-			const result = await twitter.requestToken();
+			const result = await twitter.requestToken(frontendCallbackUrl);
```
Wayonb, line 38: Should there be a test when the callback is missing from the param?

Wayonb (approved):

faucet/authenticator/test/controllers/index_spec.js
AnthonyLaw: I had another test in `throws error when request token fail`
do you think that is enough because as long as the callback is not valid (with or without params) I will throw the error.

PR #515, opened 2023-07-20
[tools/shoestring] fix: add support for Ubuntu 20.04

Wayonb:
problem: Ubuntu 20.04 ships with an older version of Python and Openssl which caused failures when running shoestring.
               1. Openssl 1.1.x output is different from Openssl 3.x
               2. Python 3.8, ``shutil.move`` does not support a path object.
solution: 1. Normalize the Openssl output so it works for both version
               2. Python 3.8 convert the path objects to string.

### DIFF
--- a/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
+++ b/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
@@ -22,27 +22,37 @@ def _openssl_run_sclient_verify(hostname, test_args):
 		'-servername', hostname,
 		'-connect',
 		'localhost:3001'
-	] + test_args, command_input='Q', show_output=False)
+	] + test_args + [
+		'|',
+		'openssl',
+		'x509',
+		'-inform',
+		'pem',
+		'-noout',
+		'-text'
+	], command_input='Q', show_output=False, use_shell=True)
 
+	searching_for_date = [r'Not Before: (.*)', r'Not After : (.*)']
 	collect_dates = False
 	collected_dates = []
 	for line in lines:
 		line = line.strip()
-		if collect_dates:
-			res = re.search(r'NotBefore: (.*); NotAfter: (.*)', line)
+		if collect_dates and searching_for_date:
+			res = re.search(searching_for_date[0], line)
 			if res:
-				collected_dates.append((parsedate_to_datetime(res.group(1)), parsedate_to_datetime(res.group(2))))
+				collected_dates.append(parsedate_to_datetime(res.group(1)))
+				searching_for_date.pop(0)
 
-		if line == 'Certificate chain':
+		if line == 'Certificate:':
 			collect_dates = True
 
-		if line.startswith('Verify return code:'):
-			if 'Verify return code: 0 (ok)' == line:
-				return True, collected_dates
-
+		if line.startswith('verify error:'):
+			# normalize error message
+			if '1.1.1' in openssl_executor.version():
+				line = line.replace('self signed', 'self-signed')
 			return False, line
 
-	return False, 'could not parse s_client response'
+	return (True, collected_dates) if 2 == len(collected_dates) else (False, 'could not parse s_client response')
 
 
 async def validate(context):
@@ -54,7 +64,7 @@ async def validate(context):
 	if not result:
 		log.warning(_('health-rest-https-certificate-invalid').format(error_message=dates_or_error))
 	else:
-		date_range = dates_or_error[-1]
+		date_range = dates_or_error
 		log.info(_('health-rest-https-certificate-valid').format(
 			start_date=date_range[0].strftime('%y-%m-%d'),
 			end_date=date_range[1].strftime('%y-%m-%d')))
--- a/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
+++ b/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
@@ -1,6 +1,8 @@
 import os
 import re
+import tempfile
 from email.utils import parsedate_to_datetime
+from pathlib import Path
 
 from zenlog import log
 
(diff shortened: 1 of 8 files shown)


tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,45 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
 	] + test_args, command_input='Q', show_output=False)
 
+	# verify the certificate is valid
+	for line in lines:
+		line = line.strip()
+		if line.startswith('verify error:'):
+			# normalize error message
+			if '1.1.1' in openssl_executor.version():
+				line = line.replace('self signed', 'self-signed')
+			return False, line
```
Jaguar0625, line 36: is it possible to add test for this case?

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,45 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
+
+	with tempfile.TemporaryDirectory() as temp_directory:
+		temp_file = Path(temp_directory) / 'cert.txt'
+		with open(temp_file, 'w', encoding='utf-8') as cert_file:
```
Jaguar0625, line 40: cert_outfile ?

Jaguar0625: wt

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,45 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
+			cert_file.writelines(lines)
+
+		lines = openssl_executor.dispatch([
+			'x509',
+			'-inform',
+			'pem',
```
Jaguar0625, line 46: consider putting these on same line since pem is input to inform

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,45 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
+		], command_input='Q', show_output=False)
+
+	searching_for_date = [r'Not Before: (.*)', r'Not After : (.*)']
 	collect_dates = False
 	collected_dates = []
 	for line in lines:
 		line = line.strip()
-		if collect_dates:
-			res = re.search(r'NotBefore: (.*); NotAfter: (.*)', line)
+		if collect_dates and searching_for_date:
```
Jaguar0625, line 58: searching_for_date => date_patterns?

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,45 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
 			collect_dates = True
 
-		if line.startswith('Verify return code:'):
-			if 'Verify return code: 0 (ok)' == line:
-				return True, collected_dates
-
-			return False, line
-
-	return False, 'could not parse s_client response'
+	return (True, collected_dates) if 2 == len(collected_dates) else (False, 'could not parse s_client response')
```
Jaguar0625, line 67: original version would collect multiple certificate date ranges

but idk if we need that anymore, so new version is probably fine

cc: @gimre-xymcity

tools/shoestring/shoestring/internal/OpensslExecutor.py
```diff
@@ -37,6 +37,6 @@ def dispatch(self, args, command_input=None, show_output=True):
 
 			if 0 != process.returncode:
 				formatted_command_line = ' '.join(command_line)
-				raise RuntimeError(f'{formatted_command_line} exited with {process.returncode}')
+				raise RuntimeError(f'{formatted_command_line} exited with {process.returncode}, stdout: {stdout_lines}')
```
Jaguar0625, line 40: stdout:\n ?

tools/shoestring/tests/commands/test_renew_certificates.py
```diff
@@ -82,7 +82,7 @@ async def test_can_renew_node_certificate():
 
 
 async def test_can_renew_node_certificate_with_ca_password():
-	await _assert_can_renew_node_certificate('abc')
+	await _assert_can_renew_node_certificate('abcd')
```
Jaguar0625, line 85: why this change? is password limit different?

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
Wayonb: Add a testcase for the normalize message scenario?
There is a test for self-signed -  ``test_validate_fails_when_certificate_is_self_signed``

tools/shoestring/tests/commands/test_renew_certificates.py
Wayonb: yea, the password limit is different between versions.

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
Jaguar0625: i mean error branch

tools/shoestring/shoestring/internal/OpensslExecutor.py
Jaguar0625: can remove space after \n now :)

tools/shoestring/tests/healthagents/test_rest_https_certificate.py
```diff
@@ -53,15 +55,23 @@ async def _dispatch_validate(test_args=None):
 	HealthAgentContext = namedtuple('HealthAgentContext', ['hostname', 'test_args'])
 	context = HealthAgentContext('localhost', test_args or [])
 
-	await asyncio.to_thread(_validate_thread, context)
+	await asyncio.get_running_loop().run_in_executor(None, _validate_thread, context)
 
 
 async def test_validate_fails_when_certificate_is_self_signed(server, caplog):  # pylint: disable=redefined-outer-name, unused-argument
+	# Arrange:
+	error_message = 'HTTPS certificate looks invalid: verify error:num=18:self-signed certificate'
```
Jaguar0625, line 63: expected_error_message?

Jaguar0625 (approved):

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -24,25 +26,40 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
+		with open(temp_file, 'wt', encoding='utf-8') as cert_outfile:
+			cert_outfile.writelines(lines)
+
+		lines = openssl_executor.dispatch([
+			'x509',
+			'-inform', 'pem',
+			'-noout',
+			'-text',
+			'-in', temp_file
+		], command_input='Q', show_output=False)
```
gimre-xymcity, line 46: does `x509` command require `command_input`? (it should be fire and forget, that's why I doubt that)

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
gimre-xymcity: I think we do.
IIRC concatenated pem file has CA followed by node cert, and IIRC `x509` only parses FIRST certificate :| (that's why I've used `s_client` output)
There was some other command that could be used, I just don't remember what it was.

PR #517, opened 2023-07-21
[tools/shoestring]: Python 3.11 IntFlag returns enum value instead of enum name

Wayonb:
problem: In python 3.11, IntFlag __str__ change to return enum value instead of the enum name.
               This broke parsing logic for NodeFeatures.
solution: to support python 3.11, override the default __str__ to always return enum names for all python versions

### DIFF
--- a/tools/shoestring/shoestring/internal/NodeFeatures.py
+++ b/tools/shoestring/shoestring/internal/NodeFeatures.py
@@ -1,4 +1,4 @@
-from enum import IntFlag
+from enum import Enum, IntFlag
 from functools import reduce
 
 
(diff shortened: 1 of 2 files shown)


Jaguar0625:
I think you have overcomplicated. Wouldn't something like this work?
https://github.com/symbol/product/commit/2d3e188cba620215913b9725184588bd4447b0d0

Wayonb:
> I think you have overcomplicated. Wouldn't something like this work? [2d3e188](https://github.com/symbol/product/commit/2d3e188cba620215913b9725184588bd4447b0d0)

yea, maybe a bit 😅 .  But my goal was to remove dependency on their ``__str__`` to prevent future issues like this.

Wayonb:
Close this since @Jaguar0625 have a similar version - https://github.com/symbol/product/tree/shoestring/py311-alt

PR #527, opened 2023-07-26
[tools/shoestring]: allow collect of multiple certificates date

Wayonb:
Update to use ``openssl storeutl`` to parse all certificates instead of ``openssl x509`` which only returns information for the first.
``openssl storeutl`` will fail if the input is not a valid pem, extract the certificate information from the output of ``openssl s_client``.

### DIFF
--- a/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
+++ b/tools/shoestring/shoestring/healthagents/rest_https_certificate.py
@@ -27,25 +27,38 @@ def _openssl_run_sclient_verify(hostname, test_args):
 	] + test_args, command_input='Q', show_output=False)
 
 	# verify the certificate is valid
+	certificates = []
+	collect_certificates = False
 	for line in lines:
 		line = line.strip()
 		if line.startswith('verify error:'):
 			return False, line
 
+		# collect all certificates
+		if line.startswith('-----BEGIN CERTIFICATE-----'):
+			collect_certificates = True
+
+		if collect_certificates:
+			certificates.append(f'{line}\n')
+
+		if line.startswith('-----END CERTIFICATE-----'):
+			collect_certificates = False
+
 	with tempfile.TemporaryDirectory() as temp_directory:
 		temp_file = Path(temp_directory) / 'cert.txt'
 		with open(temp_file, 'wt', encoding='utf-8') as cert_outfile:
-			cert_outfile.writelines(lines)
+			cert_outfile.writelines(certificates)
 
 		lines = openssl_executor.dispatch([
-			'x509',
-			'-inform', 'pem',
-			'-noout',
+			'storeutl',
 			'-text',
-			'-in', temp_file
-		], command_input='Q', show_output=False)
+			'-noout',
+			'-certs',
+			temp_file
+		], show_output=False)
 
-	date_patterns = [r'Not Before: (.*)', r'Not After : (.*)']
+	date_patterns = []
+	certificate_dates = []
 	collect_dates = False
 	collected_dates = []
 	for line in lines:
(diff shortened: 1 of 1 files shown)


tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -27,39 +27,63 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
 		line = line.strip()
 		if line.startswith('verify error:'):
 			return False, line
 
+		# collect all certificates
+		if line.startswith('-----BEGIN CERTIFICATE-----'):
+			collect_certificates = True
+
+		if collect_certificates:
+			certificates.append(line + '\n')
```
Jaguar0625, line 42: use f string

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -27,39 +27,63 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
+
+		if line.startswith('-----END CERTIFICATE-----'):
+			collect_certificates = False
+
 	with tempfile.TemporaryDirectory() as temp_directory:
 		temp_file = Path(temp_directory) / 'cert.txt'
 		with open(temp_file, 'wt', encoding='utf-8') as cert_outfile:
-			cert_outfile.writelines(lines)
-
```
Jaguar0625, line 39: add back blank line for readability

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -27,39 +27,63 @@ def _openssl_run_sclient_verify(hostname, test_args):
 	] + test_args, command_input='Q', show_output=False)
 
 	# verify the certificate is valid
+	certificates = []
+	collect_certificates = False
```
Jaguar0625, line 31: "cert" and "certificate" are being used interchangeably in variable names, consider picking one and using it throughout

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -27,39 +27,63 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
-		if line == 'Certificate:':
+				continue
+
+		if 'Certificate:' == line:
 			collect_dates = True
+			date_patterns = [r'Not Before: (.*)', r'Not After : (.*)']
+
+		if line.startswith('Total found:'):
+			if f'Total found: {len(collected_dates)}' == line:
+				return True, collected_dates
```
Jaguar0625, line 84: why do you need this condition vs just returning true at end if len(collected_dates) != 1

Wayonb: Currently, it verifies that the certificates found by OpenSSL match the number of dates collected.
If this is a best-effort scenario, then it can check that at least one date is collected.

Jaguar0625 (approved):
looks good to me but wait for @gimre-xymcity review before merging

tools/shoestring/shoestring/healthagents/rest_https_certificate.py
```diff
@@ -27,39 +27,63 @@ def _openssl_run_sclient_verify(hostname, test_args):
...
-			'x509',
-			'-inform', 'pem',
-			'-noout',
+			'storeutl',
 			'-text',
-			'-in', temp_file
+			'-noout',
+			'-certs',
+			temp_file
 		], command_input='Q', show_output=False)
```
gimre-xymcity, line 58: this probably should not require `command input` ?

PR #690, opened 2023-11-01
[explorer/rest]: added sort block by height filter in blocks endpoints

AnthonyLaw:
## What was the issue?
- Currently, blocks are returned in ASC order by height, which is not user-friendly, In the explorer, we need the blocks to be sorted by default in DESC order of height.

## What's the fix?
- Added a sort filter to the blocks list.
- Set the default sorting of block height to DESC order.

### DIFF
--- a/explorer/rest/rest/db/NemDatabase.py
+++ b/explorer/rest/rest/db/NemDatabase.py
@@ -39,15 +39,16 @@ class NemDatabase(DatabaseConnectionPool):
 
 			return self._create_block_view(result) if result else None
 
-	def get_blocks(self, limit, offset, min_height):
+	def get_blocks(self, limit, offset, min_height, sort):
 		"""Gets blocks pagination in database."""
 
 		with self.connection() as connection:
 			cursor = connection.cursor()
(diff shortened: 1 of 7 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -39,15 +39,16 @@ def get_block(self, height):
...
 		"""Gets blocks pagination in database."""
 
 		with self.connection() as connection:
 			cursor = connection.cursor()
-			cursor.execute('''
+			cursor.execute(f'''
 				SELECT *
 				FROM blocks
 				WHERE height >= %s
+				ORDER BY height {sort}
```
Jaguar0625, line 51: is there a situation where blocks are added to the db out of order? if not, it should be faster to sort by id instead of height

AnthonyLaw: nope, the block is inserted in order actually.

I just found out, my block table didn't have id 😬 
I will update my table.

Jaguar0625 (approved):
👍

PR #692, opened 2023-11-02
[explorer/rest]: added size filed in blocks

AnthonyLaw:
## What was the issue?
- Missing block size in the block view field.

## What's the fix?
- Added `size` in block view and model


Jaguar0625 (approved):
👍

PR #693, opened 2023-11-02
[explorer/rest]: Implemented CORS

AnthonyLaw:
## What was the issue?
- The frontend UI was experiencing issues accessing the server due to missing CORS implementation.

## What's the fix?
- Implemented CORS to resolve the access issues.

### DIFF
--- a/explorer/rest/tests/test_rest.py
+++ b/explorer/rest/tests/test_rest.py
@@ -68,6 +68,7 @@ def _assert_get_api_nem_block_by_height(client, height, expected_status_code, ex
 	# Assert:
 	assert expected_status_code == response.status_code
 	assert expected_result == response.json
+	assert response.headers['Access-Control-Allow-Origin'] == '*'
 
 
 def test_api_nem_block_by_height(client):  # pylint: disable=redefined-outer-name
@@ -125,6 +126,7 @@ def test_api_nem_blocks_without_params(client):  # pylint: disable=redefined-out
 	# Assert:
 	assert 200 == response.status_code
 	assert [BlockView(*BLOCKS[1]).to_dict(), BlockView(*BLOCKS[0]).to_dict()] == response.json
+	assert response.headers['Access-Control-Allow-Origin'] == '*'
 
 
 def test_api_nem_blocks_applies_limit(client):  # pylint: disable=redefined-outer-name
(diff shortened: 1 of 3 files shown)


explorer/rest/tests/test_rest.py
```diff
@@ -68,6 +68,7 @@ def _assert_get_api_nem_block_by_height(client, height, expected_status_code, ex
 	# Assert:
 	assert expected_status_code == response.status_code
 	assert expected_result == response.json
+	assert response.headers['Access-Control-Allow-Origin'] == '*'
```
Jaguar0625, line 71: consider helper function that tests status code + headers, so this is checked in all cases

e.g.
_assert_status_code_and_headers(response, 200)

explorer/rest/tests/test_rest.py
```diff
@@ -59,15 +59,21 @@ def client(app):  # pylint: disable=redefined-outer-name
...
 # region /block/<height>
 
 def _assert_get_api_nem_block_by_height(client, height, expected_status_code, expected_result):  # pylint: disable=redefined-outer-name
 	# Act:
 	response = client.get(f'/api/nem/block/{height}')
 
 	# Assert:
-	assert expected_status_code == response.status_code
 	assert expected_result == response.json
+	_assert_status_code_and_headers(response, expected_status_code)
```
Jaguar0625, line 76: generally would swap order because `_assert_status_code_and_headers` is testing common part and should be in most tests and then more specific checks follow

explorer/rest/tests/test_rest.py
```diff
@@ -59,15 +59,21 @@ def client(app):  # pylint: disable=redefined-outer-name
 # endregion
 
 
+def _assert_status_code_and_headers(response, expected_status_code):
+	# Assert:
```
Jaguar0625, line 63: can remove this since name of function is clear that this is part of assert stage

explorer/rest/tests/test_rest.py
```diff
@@ -102,29 +108,29 @@ def _assert_get_api_nem_blocks(client, expected_code, expected_result, **query_p
 	response = _get_api_nem_blocks(client, **query_params)
 
 	# Assert:
-	assert expected_code == response.status_code
 	assert expected_result == response.json
+	_assert_status_code_and_headers(response, expected_code)
```
Jaguar0625, line 112: try to be consistent with naming wrt `expected_code` and `expected_status_code` throughout file

explorer/rest/tests/test_rest.py
AnthonyLaw: make sense 👍🏼

Jaguar0625 (approved):
👍

PR #694, opened 2023-11-02
[tools/shoestring]: update shoestring to support python 3.11

Jaguar0625:
problem: python 3.11 changed Enum.__str__ to use int.__str__
solution: add NodeFeatures.to_formatted_string and use as appropriate


Wayonb (approved):
👍🏾

PR #702, opened 2023-11-07
[explorer/rest]: added namespace and namespaces endpoint

AnthonyLaw:
## What was the issue?
- Implement namespace and namespaces endpoint.

## What's the fix?
- Added search namespace by name.
- Added search namespace listing.
- Added unit test


AnthonyLaw:
Reopen when it ready

PR #703, opened 2023-11-08
[explorer/rest]: patch missing project name inside gitlintmodules

AnthonyLaw:
## What was the issue?
- Jenkins verify commit failed

## What's the fix?
- patch missing name inside gitlintmodules


Wayonb (approved):
Thanks 👍🏾

PR #837, opened 2024-03-27
[snap/frontend]: added basic components

AnthonyLaw:
## What was the issue?
- It needs basic components

## What's the fix?
- added button
- added modal box
- added input box
- added dropdown
- added eol-last eslint config


snap/frontend/.eslintrc
```diff
@@ -5,6 +5,9 @@ extends:
 - plugin:jsdoc/recommended-error
 - ../../linters/javascript/default.eslintrc
 rules:
+  eol-last:
```
Wayonb, line 8: eol-last pluggin seems to be deprecated and should use https://eslint.style/rules/js/eol-last

snap/frontend/components/Button/Button.spec.jsx
```diff
@@ -0,0 +1,56 @@
+import Button from '.';
+import { fireEvent, render, screen } from '@testing-library/react';
+
+describe('components/Button', () => {
+	it('renders Button', () => {
+		// Arrange:
```
Wayonb, line 6: what you think about ``can render`` instead of ``renders Button`` since the main describe is ``Button``?  @Jaguar0625 might have a better suggestion.

snap/frontend/components/Button/Button.spec.jsx
```diff
@@ -0,0 +1,56 @@
...
+		const buttonText = 'Click me';
+
+		// Act:
+		render(<Button>Click me</Button>);
+		const buttonElement = screen.getByRole('button');
```
Wayonb, line 11: Should this line be in assert?  It seems the act it just render.

Wayonb: same for the below tests

Wayonb (changes requested):

Wayonb (approved):
👍🏾

PR #843, opened 2024-04-06
[tools/shoestring] fix: initial mainnet package for Catapult 1.0.3.7 release incorrect

Wayonb:
problem: initial mainnet package for catapult 1.0.3.7 had only dirs
solution: recreate the mainnet package


Jaguar0625 (approved):
👍

PR #851, opened 2024-04-09
[tools/shoestring]: fix health check failure on nodes with https REST

Jaguar0625:
problem: health agent is accessing wrong property from context
solution: fix access and update agents to use domainname when specified


Wayonb (approved):
👍🏾

Issue #867, opened 2024-04-16
[tools/shoestring]: announce-transaction fails with HTTP request failed with code 202


Jaguar0625:
202 is HTTP success code and should not result in exception. See stack below for more information.

```
  File "/usr/local/lib/python3.10/site-packages/shoestring/__main__.py", line 49, in main
    await possible_task
  File "/usr/local/lib/python3.10/site-packages/shoestring/commands/announce_transaction.py", line 30, in run_main
    await announce_transaction(transaction)
  File "/usr/local/lib/python3.10/site-packages/symbollightapi/connector/SymbolConnector.py", line 169, in announce_transaction
    await self._announce_transaction(transaction_payload, 'transactions')
  File "/usr/local/lib/python3.10/site-packages/symbollightapi/connector/SymbolConnector.py", line 162, in _announce_transaction
    await self.put(url_path, {
  File "/usr/local/lib/python3.10/site-packages/symbollightapi/connector/BasicConnector.py", line 54, in put
    return await self._dispatch('put', url_path, property_name, data=json.dumps(request_payload), headers={
  File "/usr/local/lib/python3.10/site-packages/symbollightapi/connector/BasicConnector.py", line 34, in _dispatch
    raise NodeException(error_message)
symbollightapi.model.Exceptions.NodeException: HTTP request failed with code 202
packet 9 was pushed to the network via /transactions
```

PR #869, opened 2024-04-18
[lightapi/python]: treat all HTTP statuses less than 400 as success

Jaguar0625:
problem: only HTTP statuses 200 and 404 are treated as non-error
solution: treate all statuses less than 400 and 404 as non-error

fixes: #867


Wayonb (approved):
👍🏾

Issue #853, opened 2024-04-11
[tools/shoestring]: harvesters are lost when migrating from bootstrap and/or when renewing certificates


Jaguar0625:
# Problem Description

A delegated harvester can send a special harvesting request encrypted message to a node operator. The message is encrypted with a shared key derived from an ephemeral encryption key (from the sender) and the node operator's public key. Once this message is decrypted, it is next inspected for validity. Assuming it is valid, an entry is created for this (accepted) request in the `harvesters.dat` file.

This entry is composed of the following data:
[32 bytes]: ephemeral public key

PR #870, opened 2024-04-19
[tools/shoestring]: add 'import-harvesters' command

Jaguar0625:
problem: there is no way to import a harvesters.dat file to a previously deployed node
solution: add functionality for inspecting and importing harvester.dat files

fixes: #853


Wayonb (approved):
👍🏾

Wayonb:
A couple of scenarios to think about:
1. import from bootstrap; should this command also run, or should the docs be updated?
2. the same for cert renewal.

PR #903, opened 2024-05-07
[tools/vanity] fix: TMPDIR is not defined in Ubuntu

Wayonb:
problem: vanity example is failing due to TMPDIR is undefine on Ubuntu
solution: use the mktemp to generate a temp file name


Jaguar0625 (approved):
did we miss this because examples are only being run in dev/main ?

Wayonb:
> did we miss this because examples are only being run in dev/main ?

Yes

PR #931, opened 2024-05-30
[snap/frontend]: implement initial snap

AnthonyLaw:
## What was the issue?
The page didn't load state when the page was first loaded.

## What's the fix?
- Call the `initialSnap` method when the page loads and snap is installed.
- Added loading status when waiting for initialSnap return.

### DIFF
--- a/snap/frontend/components/Navbar/index.jsx
+++ b/snap/frontend/components/Navbar/index.jsx
@@ -1,11 +1,11 @@
 import { actionTypes, useWalletContext } from '../../context';
 import Dropdown from '../Dropdown';
 import Image from 'next/image';
-import { useState } from 'react';
+import { useEffect, useState } from 'react';
 
 const Navbar = () => {
 	const { walletState, dispatch, symbolSnap } = useWalletContext();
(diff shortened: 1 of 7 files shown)


snap/frontend/components/Navbar/index.jsx
```diff
@@ -29,13 +29,21 @@ const Navbar = () => {
...
 	};
 
 	const handleSelectCurrency = option => {
 		setSelectedCurrency(option.label);
 	};
 
+	useEffect(() => {
+		// set selected network in dropdown label
+		if (network && network.hasOwnProperty('networkName')) {
+			const selected = networks.find(n => n.value === network.networkName);
```
Wayonb, line 42: What happens if the network is not found?

Wayonb (approved):
👍🏾

snap/frontend/components/Navbar/index.jsx
AnthonyLaw: I added logic to handle it.
It will skip to update network name instead of hit error.

Wayonb (approved):
👍🏾

PR #943, opened 2024-06-07
[snap/backend]: Added create account method in account utils

AnthonyLaw:
## What was the issue?
- The user needs a way to create an account from the metamask secret recovery phase.

## What's the fix?
- Added `createAccount` to create an account based on the path bip 44 index.
- Added `getLatestAccountIndex` to get the latest index from the accounts state, so use the index to create a new account.


snap/backend/src/utils/accountUtils.js
```diff
@@ -25,7 +31,79 @@ const accountUtils = {
...
+			const { walletLabel } = requestParams;
+			const { network, accounts } = state;
+
+			const facade = new SymbolFacade(network.networkName);
+
+			// Get the latest account index and increment it if it exists
+			const newAddressIndex = this.getLatestAccountIndex(accounts, network.networkName) + 1 || 0;
+
+			const newKeyPair = await this.deriveKeyPair(network.networkName, newAddressIndex);
+			const accountId = uuidv4();
```
Wayonb, line 59: How will the accountId be used vs addressIndex?  They both are unique per account.

AnthonyLaw: You are right, both of them are unique but serve different purposes and use cases.

`accountId` - this ID represents each account.
`addressIndex` - This index is used to derive the private key (for the metamask wallet).

However, we do support importing wallet private keys, in which case `addressIndex` will be null.

Wayonb (approved):
👍🏾

PR #954, opened 2024-06-12
[snap]: create default account

AnthonyLaw:
## What was the issue?
- Snap should generate the first account (mainnet/testnet) after installation.

## What's the fix?
- Refactored `initialSnap` to take `networkName` as params.
- Added `createAccount` in symbolSnapFactory.
- Added `getAccounts` method
- Added `setupSnap` into `helper.js` to be shared across components.
- Implement the changes in components.

Note: 
- when the page first loads or the network switches, `setupSnap` will always be re-called.
- if the account state is empty, it will generate the first account from the Metamask secret phrase.


Wayonb (commented):
Can you give more information of why Snap needs to create the first account?

AnthonyLaw:
> Can you give more information of why Snap needs to create the first account?

`Metamask` only provides a seed phase, Symbol Snap will take the seed phase to generate a Symbol account.

If without creating the first account, the wallet can't be used, so when the user installs the snap, the system will default generate the account from seed phase.

Wayonb:
> If without creating the first account, the wallet can't be used, so when the user installs the snap, the system will default generate the account from seed phase.

Do you mean the entire ``Metamask`` wallet is unusable, or just Symbol Snap?  I expect the user to create an account before using it.

Or are you saying the user cannot create an account without this default account?

Wayonb (approved):
👍🏾

AnthonyLaw:
> Do you mean the entire `Metamask` wallet is unusable, or just Symbol Snap?

Just Symbol snap wallet

> I expect the user to create an account before using it.

yeah, I just skipped the step (manually create) first the first wallet.
it just created the default wallet when they first installed it.

let me share the video (flow)
https://www.loom.com/share/aedb123f94df4d94b0da9dfe75420bc1?sid=2157e098-5289-4a07-874d-2f95de43fea2

PR #955, opened 2024-06-16
[snap/frontend]: added address component

AnthonyLaw:
## What was the issue?
- The address could not be copied.

## What's the fix?
- Refactor it into address components.
- Allow the address to be copied when clicking the copy button.


Wayonb (approved):
👍🏾

PR #973, opened 2024-06-21
[snap/frontend]: create multiple account

AnthonyLaw:
## What was the issue?
- it does not allow create multiple accounts

## What's the fix?
- Added AccountCreateModalBox components containing an input box for the wallet name.
- When a new account is created, it will update the account list and selected account state.
- Validation of wallet name

## Screeshot
image
Validation Input box

image
Success created new account


Wayonb (approved):
👍🏾

PR #978, opened 2024-06-24
[tools/shoestring] fix: bump lightapi and symbol-sdk to the latest version

Wayonb:
problem: when upgrading shoestring, the requirements are not upgraded to the latest.
solution: bump deps version in requirements.txt file


Jaguar0625 (approved):
👍

PR #982, opened 2024-06-28
[snap]: Implement currecny (USD/JPY) support in navbar

AnthonyLaw:
## What was the issue?
- Currency in the navbar has not been implemented.

## What's the fix?
- Added `cryptoCompareClient` to fetch XYM price in USD and JPY.
- Added `priceUtils`
    - `getPrice` fetch the price from `cryptoCompareClient`.
    - `getCurrencyPrice` return price by given symbol (usd/jpy).
- Added `currency` params in `setupSnap` method.
- Added cronjob every minute will fetch the price.
- Implement in frontend Navbar component, which the user can switch between USD / JPY

## Note
- In the backend, every minute (cronjob from snap) will request the latest price from `cryptoCompareClient` and store it as a cache.
- Frontend will access the currency price from the backend cache.

### DIFF
--- a/snap/backend/src/services/cryptocompareClient.js
+++ b/snap/backend/src/services/cryptocompareClient.js
@@ -1,0 +1,32 @@
+import config from '../config.js';
+import fetchUtils from '../utils/fetchUtils.js';
+
+const cryptoCompareClient = {
+	/**
+     * Fetch the price of the supported currencies.
+     * @returns {Promise<PriceInfo>} The price of the supported currencies.
+     */
+	fetchPrice: async () => {
+		try {
+			const currencies = config.supportCurrency.join(',');
+			const prices = await fetchUtils.fetchData(`${config.cryptoCompareURL}${currencies}`, 'GET');
+
+			return prices;
+		} catch (error) {
+			throw new Error(`Failed to fetch price from cryptoCompare service: ${error.message}`);
+		}
+	}
+};
+
+export default cryptoCompareClient;
+
+// region type declarations
+
+/**
+ * Result of a price request.
+ * @typedef {object} PriceInfo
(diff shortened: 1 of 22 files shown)


snap/backend/src/services/cryptocompareClient.js
```diff
@@ -0,0 +1,35 @@
+import config from '../config.js';
+import fetchUtils from '../utils/fetchUtils.js';
+
+const cryptoCompareClient = {
+	/**
+     * Fetch the price of the supported currencies.
```
Wayonb, line 6: extra space

snap/frontend/components/AccountBalance/index.jsx
```diff
@@ -3,10 +3,10 @@ import { useWalletContext } from '../../context';
 const AccountBalance = () => {
 	const { walletState } = useWalletContext();
 	const { mosaics, currency } = walletState;
-	const { symbol, currencyPerXYM } = currency;
+	const { symbol, price } = currency;
 
 	const xymBalance = mosaics.find(m => 'symbol.xym' === m.name)?.amount || 0;
-	const convertToCurrency = (xymBalance * currencyPerXYM).toFixed(2);
+	const convertToCurrency = (xymBalance * price).toFixed(2);
```
Wayonb, line 9: On testnet accounts, do we show the price also? 🤔

Wayonb (changes requested):

snap/frontend/components/AccountBalance/index.jsx
AnthonyLaw: Nope, the price will be disabled to view in testnet.
But for now, I leave it because it's easy for me to test in testnet.

Wayonb (approved):
👍🏾

PR #994, opened 2024-07-09
[Snap]: fetch XYM mosaic balance and display in account balance components

AnthonyLaw:
## What was the issue?
- The XYM balance was not being fetched from the network.

## What's the fix?
- Added a SymbolClient service to query the REST API.
- Added `currencyMoasicId` properties in the snap network state.
- Added `mosaicInfo` state to store all mosaic information as a cache, reducing the number of queries and making it easier to format the mosaic balance.
- Implemented querying of the account balance when create / import actions are triggered.
- Implemented querying of the first account balance when visiting the wallet. Subsequent updates of all accounts will happen in the background for a better user experience.
- Implemented these changes in the frontend account balance components.

### DIFF
--- a/snap/backend/src/services/symbolClient.js
+++ b/snap/backend/src/services/symbolClient.js
@@ -1,0 +1,18 @@
+import fetchUtils from '../utils/fetchUtils.js';
+
+const symbolClient = nodeUrl => ({
+	/**
+	 * Fetch native mosaic from network properties.
+	 * @returns {Promise<string>} the network currency mosaic id
+	 */
+	fetchNetworkCurrencyMosaicId: async () => {
+		try {
+			const { chain } = await fetchUtils.fetchData(`${nodeUrl}/network/properties`);
+			return chain.currencyMosaicId.slice(2).replace(/'/g, '');
+		} catch (error) {
+			throw new Error(`Failed to fetch network properties info: ${error.message}`);
+		}
+	}
+});
+
+export default symbolClient;
--- a/snap/backend/src/services/symbolClient.js
+++ b/snap/backend/src/services/symbolClient.js
@@ -1,4 +1,6 @@
 import fetchUtils from '../utils/fetchUtils.js';
+import { utils } from 'symbol-sdk';
+import { Address } from 'symbol-sdk/symbol';
 
 const symbolClient = {
 	create(nodeUrl) {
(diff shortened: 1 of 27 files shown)


snap/backend/src/services/symbolClient.js
```diff
@@ -0,0 +1,83 @@
...
+			 * @returns {Promise<object<string, MosaicsInfo>>} The mosaics info.
+			 */
+			fetchMosaicsInfo: async mosaicIds => {
+				if (!mosaicIds.length)
+					return {};
+
+				try {
+					const mosaics = await fetchUtils.fetchData(`${nodeUrl}/mosaics`, 'POST', { mosaicIds });
+
+					return mosaics.reduce((mosaicObj, { mosaic }) => {
```
Wayonb, line 32: Why reduce and not map?

snap/backend/src/utils/accountUtils.js
```diff
@@ -179,6 +197,75 @@ const accountUtils = {
...
+	 * @returns {Promise<Record<string, Array<AccountMosaics>>>} - The accounts mosaics object.
+	 */
+	async fetchAndUpdateAccountMosaics(state, addresses) {
+		const { network } = state;
+
+		const client = symbolClient.create(network.url);
+		const accountsMosaics = await client.fetchAccountsMosaics(addresses);
+
+		// Get all unique mosaic ids
+		const mosaicIds = [...new Set(Object.values(accountsMosaics).flatMap(accountMosaics => accountMosaics.map(mosaic => mosaic.id)))];
```
Wayonb, line 214: I think the mosaics list for an account is always unique.

snap/backend/src/utils/mosaicUtils.js
```diff
@@ -0,0 +1,65 @@
...
+import stateManager from '../stateManager.js';
+
+const mosaicUtils = {
+	/**
+     * Update mosaic info in state if given mosaic ids not exist in mosaicInfo.
+     * @param {object} state - The state object.
+     * @param {Array<string>} mosaicIds - The mosaic ids.
+     * @returns {Promise<void>} - A promise that resolves when mosaic info is updated.
+     */
+	updateMosaicInfo: async (state, mosaicIds) => {
```
Wayonb, line 11: In what case does the mosaic id gets into the account but then not found?  Or did I misunderstand the scenario?

snap/backend/src/utils/mosaicUtils.js
```diff
@@ -0,0 +1,65 @@
...
+				return;
+
+			const client = symbolClient.create(network.url);
+
+			const mosaics = await client.fetchMosaicsInfo(missingMosaicIds);
+
+			// update networkName in mosaic info
+			Object.keys(mosaics).forEach(mosaicId => {
+				mosaics[mosaicId].networkName = network.networkName;
+			});
```
Wayonb, line 28: Why is the mosaic need to have a network name as state?

Wayonb (changes requested):

snap/backend/src/services/symbolClient.js
```diff
@@ -0,0 +1,83 @@
+import fetchUtils from '../utils/fetchUtils.js';
+import { utils } from 'symbol-sdk';
+import { Address } from 'symbol-sdk/symbol';
+
+const symbolClient = {
```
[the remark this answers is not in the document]
AnthonyLaw: yes, Agree.
but for `lightapi` the data response will be more general info instead of custom output.

snap/backend/src/utils/mosaicUtils.js
AnthonyLaw: `mosaicInfo` is the cache in the state (global), which will be shared between the network (testnet/mainnet).

The reason added `networkName` is for filtering purposes, which switch between networks.

snap/backend/src/utils/accountUtils.js
AnthonyLaw: yes, you are right.
But in this case, I will collect all the mosaic IDs from all accounts.

For example, I have 10 accounts, and each account will have a mosaics list.
I will collect all mosaic IDs, and filter into unique mosaic ids.

use the id to query mosaic info.

snap/backend/src/services/symbolClient.js
AnthonyLaw: Because the goal is to create an Object, which **id** as `key`, and **{ divisibility }** as `value`.

If using `map` is will be an array of objects. 

example:

```json
{
  "0x123": {
    "divisibility": 6
  }
}
```

snap/backend/src/services/symbolClient.js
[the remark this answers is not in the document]
Wayonb: Yes, every usage will be different, but generally, we need to pull data from the same endpoint.
e.g. you are getting the currencyMosaic below 😉 
Just something to think about for the future.

snap/backend/src/utils/mosaicUtils.js
Wayonb: For security reasons, I am unsure if mainnet and testnet would share the same backend.

Wayonb (approved):
👍🏾

snap/backend/src/utils/mosaicUtils.js
AnthonyLaw: I think we can decide later after I make a release for the internal team test.

PR #996, opened 2024-07-12
[snap/backend]: Added transaction related rest query.

AnthonyLaw:
## What was the issue?
- Missing transaction related query.

## What's the fix?
- Added a query for transactions by address with offset ID, used for infinite scrolling.
- Added a query for transaction detail by transactions ID, used for extracting inner transactions


snap/backend/src/services/symbolClient.js
```diff
@@ -81,6 +81,50 @@ const symbolClient = {
 				} catch (error) {
 					throw new Error(`Failed to fetch mosaics namespace: ${error.message}`);
 				}
+			},
+			/**
+			 * Fetch transactions page by address with offset id.
+			 * @param {string} address - The address to account.
+			 * @param {string} offsetId - The offset id of transaction.
+			 * @param {'confirmed' | 'unconfirmed'} group - The group of transactions.
```
Wayonb, line 89: Will partial group be supported also?

snap/backend/src/services/symbolClient.js
```diff
@@ -81,6 +81,50 @@ const symbolClient = {
...
+			 * @returns {Promise<Record<string, Array<Transaction>>>} The transaction.
+			 */
+			fetchInnerTransactionByAggregateIds: async (transactionIds, group = 'confirmed') => {
+				if (!transactionIds.length)
+					return {};
+
+				try {
+					const innerTransactions = {};
+
+					const transactions = await fetchUtils.fetchData(`${nodeUrl}/transactions/${group}/ids`, 'POST', { transactionIds });
```
Wayonb, line 118: which endpoint is ``/ids``?

Wayonb (changes requested):

snap/backend/src/services/symbolClient.js
AnthonyLaw: Nope, right now only for `confirmed`.

For partial group I think, it should implement together with multisign account

Wayonb (approved):
👍🏾

PR #1004, opened 2024-07-16
[snap/backend]: Added transaction utils and rpc.

AnthonyLaw:
## What was the issue?
- Missing implementation transaction utils and RPC.

## What's the fix?
- Added `fetchAccountTransactions` to the RPC request.
- Added `fetchAccountTransactions` in `transactionUtils` to query account transactions, and request inner transactions if there is an aggregate transaction.
- Added transaction formatting and converted all inner transactions to a flat structure.


AnthonyLaw:
I pushed to wrong branch, cause the PR closed.
Refer the new PR here #1005

PR #1015, opened 2024-07-22
[Snap/frontend]: Refactor transaction table, and implement backend service

AnthonyLaw:
## What was the issue?
- Backend service was not implemented.
- The container style of the transaction table was misaligned when transactions loaded.

## What's the fix?
- Added infinity scrolling to the transaction table.
- Fixed the container style issue in the transaction table.
- Implemented fetching account transactions in the snap factory.
- Added explorer URL configuration, allowing generation of explorer links in the transaction table.
- Replaced confirm/pending status with the transaction type name.

## Note:
- The transaction table will initially fetch transactions from the selected address. When the user scrolls to the bottom, it will request more transactions. This will continue until the last page is reached, after which no more requests will be made.

image

### DIFF
--- a/snap/backend/snap.manifest.json
+++ b/snap/backend/snap.manifest.json
@@ -27,6 +27,12 @@
           "request": {
             "method": "fetchCurrencyPrice"
           }
+        },
+        {
+          "expression": "* * * * *",
+          "request": {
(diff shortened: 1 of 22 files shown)


Wayonb (approved):
👍🏾

PR #1028, opened 2024-07-30
[snap/backend]: Added fee multipler 

AnthonyLaw:
## What was the issue?
- The fee multiplier is missing; we will use it to calculate the transaction fee.

## What's the fix?
- Added fee multiplier endpoint
- Extract slow, average, and fast multiplier
- Added fetch multiplier in onCronjob


Wayonb (approved):
👍🏾

PR #1029, opened 2024-08-01
[snap/frontend]: Added transfer form component

AnthonyLaw:
## What was the issue?
- The transfer transaction form is missing when clicking on the send button on the home page.

## What's the fix?
- Added transfer transaction form.
- Click on the **Send** button to pop up the modal box.
- Added form validation, such as Address (network check), mosaic amount, and message length.
- Handle mosaic selector logic, when the user clicks on "Add" mosaic, the options will be filtered (no duplicate).

Preview:

image

### DIFF
--- a/snap/frontend/components/TransferModalBox/index.jsx
+++ b/snap/frontend/components/TransferModalBox/index.jsx
@@ -31,23 +31,23 @@ const TransferModalBox = ({ isOpen, onRequestClose }) => {
 				return 1024 >= value.length ? null : 'Message length should not exceed 1024 characters';
 			case 'amount':
 				return value.map((amount, index) => {
-					const maxBalance = mosaicsBalance.find(mosaic => mosaic.id === selectedMosaics[index].id).amount;
+					const currentBalance = mosaicsBalance.find(mosaic => mosaic.id === selectedMosaics[index].id).amount;
 					const absoluteAmount = amount * (10 ** mosaicInfo[selectedMosaics[index].id].divisibility);
 
-					return absoluteAmount > maxBalance ? 'Not enough balance' : null;
+					return absoluteAmount > currentBalance ? 'Not enough balance' : null;
 				});
 			default:
 				return null;
 			}
 		};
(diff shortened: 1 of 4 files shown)


snap/frontend/components/TransferModalBox/index.jsx
```diff
@@ -0,0 +1,274 @@
...
+
+	const validate = (field, value) => {
+		switch (field) {
+		case 'address':
+			return facade.network.isValidAddressString(value) ? null : 'Invalid address';
+		case 'message':
+			return 1024 >= value.length ? null : 'Message length should not exceed 1024 characters';
+		case 'amount':
+			return value.map((amount, index) => {
+				const maxBalance = mosaicsBalance.find(mosaic => mosaic.id === selectedMosaics[index].id).amount;
```
Wayonb, line 33: maxBalance => currentBalance?

snap/frontend/components/TransferModalBox/index.jsx
```diff
@@ -0,0 +1,274 @@
...
+
+				return absoluteAmount > maxBalance ? 'Not enough balance' : null;
+			});
+		default:
+			return null;
+		}
+	};
+
+	const validateForm = () => {
+		const validate = (field, value) => {
```
Wayonb, line 44: This validate function looks the same as above function?

Wayonb (changes requested):
It looks good, except for a few comments.

Also, when the user clicks send, is there a verification screen for the tx?

AnthonyLaw:
> Also, when the user clicks send, is there a verification screen for the tx?

Yes, this will be another PR.
The final pop-up confirmation will come from metamask.

snap/frontend/components/TransferModalBox/index.jsx
AnthonyLaw: good catch! 👍🏼

Wayonb (approved):
👍🏾

PR #1037, opened 2024-08-05
[snap/frontend]: Added fee multiplier component.

AnthonyLaw:
## What was the issue?
- The fee multiplier option is missing in the transfer form.

## What's the fix?
- Query fee multiplier from snap state.
- Added fee multiplier frontend.

## Preview
image

### DIFF
--- a/snap/frontend/components/FeeMultiplier/FeeMultiplier.spec.jsx
+++ b/snap/frontend/components/FeeMultiplier/FeeMultiplier.spec.jsx
@@ -28,7 +28,7 @@ describe('components/FeeMultiplier', () => {
 		});
 	});
 
-	it('update selected fee multiplier', async () => {
+	it('update selected fee multiplier (key)', async () => {
 		// Arrange:
 		context.symbolSnap.getFeeMultiplier.mockResolvedValue({
 			slow: 10,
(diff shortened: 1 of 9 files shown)


Wayonb (changes requested):
Looks good, but a couple of questions
1. how does the user know what fee values are slow, average, and fast?
2. Can the user set a custom fee?

AnthonyLaw:
> Looks good, but a couple of questions
> 
> 1. how does the user know what fee values are slow, average, and fast?
> 2. Can the user set a custom fee?

1. I was thinking of implementing fees (tx size * fee multiplier) in this screen, which means I must create and prepare transactions in the front end, which doesn't look good (because I want it to implement prepare transactions in the backend)

- does it make sense we just showing fee multiplier and the fees will be displayed in the last confirmation screen?
image


2. Nope, but I can add it in the future.

Wayonb:
> > Looks good, but a couple of questions
> > 
> > 1. how does the user know what fee values are slow, average, and fast?
> > 2. Can the user set a custom fee?
> 
> 1. I was thinking of implementing fees (tx size * fee multiplier) in this screen, which means I must create and prepare transactions in the front end, which doesn't look good (because I want it to implement prepare transactions in the backend)
> 
> * does it make sense we just showing fee multiplier and the fees will be displayed in the last confirmation screen?
> 
> image
> 2. Nope, but I can add it in the future.
1. yea, that might work but I would verify with @0x6861746366574 
   tbh, not sure if these are needed since most tx will get committed in the next block

Wayonb (approved):
👍🏾

Wayonb (approved):
👍🏾

PR #1038, opened 2024-08-07
[snap/backend]: Added announce endpoint.

AnthonyLaw:
## What was the issue?
- Missing endpoint to announce transaction.

## What's the fix?
- Added announce transaction endpoint


Wayonb (approved):
👍🏾

PR #1039, opened 2024-08-09
[faucet/authenticator]: Added profileImageUrl (original dimension).

AnthonyLaw:
## What was the issue?
- The user object doesn't have a Twitter profile URL.

## What's the fix?
- Query original dimension profile images, and added in profileImageUrl properties.


OlegMakarenko (approved):

Wayonb (approved):
Looks good 👍🏾 
What issue is caused by the missing Twitter profile URL?

AnthonyLaw:
> Looks good 👍🏾 What issue is caused by the missing Twitter profile URL?

It is a new request feature, it will be used in XYMPOSIUM website

PR #1054, opened 2024-08-12
[snap/backend]: added sign and implment announce transfer transaction 

AnthonyLaw:
## What was the issue?
- Missing sign transfer transaction

## What's the fix?
- Added sign transfer transaction and announced to network.


Wayonb (changes requested):

Wayonb (approved):
👍🏾

PR #1057, opened 2024-08-14
[snap/backend]: Fix config naming in env

AnthonyLaw:
## What was the issue?
- Missing `NEXT_PUBLIC_` naming in env config

## What's the fix?
- Added `NEXT_PUBLIC_` naming.


Wayonb (approved):
👍🏾

PR #1063, opened 2024-08-19
[snap/frontend]: WS implementation

AnthonyLaw:
## What was the issue?
- We need a WebSocket implementation to improve the wallet's user experience.

## What's the fix?
- Added WebSocket client.
- Added listenConfrimTransaction method listener.
- Implement web socket in the `helper.setupSnap` and `open()` socket when the network changes.
- Store socket in context state,
- Implementation of listenConfrimTransaction in the transaction table component.

### DIFF
--- a/snap/frontend/config/index.js
+++ b/snap/frontend/config/index.js
@@ -6,3 +6,7 @@ export const explorerUrl = {
 	mainnet: 'https://symbol.fyi',
 	testnet: 'https://testnet.symbol.fyi'
 };
+
+export const Channels = {
+	confirmedAdded: 'confirmedAdded'
+};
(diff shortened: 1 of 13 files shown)


snap/frontend/config/index.js
```diff
@@ -6,3 +6,7 @@ export const explorerUrl = {
 	mainnet: 'https://symbol.fyi',
 	testnet: 'https://testnet.symbol.fyi'
 };
+
+export const Channels = {
+	confirmedAdded: 'confirmedAdded'
```
Wayonb, line 11: Do you also plan to support cosigning transactions?

Wayonb (approved):
👍🏾

snap/frontend/config/index.js
AnthonyLaw: yes, but no now. I will do the basic release first to support the transfer transaction only

PR #1064, opened 2024-08-19
[snap/frontend]: Pulling latest account mosaic when changing account

AnthonyLaw:
## What was the issue?
- Account mosaic is not updated when switching to another address.

## What's the fix?
- Always query the latest mosaic balance as long as the user changes the address.


Wayonb (approved):
👍🏾

PR #1066, opened 2024-08-22
[snap]: Implemented unconfirmed transactions

AnthonyLaw:
## What was the issue?
- Missing unconfirmed transaction display in table.

## What's the fix?
- Added unconfirmed transaction listener in WebSocket.
- Added transaction group into `fetchAccountTransactions` method param.
- Added `updateUnconfirmedTransactions` in helper, which updates unconfirmed transactions in the state.
- Implement unconfirmed transactions in transaction table component.

### DIFF
--- a/snap/backend/src/utils/transactionUtils.js
+++ b/snap/backend/src/utils/transactionUtils.js
@@ -18,11 +18,11 @@ const transactionUtils = {
      */
 	async fetchAccountTransactions({ state, requestParams }) {
 		const { network } = state;
-		const { address, offsetId } = requestParams;
+		const { address, offsetId, group } = requestParams;
 
 		const client = symbolClient.create(network.url);
 
-		const transactions = await client.fetchTransactionPageByAddress(address, offsetId);
+		const transactions = await client.fetchTransactionPageByAddress(address, offsetId, group);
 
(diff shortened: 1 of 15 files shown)


snap/backend/src/utils/transactionUtils.js
```diff
@@ -90,7 +90,10 @@ const transactionUtils = {
 	createTransaction(id, meta, transaction, network, transactionTypeOverride = null) {
 		const facade = new SymbolFacade(network.networkName);
 		const senderAddress = facade.network.publicKeyToAddress(new PublicKey(transaction.signerPublicKey)).toString();
-		const date = moment.utc(facade.network.toDatetime({ timestamp: BigInt(meta.timestamp) })).format('YYYY-MM-DD HH:mm:ss');
+
+		const date = meta && meta.timestamp
```
Wayonb, line 94: why would meta be undefined?

Wayonb (approved):
👍🏾

snap/backend/src/utils/transactionUtils.js
AnthonyLaw: only the timestamp will undefine as an unconfirmed transaction

PR #1067, opened 2024-08-23
[snap/frontend]: fix mosaic info is not load correctly.

AnthonyLaw:
## What was the issue?
- An error occurs when the user creates or imports an account with an existing mosaic, and the mosaic info in the snap state is not current.

## What's the fix?
- Refactor AssetList component, join `renderMosaicNameOrId` and `renderMosaicDetails` method in `renderMosaicDetailsWithName` method
- Query the latest mosaic Info after creating / import the account


Wayonb (approved):
👍🏾

PR #1068, opened 2024-08-23
[snap/frontend]: bug fix return transaction hash

AnthonyLaw:
## What was the issue?
- After approving the transactions, the transfer transaction modal box didn't close, the reason is that after signing the transaction, it didn't return the transaction hash.

## What's the fix?
- Return transaction hash after signing the transaction in helper.


Wayonb (approved):
👍🏾

PR #1069, opened 2024-08-24
Snap/backend configuration

AnthonyLaw:
## What was the issue?
- Update message for [publishing snap](https://docs.metamask.io/snaps/how-to/publish-a-snap/).

## What's the fix?
- update manifest info
- replace the icon.

Screenshot 2024-08-24 at 5 39 21 PM


snap/backend/snap.manifest.json
```diff
@@ -1,18 +1,18 @@
 {
   "version": "0.1.0",
-  "description": "Symbol snap.",
+  "description": "Manage Symbol mosaics, and connect to Symbol apps with MetaMask.",
```
Wayonb, line 3: What Symbol apps can be connected to? 🤔

Wayonb (changes requested):

snap/backend/snap.manifest.json
AnthonyLaw: In my vision, in the future it may have a lot of different Symbol apps can be create.
So in example, the wallet (frontend) can consider as Symbol apps.

In future we can create many different apps, such as a website can sell NFT.
The apps will asking `snap` transfer fund to creators as purchase NFT.

Wayonb (approved):
👍🏾

PR #1080, opened 2024-08-27
[Snap]: Add currecy mosaic in asset list as default.

AnthonyLaw:
## What was the issue?
- The asset list appears empty when the account doesn't have a mosaic.
- Added a currency mosaic to the asset list with an amount of 0 if the account lacks a currency mosaic.

## What's the fix?
- Query currency mosaic info when during the `initial setup`
- Added currency mosaic in asset list, if user account lacks of currency mosaic.

Screenshot 2024-08-28 at 2 15 46 AM

### DIFF
--- a/snap/backend/src/index.js
+++ b/snap/backend/src/index.js
@@ -32,6 +32,7 @@ export const onRpcRequest = async ({ request }) => {
 		await networkUtils.switchNetwork(apiParams);
 		await priceUtils.getPrice(apiParams);
 		await transactionUtils.getFeeMultiplier(apiParams);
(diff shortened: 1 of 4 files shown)


Wayonb (approved):
👍🏾 

Would it be better to say there is no token to show?
My only concern is that it does not match an account with actually 0 xym

AnthonyLaw:
> 👍🏾
> 
> Would it be better to say there is no token to show? My only concern is that it does not match an account with actually 0 xym

I have the same thought, but our desktop wallet displays 0 XYM for the newly generated wallet.

PR #1081, opened 2024-08-30
[Snap]: Apply live transaction fees calculator.

AnthonyLaw:
## What was the issue?
- For better validation transaction fees, it should have fee calculation and keep updated when the user inputs some data such as a message.

## What's the fix?
- Added fee calculation, it displays fees in the transfer transaction form.
- Fees validation to ensure the account has enough balance to do transactions.
- Refactor `signTransferTransaction` method, and replace `feeMultiplierType` with `fees` because the calculation is done by frontend.

image


Wayonb (approved):
👍🏾

PR #1087, opened 2024-09-03
[snap/backend]: bug fix filter account private key and network

AnthonyLaw:
## What was the issue?
- Users cannot import the same private key in different networks.

## What's the fix?
- Validate filter private key and network.


Wayonb (approved):
👍🏾

Issue #1090, opened 2024-09-05
About metal decode.


0x070696E65:
If I write the metal config in `rest.json`, it works fine, but if I change the state with `upgrade`, it is removed from `rest.json`.

Is it possible to describe it in `override.ini` so that it is not deleted when upgrading?

Jaguar0625:
currently, there is `--metadata` argument that allows user to specify contents of `rest.config!nodeMetadata` only.

i propose to deprecate this options and replace it with `--rest-overrides`, which will allow user to override (and add) any and all settings in the generated rest.config.

please see PoC in commit [86e2418](https://github.com/symbol/product/commit/86e2418859d66ee041634960a17bb8285d0def60)

let us know if this will be acceptable to you 🙏

0x070696E65:
I think its a good suggestion.

I understand that all override settings related to REST can be specified with that option.

I guess there are a certain number of node administrators who forget to put that option on when upgrading (and had set it once in the past), but we don't need to care about.

Jaguar0625:
Added in [cc6270b](https://github.com/symbol/product/commit/cc6270b938dc0eba6d661eeedf509b290df08f46).

PR #1091, opened 2024-09-05
[tools/shoestring]: add support for advanced customization of rest.json

Jaguar0625:
problem: shoestring only allows user to specify nodeMetadata property of rest.config
    solution: allow user to provide differential config file that overrides defaults (similar to REST native support)
    
       issue: #1090

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,6 +23,7 @@ _data/
 .next/
 
 # python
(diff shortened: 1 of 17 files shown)


Wayonb (approved):
👍🏾

PR #1099, opened 2024-09-10
[snap/frontend]: Refactor on snap factory, extract invoke snap method

AnthonyLaw:
## What was the issue?
- Duplicated invoke a snap method in the snap factory.

## What's the fix?
- extract the invoke method.

### DIFF
--- a/snap/frontend/utils/snap.js
+++ b/snap/frontend/utils/snap.js
@@ -5,6 +5,24 @@ const symbolSnapFactory = {
 		return {
 			provider: provider,
 			/**
+			 * A helper function to invoke a method in the snap.
+			 * @param {string} method - The method to invoke.
+			 * @param {object} params - The params to pass with the method.
+			 * @returns {object} The response from the snap.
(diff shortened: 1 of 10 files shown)


snap/frontend/utils/snap.js
```diff
@@ -4,6 +4,24 @@ const symbolSnapFactory = {
...
+			 * @returns {object} The response from the snap.
+			 */
+			async invokeSnapMethod(method, params = {}) {
+				return await this.provider.request({
+					method: 'wallet_invokeSnap',
+					params: {
+						snapId: defaultSnapOrigin,
+						request: {
+							method,
+							...(Object.keys(params).length && { params })
```
Wayonb, line 20: Do you need the ``keys().length`` check? If the params are empty, then nothing gets copied.

AnthonyLaw: yes, if `params` is empty, I will skip the `params`.

PR #1100, opened 2024-09-11
[snap/frontend]: Added react-toastify (toast)

AnthonyLaw:
## What was the issue?
- All the errors will be thrown in console logs (browser), and adding a toast notification, will improve user experience.

## What's the fix?
- Added react-toastify library.
- Handle error messages from the snap factory.
- Handle successful messages: detect for announce/pending/confirmed transactions


Wayonb (approved):
👍🏾

PR #1173, opened 2024-11-12
[tools/shoestring] fix: renew certs uses generates a new ca cert

Wayonb:
problem: renew certificates command creates a new ca cert without renew-ca option
solution: only generate a new ca cert with the renew-ca option

### DIFF
--- a/tools/shoestring/shoestring/commands/renew_certificates.py
+++ b/tools/shoestring/shoestring/commands/renew_certificates.py
@@ -17,7 +17,10 @@ async def run_main(args):
 
 	openssl_executor = OpensslExecutor(os.environ.get('OPENSSL_EXECUTABLE', 'openssl'))
 	with CertificateFactory(openssl_executor, ca_key_path, config.node.ca_password) as factory:
-		factory.generate_ca_certificate(config.node.ca_common_name)
+		if args.renew_ca:
+			factory.generate_ca_certificate(config.node.ca_common_name)
+		else:
+			factory.reuse_ca_certificate(config.node.ca_common_name, directories.certificates)
 		factory.generate_random_node_private_key()
 		factory.generate_node_certificate(config.node.node_common_name)
 		factory.create_node_certificate_chain()
--- a/tools/shoestring/shoestring/commands/renew_certificates.py
+++ b/tools/shoestring/shoestring/commands/renew_certificates.py
@@ -19,8 +19,9 @@ async def run_main(args):
 	with CertificateFactory(openssl_executor, ca_key_path, config.node.ca_password) as factory:
 		if args.renew_ca:
 			factory.generate_ca_certificate(config.node.ca_common_name)
-		else:
+		else:  # health node full certificate check needs current ca cert to pass
 			factory.reuse_ca_certificate(config.node.ca_common_name, directories.certificates)
+
 		factory.generate_random_node_private_key()
 		factory.generate_node_certificate(config.node.node_common_name)
 		factory.create_node_certificate_chain()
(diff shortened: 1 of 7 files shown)


tools/shoestring/shoestring/internal/CertificateFactory.py
```diff
@@ -106,6 +106,20 @@ def generate_ca_certificate(self, ca_cn, days=7300):
 		with open('index.txt', 'wt', encoding='utf8') as outfile:
 			outfile.write('')
 
+	def reuse_ca_certificate(self, ca_cn, ca_cert_path):
```
Jaguar0625, line 109: please add test for this new function

tools/shoestring/shoestring/commands/renew_certificates.py
```diff
@@ -17,7 +17,10 @@ async def run_main(args):
 
 	openssl_executor = OpensslExecutor(os.environ.get('OPENSSL_EXECUTABLE', 'openssl'))
 	with CertificateFactory(openssl_executor, ca_key_path, config.node.ca_password) as factory:
-		factory.generate_ca_certificate(config.node.ca_common_name)
+		if args.renew_ca:
+			factory.generate_ca_certificate(config.node.ca_common_name)
+		else:
+			factory.reuse_ca_certificate(config.node.ca_common_name, directories.certificates)
```
Jaguar0625, line 23: add comment, that this is needed to pass health full certificate check
also,  generally prefer a newline after end of if/else block (for readability)

tools/shoestring/shoestring/lang/messages.pot
```diff
@@ -11,7 +11,7 @@ msgid "argument-help-announce-transaction-transaction"
 msgstr ""
```
Jaguar0625, line 11: since you didn't add any messages, you should be able to revert lang changes (only changed line numbers in comments), but up to you

Jaguar0625 (commented):
👍

tools/shoestring/shoestring/lang/messages.pot
Wayonb: yea can revert, wasn't sure if the line number change would cause an issue.

tools/shoestring/tests/internal/test_CertificateFactory.py
```diff
@@ -269,6 +269,29 @@ def test_cannot_generate_node_certificate_without_common_name(self):
...
+					factory.package(package_directory)
+
+					# Act:
+					with CertificateFactory(self._create_executor(), ca_private_key_path) as renew_factory:
+						renew_factory.reuse_ca_certificate(ca_common_name, Path(package_directory))
+
+						# Assert:
+						assert Path('ca.crt.pem').exists()
+						assert Path('ca.cnf').exists()
+						assert ca_certificate_path.exists()
```
Jaguar0625, line 293: i don't think this test is correct because i believe thes would be created by (first) CertificateFactory. to emulate real behavior, i think you want two temporary directories one for each CertificateFactory

Wayonb: hmm..I will take a look but I thought the second CertificateFactory would create another temp folder? 🤔

Wayonb: In which case the ``factory.reuse_ca_certificate`` would fail to create the ``new_certs`` folder since its already present.  but will move the second CertificateFactory for clarity

Jaguar0625 (approved):
👍

tools/shoestring/tests/internal/test_CertificateFactory.py
Wayonb: Ok.  Will add a sanity check to verify that the first CertificateFactory folder was removed.

``ca_certificate_filepath`` will exist since this is where the certs were install from the first CertificateFactory

In the second CertificateFactory, the ca cert is copied from the ``ca_certificate_filepath``

PR #1174, opened 2024-11-13
[tools/shoestring]: add support for websockets 14

Wayonb:
problem: serve() moved to the asyncio.server
solution: update import to use new path


Jaguar0625 (approved):
👍

PR #1204, opened 2024-12-03
[explorer/nodewatch] fix: update the catapult compatible matrix

Wayonb:
problem: nodewatch does not show the latest version as green.
         plus incompatible catapult are not marked as red.
solution: update nodewatch to show the correct colors


Jaguar0625 (approved):
👍

PR #1258, opened 2025-02-07
[tools/shoestring] fix: disable too many lines plus update sai network file.

Wayonb:
problem: pylint error due to too many lines
               missing rest-light.json file in sai package
solution: disable pylint error
               add missing file.


AnthonyLaw (approved):
👍🏼

PR #1266, opened 2025-02-14
[tools/shoestring]: add recovery tool

Wayonb:
problem: There is no way to recover from lock files in the data folder
solution: create a docker-compose-recovery.yaml which will run the recovery tool

### DIFF
--- a/tools/shoestring/README.md
+++ b/tools/shoestring/README.md
@@ -257,6 +257,13 @@ When `--purge-harvesters` is set, delegates discovered using old keys will be di
 
 ## Files
 
+### docker-compose-recovery.yaml
+
+The docker-compose-recovery.yaml file is used to recovery your node when there is a `server.lock` or `broker.lock` 
+present in the `data` folder after stopping your node.
+
+`docker compose -f docker-compose-recovery.yaml up --abort-on-container-exit`
+
 ### Shoestring Configuration INI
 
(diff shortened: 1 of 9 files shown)


tools/shoestring/README.md
```diff
@@ -257,6 +257,13 @@ When `--purge-harvesters` is set, delegates discovered using old keys will be di
 
 ## Files
 
+### docker-compose-recovery.yaml
+
+The docker-compose-recovery.yaml file is used to recovery your node when there is a `server.lock` or `broker.lock`
```
Jaguar0625, line 262: used to recover~~y~~

but this should be in its own Recovery/Troubleshooting section 

Files section is discussing input files

tools/shoestring/shoestring/templates/docker-compose-dual.yaml
```diff
@@ -33,7 +33,11 @@ services:
       options:
         max-size: 25M
         max-file: '10'
+  {%- if recovery %}
```
Jaguar0625, line 36: indenting of if blocks is inconsistent, in some places aligned with inner text in other places inner text is indented

Jaguar0625 (commented):
👍

AnthonyLaw (approved):
👍🏼

Jaguar0625 (approved):
👍

PR #1276, opened 2025-02-19
[tools/shoestring] fix: upgrade to MongoDb 7

Wayonb:


tools/shoestring/shoestring/templates/docker-compose-dual.yaml
```diff
@@ -2,7 +2,7 @@
 version: '2'
 services:
   db:
-    image: mongo:6.0.14
+    image: mongo:7.0.16
```
AnthonyLaw, line 5: In another project, you upgrade to `mongo:8`.
Any reason to use `mongo:7` here?

Wayonb: No, for a couple of reasons

1. Mainnet is currently using MongoDB 6, and MongoDB only allows upgrades to the next version, which is 7.
2. We have not tested MongoDB 8 yet. The other PR is to upgrade tests to start testing MongoDB 8.

Wayonb: > Any reason to use mongo:7 here?

I used a specific version of MongoDB to ensure each node has the same version.

AnthonyLaw (approved):
👍🏼

PR #1283, opened 2025-02-26
[tools/shoestring] fix: update mainnet package hash

Wayonb:
problem: mainnet package hash needs to be updated for each release
solution: update mainnet package hash for catapult 1.0.3.8 release


Jaguar0625 (approved):
👍

PR #1284, opened 2025-02-27
[explorer/nodewatch]: added new endpoint to query node by node public key

AnthonyLaw:
Problem: The endpoint is missing when using the node public key to query node info.
Solution: 
- Added a new endpoint to query node info by the node public key.


Wayonb (approved):
👍🏾

PR #1285, opened 2025-02-28
[lightapi/python] fix: disable RFC5280 validation

Wayonb:
problem: RFC5280 validation is enabled by default in Python 3.13 which causes cert validation failure.
solution: disable RFC5280 validation since our certs are X509 compliant and RFC5280 is not needed.


Jaguar0625 (approved):
👍

PR #1303, opened 2025-03-12
[explorer/nodewatch]: Added query params for get node list endpoint

AnthonyLaw:
Problem: The filter node is missing on the API node list endpoint.
Solution: Added query params `only_ssl`, `limit`, and `order`.

`only_ssl`: to filter out SSL enable node, it needs explorer and wallet.
`limit`: the number of numbers returned from the endpoint.
`order`: shuffle node list when required

### DIFF
--- a/explorer/nodewatch/nodewatch/RoutesFacade.py
+++ b/explorer/nodewatch/nodewatch/RoutesFacade.py
@@ -73,19 +73,16 @@ class BasicRoutesFacade:
 		"""Returns all nodes with condition."""
 
 		role = kwargs.get('role')
-		exact_match = kwargs.get('exact_match')
+		exact_match = kwargs.get('exact_match', False)
 
 		limit = kwargs.get('limit')
 		only_ssl = kwargs.get('only_ssl')
 		order = kwargs.get('order')
 
 		def custom_filter(descriptor):
-			if role is not None:
-				role_condition = role == descriptor.roles if exact_match else role == (role & descriptor.roles)
-				if not role_condition:
-					return False
+			role_condition = role == descriptor.roles if exact_match else role == (role & descriptor.roles)
 
-			return descriptor.is_ssl_enabled if only_ssl else True
+			return role_condition and (descriptor.is_ssl_enabled if only_ssl else True)
 
 		nodes = list(map(
 			lambda descriptor: descriptor.to_json(),
--- a/explorer/nodewatch/nodewatch/RoutesFacade.py
+++ b/explorer/nodewatch/nodewatch/RoutesFacade.py
@@ -1,5 +1,6 @@
 import asyncio
 import datetime
+import random
 
 from zenlog import log
 
@@ -68,15 +69,35 @@ class BasicRoutesFacade:
 			'explorer_endpoint': self.explorer_endpoint
 		})
 
-	def json_nodes(self, role, exact_match=False):
-		"""Returns all nodes with matching role."""
+	def json_nodes(self, **kwargs):
+		"""Returns all nodes with condition."""
 
-		def role_filter(descriptor):
-			return role == descriptor.roles if exact_match else role == (role & descriptor.roles)
+		role = kwargs.get('role')
+		exact_match = kwargs.get('exact_match')
 
-		return list(map(
+		limit = kwargs.get('limit')
+		only_ssl = kwargs.get('only_ssl')
+		order = kwargs.get('order')
+
+		def custom_filter(descriptor):
+			role_condition = True
+
+			if role is not None:
+				role_condition = role == descriptor.roles if exact_match else role == (role & descriptor.roles)
+
+			if only_ssl:
+				return role_condition and descriptor.is_ssl_enabled
+
+			return role_condition
+
+		nodes = list(map(
 			lambda descriptor: descriptor.to_json(),
-			filter(role_filter, self.repository.node_descriptors)))
+			filter(custom_filter, self.repository.node_descriptors)))
+
+		if order == 'random':
+			random.shuffle(nodes)
+
+		return nodes if limit == 0 else nodes[:limit]
 
 	def json_node(self, filter_field, public_key):
 		"""Returns a node with matching main or node public key."""
--- a/explorer/nodewatch/nodewatch/RoutesFacade.py
+++ b/explorer/nodewatch/nodewatch/RoutesFacade.py
@@ -69,10 +69,9 @@ class BasicRoutesFacade:
 			'explorer_endpoint': self.explorer_endpoint
 		})
 
-	def json_nodes(self, **kwargs):
+	def json_nodes(self, role, **kwargs):
 		"""Returns all nodes with condition."""
 
-		role = kwargs.get('role')
 		exact_match = kwargs.get('exact_match', False)
 
 		limit = kwargs.get('limit')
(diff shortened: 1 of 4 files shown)


explorer/nodewatch/nodewatch/RoutesFacade.py
```diff
@@ -68,15 +69,35 @@ def html_nodes(self):
...
-		return list(map(
+		limit = kwargs.get('limit')
+		only_ssl = kwargs.get('only_ssl')
+		order = kwargs.get('order')
+
+		def custom_filter(descriptor):
+			role_condition = True
+
+			if role is not None:
+				role_condition = role == descriptor.roles if exact_match else role == (role & descriptor.roles)
```
Wayonb, line 86: If the roles does not match, it seems you can return ``False``

explorer/nodewatch/nodewatch/RoutesFacade.py
```diff
@@ -68,15 +69,35 @@ def html_nodes(self):
...
+
+			if only_ssl:
+				return role_condition and descriptor.is_ssl_enabled
```
Wayonb, line 89: Should be able to do this ``return descriptor.is_ssl_enabled  if only_ssl else True``

explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -91,6 +91,17 @@ def symbol_summary():  # pylint: disable=unused-variable
 		template_name, context = symbol_routes_facade.html_summary()
 		return render_template(template_name, **context)
 
+	def _get_json_nodes(role, exact_match, request_args):
+		only_ssl = None
+		if 'only_ssl' in request_args:
+			only_ssl = True
```
Wayonb, line 97: only_ssl = True if 'only_ssl' in request_args else None

explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -91,6 +91,17 @@ def symbol_summary():  # pylint: disable=unused-variable
...
+
+		order = request_args.get('order', None)
+
+		limit = int(request_args.get('limit', 0))
+
+		return jsonify(symbol_routes_facade.json_nodes(role=role, exact_match=exact_match, only_ssl=only_ssl, limit=limit, order=order))
+
```
Wayonb, line 104: It seems ``role`` is always a required arg?

Wayonb (changes requested):

explorer/nodewatch/nodewatch/__init__.py
AnthonyLaw: Yes, you are right.
It filters the `/api/symbol/nodes/peer` and `/api/symbol/nodes/api` endpoint.

explorer/nodewatch/nodewatch/RoutesFacade.py
```diff
@@ -68,15 +69,32 @@ def html_nodes(self):
 			'explorer_endpoint': self.explorer_endpoint
 		})
 
-	def json_nodes(self, role, exact_match=False):
-		"""Returns all nodes with matching role."""
+	def json_nodes(self, **kwargs):
```
Wayonb, line 72: Is the ``role`` required here also?

explorer/nodewatch/nodewatch/RoutesFacade.py
```diff
@@ -68,15 +69,32 @@ def html_nodes(self):
...
+		only_ssl = kwargs.get('only_ssl')
+		order = kwargs.get('order')
+
+		def custom_filter(descriptor):
+			if role is not None:
+				role_condition = role == descriptor.roles if exact_match else role == (role & descriptor.roles)
+				if not role_condition:
+					return False
+
+			return descriptor.is_ssl_enabled if only_ssl else True
```
Wayonb, line 88: Should we be able to do this?
``role_condition and (descriptor.is_ssl_enabled if only_ssl else True)``

Wayonb (changes requested):

explorer/nodewatch/nodewatch/RoutesFacade.py
AnthonyLaw: yes, it filters out the API or Peer node endpoint.

https://github.com/symbol/product/blob/9afd54889c3dac205e900e82dcf5c32e46b04b03/explorer/nodewatch/nodewatch/__init__.py#L117

https://github.com/symbol/product/blob/9afd54889c3dac205e900e82dcf5c32e46b04b03/explorer/nodewatch/nodewatch/__init__.py#L121

explorer/nodewatch/nodewatch/RoutesFacade.py
AnthonyLaw: I did the updates about the `custom_filter` method.
As `role` is always required, which check `role=none` is pointless.

explorer/nodewatch/nodewatch/RoutesFacade.py
Wayonb: I was asking if json_nodes should be defined as this? ``def json_nodes(self, role, **kwargs):``

AnthonyLaw: I see, it makes sense! 
I updated

explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -91,6 +91,15 @@ def symbol_summary():  # pylint: disable=unused-variable
 		template_name, context = symbol_routes_facade.html_summary()
 		return render_template(template_name, **context)
 
+	def _get_json_nodes(role, exact_match, request_args):
+		only_ssl = True if 'only_ssl' in request_args else None
```
Wayonb, line 95: Is there a reason you are not checking if ``only_ssl`` is true?  vs it is in the query list 🤔

Wayonb (changes requested):

Wayonb (approved):
👍🏾

explorer/nodewatch/nodewatch/__init__.py
AnthonyLaw: yes, as the variable is named `only_ssl`, we can consider the SSL to be true.

the query will be like this:

`/api/symbol/nodes/peer?only_ssl`

Wayonb: Yes, but what if some use this without knowing the API?
``/api/symbol/nodes/peer?only_ssl=false``

AnthonyLaw: make sense. let me make some improvement for that.

explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -91,6 +91,18 @@ def symbol_summary():  # pylint: disable=unused-variable
 		template_name, context = symbol_routes_facade.html_summary()
 		return render_template(template_name, **context)
 
+	def _get_json_nodes(role, exact_match, request_args):
+		only_ssl = request_args.get('only_ssl')
+
+		if only_ssl is not None:
+			only_ssl = only_ssl == '' or only_ssl.lower() == 'true'
```
Wayonb, line 98: Should be able to rewrite 
``only_ssl = request_args.get('only_ssl', 'false').lower() in ('', 'true')``

Wayonb (changes requested):

Wayonb (approved):
👍🏾

PR #1305, opened 2025-03-13
[explorer/frontend] fix: sharp missing in production

OlegMakarenko:
## Problem

The following error in console:
```
⨯ Error: 'sharp' is required to be installed in standalone mode for the image optimization to function correctly. Read more at: https://nextjs.org/docs/messages/sharp-missing-in-production
```

## Solution
Installed `sharp`.


Wayonb (approved):
👍🏾

PR #1309, opened 2025-03-17
[explorer/nodewatch]: Added geo location in node descriptor

AnthonyLaw:
Problem: missing geo-location information in the node list.
Solution: load the **geo location** file from the specific path, and create a mapping for node.


explorer/nodewatch/nodewatch/NetworkRepository.py
```diff
@@ -107,6 +110,34 @@ def __init__(self, account_dict):
 		self.version = account_dict['version']
 
 
+class GeoLocationDescriptor:
```
Wayonb, line 113: Is this class needed to filter the geo-location info?

Wayonb (changes requested):

explorer/nodewatch/nodewatch/NetworkRepository.py
AnthonyLaw: nope, this class is just an instance.

Wayonb: Is the class needed since this should also work? Or am I missing something? 🤔 
``self.geo_location_map[entry['host']] = entry['geolocation']``

AnthonyLaw: yeah, it will look like this.

```py
if 'query' in entry['geolocation']:
    del entry['geolocation']['query']

self.geo_location_map[entry['host']] =  entry['geolocation']
```

Wayonb (approved):
👍🏾

PR #1322, opened 2025-03-24
[tools/shoestring] fix: upgrade command fails with harvesting node

Wayonb:
problem: the config patch operation fails due to an invalid path in the config-harvesting.properties file.
                 the patch() method only requires the filename.
solution: pass only the filename to the configuration manager patch method

```
  File "/lib/python3.12/site-packages/shoestring/internal/ConfigurationManager.py", line 48, in patch
    with open(config_filepath, 'rt', encoding='utf8') as infile:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'userconfig/resources/userconfig/resources/config-harvesting.properties'
```

The reason why tests are passing is that the [absolute path](https://docs.python.org/3/library/pathlib.html#operators) is used.


Jaguar0625 (approved):
👍

PR #1324, opened 2025-03-26
[explorer/nodewatch]: Added time series nodes count endpoint

AnthonyLaw:
Problem: The nodes count endpoint is missing.
Solution: Read the JSON file `symbol_time_series_nodes_count` and allow query data from the new endpoint.


Wayonb (approved):
👍🏾

PR #1343, opened 2025-04-08
[tools/shoestring] fix: package hash and bump verion

Wayonb:


tools/shoestring/CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ The changelog format is based on [Keep a Changelog](https://keepachangelog.com/e
 
 ## next
 
-## [0.2.0] - 25 Feb-2025
+## [0.2.1] - 08 Apr-2025
```
AnthonyLaw, line 8: Should we just keep it as `0.2.0` because we never do any release after 0.1.3 ?

Wayonb: Sorry, I just noticed I did not add the problem description.

The issue is that I pushed 0.2.0 yesterday, then realized the package hash did not match since I renamed the property node_key -> nodeKey.  So I removed 0.2.0 version and bumped the version with the new hash.

AnthonyLaw (approved):
👍🏼

PR #1349, opened 2025-04-14
[explorer/nodewatch] fix: update the catapult compatible matrix

Wayonb:
problem: NodeWatch does not show the latest versions of NIS and Catapult as green.
solution: Update NodeWatch to show the correct colors


Jaguar0625 (approved):
👍

PR #1382, opened 2025-05-14
[explorer/nodewatch]: Added finalized properties in node descriptor

AnthonyLaw:
Problem: missing `finalizedEpoch`, `finalizedHash`, and `finalizedPoint` in node descriptor
Solution: added finalized properties in node descriptor


Wayonb (approved):
👍🏾

PR #1407, opened 2025-06-02
[lightapi/python] fix: pytest asyncio removed event loop

Wayonb:
problem: in pytest asyncio v1 the event loop was removed
solution: remove the dependency on event loop and convert fixures and tests to async where needed


Jaguar0625 (approved):
👍

PR #1409, opened 2025-06-02
[faucet/frontend] fix: linting error on import/extensions for the SDK

Wayonb:
problem: missing file extension "ts" for "symbol-sdk/symbol"  import/extensions
solution: disable the import extension for TS


Jaguar0625 (approved):
👍

PR #1420, opened 2025-06-12
[tools/shoestring]: Japanese Language Support

ccHarvestasya:
## Problem

- Japanese translation files exist, but all messages are not set.

## Solution

- Japanese translation files were updated.
- Added frequently used but untranslated messages to the translation file.
- Improved wizard display process to support multi-byte characters (double-byte characters) such as Japanese:
  - Changed tab display character width acquisition from `len()` to `get_cwidth()`, which takes into account the actual display width.
  - To correct a problem with button width calculations being off, we have overridden `prompt_toolkit.widgets.Button` and changed the width calculations to support multibyte characters.
  before: before
  after: after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,6 +29,7 @@ __pycache__/
 build/
 dist/
 *.mo
+.coverage
 
 # other build artifacts
 _generated/
--- a/.gitignore
+++ b/.gitignore
@@ -29,7 +29,6 @@ __pycache__/
 build/
 dist/
 *.mo
-.coverage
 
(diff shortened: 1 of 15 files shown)


Wayonb (approved):
Looks good. 👍🏾 
There might new strings in the dev branch which also needs to be translated 🙏🏾

ccHarvestasya:
Added Japanese translation of new strings in dev.

Wayonb (approved):
👍🏾

PR #1433, opened 2025-06-16
[explorer/nodewatch]: added finalized information in header and voter page

AnthonyLaw:
Problem: Page was missing finalized information and voting key expiry info.
Solution: Added finalized information at the header and voting key expiry column in the voter table.

Before:
image

After:
image

### DIFF
--- a/explorer/nodewatch/nodewatch/templates/base.html
+++ b/explorer/nodewatch/nodewatch/templates/base.html
@@ -34,11 +34,22 @@
           <div class="badge badge-pill badge-success my-0 float-right">
             Symbol Height&nbsp;
             <span id="symbol-block-height">...</span>
+            | Finalized&nbsp;
+            <span id="symbol-finalization-height">...</span>
+            (Cyprus)
+          </div>
+          <br/>
+          <div class="badge badge-pill badge-success my-0 float-right">
+            Epoch&nbsp;
+            <span id="symbol-finalization-epoch">...</span>
           </div>
           <br/>
           <div class="badge badge-pill badge-warning my-0 float-right">
             NEM Height&nbsp;
             <span id="nem-block-height">...</span>
+            | Finalized&nbsp;
+            <span id="nem-finalization-height">...</span>
+            (Harlock)
           </div>
         </div>
       </div>
@@ -70,7 +81,9 @@
         .then(response => response.json())
         .then(response => {
           const height = response['height'];
-          $('#nem-block-height').text(`${height} (Harlock)`);
+          const finalizedHeight = response['finalizedHeight'];
+          $('#nem-block-height').text(`${height}`);
+          $('#nem-finalization-height').text(`${finalizedHeight}`);
           setTimeout(update_nem_height, 30000);
         });
     };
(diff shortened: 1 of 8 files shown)


explorer/nodewatch/tests/test_NetworkRepository.py
```diff
@@ -434,7 +436,7 @@ def test_can_load_symbol_harvester_descriptors(self):
 
 	# endregion
 
-	# region load voter descriptors
+	# region load voter descriptorsestimate_finalized_height
```
Wayonb, line 439: Why change the region description?

explorer/nodewatch/nodewatch/templates/symbol_nodes.html
```diff
@@ -66,10 +67,42 @@ <h2 class="text-center py-4">{{ title }}</h2>
...
+          const currentEpoch = response['epoch'];
+
+          // Update each row's expiry days
+          document.querySelectorAll('tbody tr').forEach(row => {
+            const votingEndEpoch = row.querySelector('.voting-end-epoch').textContent;
+            const expiryCell = row.querySelector('.voting-expiry');
+
+            if (votingEndEpoch && expiryCell) {
+              const blocksPerDay = 2880;
+              const blocksPerEpoch = 1440;
```
Wayonb, line 94: These values can/will be different between testnet and mainnet.

explorer/nodewatch/nodewatch/templates/base.html
```diff
@@ -70,7 +81,9 @@
         .then(response => response.json())
         .then(response => {
           const height = response['height'];
-          $('#nem-block-height').text(`${height} (Harlock)`);
+          const finalizedHeight = response['finalizedHeight'];
+          $('#nem-block-height').text(`${height}`);
+          $('#nem-finalization-height').text(`${finalizedHeight}`);
```
Wayonb, line 86: What is the finalized height for NEM(current height - 360)?

Wayonb (changes requested):

Jaguar0625:
would it be possible to put epoch information on same line as height otherwise header looks a bit too thick?

explorer/nodewatch/nodewatch/templates/base.html
AnthonyLaw: yes

explorer/nodewatch/tests/test_NetworkRepository.py
AnthonyLaw: Good catch! It must be a typo

explorer/nodewatch/nodewatch/templates/symbol_nodes.html
AnthonyLaw: After some thought, `miscellaneous` might not be the best option because we only need the network config once when starting the instance. Instead, I added it into the app config.

here's the commit [e2608f6](https://github.com/symbol/product/pull/1433/commits/e2608f653cc7982a3fe0d934b111d657e3a8b14e)

explorer/nodewatch/nodewatch/templates/symbol_nodes.html
```diff
@@ -66,10 +67,45 @@ <h2 class="text-center py-4">{{ title }}</h2>
...
+          const votingSetGrouping = network_config['votingSetGrouping'];
+
+          // Update each row's expiry days
+          document.querySelectorAll('tbody tr').forEach(row => {
+            const votingEndEpoch = row.querySelector('.voting-end-epoch').textContent;
+            const expiryCell = row.querySelector('.voting-expiry');
+
+            if (votingEndEpoch && expiryCell) {
+              const blocksPerDay = 60 * 60 * 24 / targetBlockGenerationTime;
+              const epochRemaining = votingEndEpoch - currentEpoch;
```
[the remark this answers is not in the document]
AnthonyLaw: > The currentEpoch is the common epoch across the chain?

yes

> If the node is syncing, stuck, or forked, then this information can be misleading since the common epoch will be different from the actual node's epoch

So, in the Nodewatch, we basically read data from a processed JSON file.
To avoid this issue, we should instead read the epoch from a specific "trusted" node? eg: `001-sai-dual`

Wayonb (approved):
👍🏾

explorer/nodewatch/nodewatch/templates/symbol_nodes.html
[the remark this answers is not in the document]
AnthonyLaw: I'll rebase and merge first.

PR #1435, opened 2025-06-19
[lightapi/python]: add NemConnector network_time and announce_transaction

Jaguar0625:
problem: NemConnector does not support sending transactions
solution: add missing functions similar to SymbolConnector


Wayonb (approved):
👍🏾

PR #1438, opened 2025-06-20
[lightapi/python]: ignore deprecation warnings raised by zenlog library

Jaguar0625:
problem: zenlog is calling deprecated warn function on Logger object from logger library
solution: configure pytest to ignore these warnings


Wayonb (approved):
Looks good👍🏾

Should we consider replacing zenlog package?  It seems it has not been updated in awhile

PR #1449, opened 2025-06-23
[lightapi/python]: change BasicConnector to allow 404 to be optionally treated as errors

Jaguar0625:
problem: BasicConnector always treats 404 as success, requiring caller to perform extra validation with 404 is undesirable
solution: treat 404 as error by default and add flag to optionally treat it as success


Wayonb (approved):
👍🏾

PR #1450, opened 2025-06-23
[tools/shoestring] fix: exception handling during hostname validation in the wizard

ccHarvestasya:
## Problem

- When entering a hostname that starts with a dot (e.g. `.example.com`) in the wizard, an exception occurs and a stack trace is displayed.

## Solution

- Although a `UnicodeError` is being raised, it can also be caught as a `ValueError`, so the exception handling is adjusted to catch `ValueError` to match the surrounding code.

### DIFF
--- a/tools/shoestring/shoestring/wizard/ValidatingTextBox.py
+++ b/tools/shoestring/shoestring/wizard/ValidatingTextBox.py
@@ -72,6 +72,8 @@ def is_hostname(value):
 		return True
 	except socket.gaierror:
 		return False
+	except ValueError:
+		return False
 
 
 def is_json(value):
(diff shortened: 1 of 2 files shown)


tools/shoestring/shoestring/wizard/ValidatingTextBox.py
```diff
@@ -72,6 +72,8 @@ def is_hostname(value):
 		return True
 	except socket.gaierror:
 		return False
+	except ValueError:
```
Wayonb, line 75: Should be able to do this
```
except (socket.gaierror, ValueError):
           return False
```

Wayonb (approved):
👍🏾

PR #1451, opened 2025-06-24
[bridge]: parse incoming NEM transactions into wrap requests and errors

Jaguar0625:
problem: NEM transactions need to be processed
solution: add extract_wrap_request_from_transaction


Wayonb (approved):
👍🏾

PR #1452, opened 2025-06-24
[lightapi/python]: explicitly order Symbol transactions in desc order

Jaguar0625:
problem: transactions/confirmed does not seem to be returning transactions in desc order
solution: explicitly specify desc order


Wayonb (approved):
👍🏾

PR #1465, opened 2025-06-30
[bridge]: add Symbol support for downloading wrap requests

Jaguar0625:
problem: step one of bridge requires downloading and validating wrap requests from native network
    solution: add `download_wrap_request_transactions` workflow to
              * download transactions from network
              * map each tranaction into a wrap request or wrap error


Jaguar0625:
merging with #1453

PR #1466, opened 2025-06-30
[lightapi/python]: add `balance`  to NemConnector and SymbolConnector and other fixes

Jaguar0625:
[lightapi/python]: add `balance`  to NemConnector and SymbolConnector
    
     problem: need to retrieve balance of an account
    solution: add balance function in connectors

    [lightapi/python]: standardize naming convention to use *_json
    
     problem: inconsistent naming is being used - *_json json_* _dict* dict_*
    solution: standardize on *_json

    [lightapi/python]: fix race condition in test_can_handle_timeout
    
     problem: sleep task can be destroyed while sleeping leading to a warning
    solution: explicitly cancel the sleep task at the end of the test

### DIFF
--- a/lightapi/python/symbollightapi/connector/NemConnector.py
+++ b/lightapi/python/symbollightapi/connector/NemConnector.py
@@ -129,7 +129,13 @@ class NemConnector(BasicConnector):
 
 	# endregion
 
-	# region GET (account_info)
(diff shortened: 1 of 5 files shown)


Wayonb (approved):
👍🏾

PR #1468, opened 2025-07-01
[bridge]: add new calculate_conversion workflow

Jaguar0625:
problem: need to calculate wrapped token : native token conversion rate
solution: add (temporary) workflow to perform calculation and add missing functions


Jaguar0625:
merging with #1453

PR #1469, opened 2025-07-02
 [bridge]: add support for creating (redemption) transactions

Jaguar0625:
problem: need to be able to create transfer transactions for unwrap/redemption process
    solution: add create_transfer_transaction to nem and symbol NetworkFacade


Jaguar0625:
merging with #1453

PR #1470, opened 2025-07-03
[bridge]: support custom message validation to support custom networks

Jaguar0625:
problem: code assumes destination (wrapped) address is on ethereum network
solution: relax assumption to allow nem <> symbol bridging


Jaguar0625:
merging with #1453

PR #1477, opened 2025-07-07
[lightapi/python]: add a few more needed extensions

Jaguar0625:
[lightapi/python]: add query_block_timestamps
    
     problem: need to lookup timestamps of block heights
    solution: add query_block_timestamps extension
              add block_headers to SymbolConnector

    [lightapi/python]: add filter_finalized_transactions
    
     problem: need to select finalized transactions from a group of transactions
    solution: add filter_finalized_transactions extension
              add filter_confirmed_transactions to SymbolConnector and NemConnector


Wayonb (approved):
👍🏾

PR #1478, opened 2025-07-08
[bridge]: add workflow for checking completed wrap payout transactions

Jaguar0625:
problem: after sending payout transactions need to monitor their completion
solution: add workflow for tracking payout transactions
          rework WrapRequestDatabase to store all required information


Wayonb (approved):
👍🏾

PR #1479, opened 2025-07-08
[bridge]: update network facades to support arbitrary mosaics

Jaguar0625:
[bridge]: update network facades to support arbitrary mosaics
    
     problem: network facades primarily support currency mosaic operations
    solution: generalize to supporting arbitrary mosaics

    [bridge]: fix some naming issues in tests
    
     problem: inconsistent namings should be fixed
    solution: rename can_create_wrap_request_*
              * can(not)_create_wrap_request_* => can(not)_extract_wrap_request_*
              * transaction_meta_json => transaction_with_meta_json

### DIFF
--- a/bridge/bridge/nem/NemUtils.py
+++ b/bridge/bridge/nem/NemUtils.py
@@ -20,15 +20,21 @@ def calculate_transfer_transaction_fee(xem_amount, message=None):
 
 # region extract_wrap_address_from_transaction
 
-def _process_transfer_transaction(is_valid_address, transaction_identifier, transaction_json):
+def _process_transfer_transaction(is_valid_address, mosaic_id, transaction_identifier, transaction_json):
 	amount = 0
 	transaction_version = transaction_json['version'] & 0x00FFFFFF
 	if 1 == transaction_version or 0 == len(transaction_json['mosaics']):
+		if mosaic_id:
+			return make_wrap_error_result(
+				transaction_identifier,
+				f'nem:xem is not supported, expected mosaic {mosaic_id[0]}:{mosaic_id[1]}')
+
 		amount = transaction_json['amount']
 	else:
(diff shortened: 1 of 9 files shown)


bridge/bridge/nem/NemUtils.py
```diff
@@ -20,15 +20,21 @@ def calculate_transfer_transaction_fee(xem_amount, message=None):
...
 
-def _process_transfer_transaction(is_valid_address, transaction_identifier, transaction_json):
+def _process_transfer_transaction(is_valid_address, mosaic_id, transaction_identifier, transaction_json):
 	amount = 0
 	transaction_version = transaction_json['version'] & 0x00FFFFFF
 	if 1 == transaction_version or 0 == len(transaction_json['mosaics']):
+		if mosaic_id:
+			return make_wrap_error_result(
+				transaction_identifier,
+				f'mosaic {mosaic_id[0]}:{mosaic_id[1]} must be transferred in mosaic bag')
```
Wayonb, line 30: Should there be a different error for when the transaction version is 1?  I guess similar to the error above 
``raise ValueError('version one does not support custom mosaics')``

But I assume this will only happen for users who create the transfer tx.

Wayonb (approved):
👍🏾

bridge/bridge/nem/NemUtils.py
Jaguar0625: this is for a bridge address that accepts a custom mosaic. this is error registered because it was sent native XEM.

Wayonb: Yes, the error message could be clearer, as I was unsure what a 'mosaic bag' meant without looking at the code. 😄 

 What you said above is better.

``Native XEM is not supported``

Wayonb (approved):
👍🏾

PR #1480, opened 2025-07-11
[explorer/frontend] feat: support runtime environment variables in docker container

OlegMakarenko:
## Problem  
Previously, environment variables needed to be baked in at build time using the `NEXT_PUBLIC_` prefix to be available to the client side. This approach meant that environment-specific values couldn't be modified at runtime when running the Docker container, making deployments less flexible and preventing separation between build and runtime configuration.

## Solution  
- Removed the need for the `NEXT_PUBLIC_` prefix from environment variables, making them accessible only on the server side.  
- During server-side rendering, the app now injects an environment variables object directly into the rendered HTML page so the client side can read them immediately before React hydration.


extpmielcarekadeo (approved):
Tried locally and it works :tada:

explorer/frontend/pages/_app.jsx
```diff
@@ -44,6 +46,7 @@ const App = ({ Component, pageProps }) => {
 
 	return (
 		<div className={styles.wrapper}>
+			<script dangerouslySetInnerHTML={{ __html: `window.appConfig = ${JSON.stringify(appConfig)};` }} />
```
Wayonb, line 49: I am assuming the ``dangerouslySetInnerHTML`` is safe because it is trusted data? 
https://blog.logrocket.com/using-dangerouslysetinnerhtml-react-application/

OlegMakarenko: Yes, it should be safe enough in this case, since the data is injected into the page by the server and consists only of environment variables provided when running the Docker container.

It would need to be sanitized if it came from a user or any third-party source, to prevent malicious scripts from being injected.

PR #1499, opened 2025-07-23
[bridge]: fix conversion rate calculation

Jaguar0625:
problem: when multiple unwrap transactions are in different wrapped network blocks
          but map to same native network block, calculation fails
solution: unwrap block adjustment should be relative to native, not wrapped network

 problem: conversion rates are calculated wrong when donations are present
solution: sum net issued wrapped tokens instead of deposited native tokens


bridge/bridge/ConversionRateCalculatorFactory.py
```diff
@@ -0,0 +1,48 @@
...
+		timestamp = self._databases.wrap_request.lookup_block_timestamp(native_height)
+		if not timestamp:
+			return None
+
+		required_conditions = [
+			self._databases.balance_change.is_synced_at_height(native_height),
+			self._databases.wrap_request.is_synced_at_timestamp(timestamp),
+			self._databases.unwrap_request.is_synced_at_timestamp(timestamp)
+		]
+		if not all(required_conditions):
```
Wayonb, line 38: Is this condition recoverable after transaction info is downloaded?

Wayonb (approved):
👍🏾

Jaguar0625:
added second fix to this PR (instead of separate one) because it's closely related to original changes

bridge/bridge/db/WrapRequestDatabase.py
```diff
@@ -182,6 +182,20 @@ def cumulative_fees_paid_at(self, timestamp):
 		sum_amount = cursor.fetchone()[0]
 		return sum_amount or 0
 
+	def payout_transaction_hashes_at(self, timestamp):
+		"""Gets cumulative amount of fees paid at or before timestamp."""
```
Wayonb, line 186: The description needs to be updated

PR #1504, opened 2025-07-23
[monorepo] feat: add job to update git submodule

Wayonb:
problem: submodule are not updated.  This leads to lint errors on dev box
solution: add a Jenkins job which runs Tuesday morning after deps has been updated.


Jaguar0625 (approved):
👍

PR #1506, opened 2025-07-25
[bridge]: wait for transaction to be unconfirmed after send

Jaguar0625:
problem: NEM transaction sends fail when they are rejected but Symbol transactions are not
    solution: wait for Symbol transactions to transition to unconfirmed status before continuing
              migrate TransactionSender tests from NEM to Symbol

### DIFF
--- a/bridge/bridge/NetworkUtils.py
+++ b/bridge/bridge/NetworkUtils.py
@@ -2,6 +2,8 @@ from collections import namedtuple
 from decimal import ROUND_UP, Decimal
 
 from symbolchain.CryptoTypes import Hash256, PrivateKey
+from symbollightapi.model.Constants import TimeoutSettings, TransactionStatus
+from symbollightapi.model.Exceptions import NodeException
 
 BalanceChange = namedtuple('BalanceChange', ['address', 'currency_id', 'amount', 'transaction_hash'])
 BalanceTransfer = namedtuple('BalanceTransfer', ['signer_public_key', 'recipient_address', 'amount', 'message'])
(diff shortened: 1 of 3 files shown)


bridge/bridge/NetworkUtils.py
```diff
@@ -68,6 +71,15 @@ async def send_transaction(self, transaction):
 
 		transaction_hash = facade.hash_transaction(transaction)
 		await connector.announce_transaction(transaction)
+
+		is_unconfirmed = await connector.try_wait_for_announced_transaction(
+			transaction_hash,
+			TransactionStatus.UNCONFIRMED,
```
Wayonb, line 77: Not sure if the transaction status here makes sense. There might be a race where the transaction is announced and gets confirmed before the status is checked.  As with the Symbol connector, the confirmed status is also checked.

lightapi/python/symbollightapi/connector/NemConnector.py
```diff
@@ -274,3 +274,24 @@ async def announce_transaction(self, transaction_payload):
...
 
 	# endregion
+
+	# region try_wait_for_announced_transaction
+
+	async def try_wait_for_announced_transaction(self, transaction_hash, desired_status, timeout_settings):
+		"""Tries to wait for a previously announced transaction to transition to a desired status."""
+
+		if TransactionStatus.UNCONFIRMED == desired_status:
+			# announce would fail if a transaction is not unconfirmed
```
Wayonb, line 284: I am assuming the status does not matter if the transaction is already confirmed?

Wayonb (changes requested):

bridge/bridge/NetworkUtils.py
Jaguar0625: it can wait for either UNCONFIRMED or CONFIRMED, so i think it makes sense? unless i misunderstand your point?

Wayonb (approved):
👍🏾

PR #1531, opened 2025-08-07
[bridge]: use checksum addresses or eth libraries will not work

Jaguar0625:
problem: prior to signing, eth_account checks if (string) address matches checksum address
          since string representation used is uppercase, it will not match and signing fails
solution: change EthereumAddress str to return checksum address


Wayonb (approved):
👍🏾

PR #1532, opened 2025-08-07
[bridge]: native and wrapped block metadata needs to be stored in separate tables

Jaguar0625:
1. store native and wrapped block metadata in separate tables
2. capture metadata for blocks with failed transactions too


bridge/tests/db/test_WrapRequestDatabase.py
```diff
@@ -902,11 +922,36 @@ def test_can_lookup_block_height_when_present(self):
...
+		# Arrange:
+		def post_insert_action(database):
+			database.set_payout_block_timestamp(1122, 98765)
+
+		# Act + Assert:
+		self._assert_can_insert_requests([], [], expected_payout_block_metadatas=[
+			(1122, self._symbol_to_unix_timestamp(98765))
+		], post_insert_action=post_insert_action)
+
+	def test_can_overrwrite_payout_block_timestamp(self):
```
Wayonb, line 937: In what case will the payout block be overwritten?  If the payout transaction fails? 🤔

bridge/tests/test_ConversionRateCalculatorFactory.py
```diff
@@ -9,13 +9,13 @@
...
 from .test.BridgeTestUtils import HASHES
-from .test.MockNetworkFacade import MockNetworkFacade
+from .test.MockNetworkFacade import MockNemNetworkFacade
 
 # region test helpers
 
 
 def _create_databases(database_directory):
-	return Databases(database_directory, MockNetworkFacade(), MockNetworkFacade())
+	return Databases(database_directory, MockNemNetworkFacade(), MockNemNetworkFacade())
```
Wayonb, line 18: Should one of these be ``MockSymbolNetworkFacade``?

Wayonb (approved):
👍🏾

bridge/tests/db/test_WrapRequestDatabase.py
Jaguar0625: i don't think this _should_ happen in current workflows, but just wanted to have test to define behavior in case that changes

Wayonb (commented):
👍🏾

PR #1533, opened 2025-08-08
[bridge]: add ETH fee calculation and convert fees into wrapped token units

Jaguar0625:
[bridge]: add price oracle for fee conversions
    
     problem: when sending transactions on wrapped network fees need to be from native currency to wrapped token
    solution: add and integrate CoinGecko connector (price oracle)

    [bridge]: add ethereum transaction fee calculation
    
     problem: ethereum fees are hardcoded
    solution: calculate fees dynamically


Wayonb (approved):
👍🏾

PR #1534, opened 2025-08-08
[bridge]: add flask server and initial endpoints

Jaguar0625:


bridge/bridge/db/Databases.py
```diff
@@ -8,12 +8,13 @@
 class Databases:  # pylint: disable=too-many-instance-attributes
 	"""Container of all databases."""
 
-	def __init__(self, database_directory, native_network, wrapped_network):
+	def __init__(self, database_directory, native_network, wrapped_network, is_read_only=False):
```
Wayonb, line 11: I would set ``is_read_only=True`` by default to prevent any scenario where a new endpoint is added and the database params is not set

Wayonb (approved):
👍🏾

PR #1554, opened 2025-08-14
[bridge]: second round of API fixes and improvements

Jaguar0625:
fix bugs and enhance API behavior

### DIFF
--- a/bridge/bridge/api/Validators.py
+++ b/bridge/bridge/api/Validators.py
@@ -1,0 +1,24 @@
+from symbolchain.CryptoTypes import Hash256
+
+
+def is_valid_address_string(network_facade, address_string):
+	"""Determines if the specified string is a valid address string."""
+
+	(is_valid, _) = network_facade.is_valid_address(address_string)
+	return is_valid
+
+
+def _contains_only_alphabet(alphabet, test_string):
+	return all(ch in alphabet for ch in test_string)
+
+
+def is_valid_hash_string(hash_string):
+	"""Determines if the specified string is a valid hex string."""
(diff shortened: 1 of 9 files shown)


bridge/bridge/api/Validators.py
```diff
@@ -0,0 +1,27 @@
...
+
+
+def _contains_only_alphabet(alphabet, test_string):
+	return all(ch in alphabet for ch in test_string)
+
+
+def is_valid_hash_string(hash_string):
+	"""Determines if the specified string is a valid hex string."""
+
+	return isinstance(hash_string, str) and 2 * Hash256.SIZE == len(hash_string) and _contains_only_alphabet('0123456789ABCDEF', hash_string)
```
AndreaLanfranchi, line 18: AFAIK python uses C case sensitive comparison hence valid hashes in lowercase would not return true. 

Also what about using regular expressions ?

```
pattern = re.compile(rf'^[0-9A-Fa-f]{{{2 * Hash256.SIZE}}}$')
return isinstance(hash_string, str) and bool(pattern.match(hash_string))
```

Jaguar0625: this is not really a general case validation function. it's currently only used to validate transaction hashes that are a parameter of a url route;  e.g 	
```
@app.route('/wrap/requests/<address>/<transaction_hash>')
```

since the path part of a uri is case-sensitive, i generally pick one casing (upper or lower) to avoid having multiple uris pointing to multiple things.

Wayonb (approved):
👍🏾

PR #1589, opened 2025-09-04
[explorer/nodewatch] Openapi - serve openapi using Flask

cryptoBeliever:
## Problem description

Since we moved to docker images openapi should be served directly inside container. Without it we can't expose openapi correctly.

## Solution

Serve openapi using Flask with rest of the API


explorer/nodewatch/nodewatch/__init__.py
```diff
@@ -46,6 +46,10 @@ def create_app():
 	def index():  # pylint: disable=unused-variable
 		return redirect(url_for('symbol_summary'))
 
+	@app.route('/api/openapi')
```
Wayonb, line 49: Can you also write a unit test for this?

Wayonb (changes requested):

explorer/nodewatch/nodewatch/__init__.py
cryptoBeliever: Added test for this endpoint.

PR #1654, opened 2025-09-26
[bridge]: fix log rotation

Jaguar0625:
problem: logs are not rotating
  TimedRotatingFileHandler will only rotate when the last log is more than a day ago
  since scripts are run more frequently than daily, this condition will never be true
solution: switch to RotatingFileHandler
          add additional configuration settings to customize its behavior


Wayonb (approved):
👍🏾

PR #1672, opened 2025-10-01
[bridge]: throttle conversion rate calculations in api

Jaguar0625:
problem: /prepare/wrap calculates conversion rate every call,
          which leads to a lot of external service (e.g. CoinMarketCap) usage
solution: throttle conversion rate calculations every CONVERSION_RATE_REFRESH_SECONDS seconds
          (default 10 minutes)


cryptoBeliever (approved):
:+1:

PR #1674, opened 2025-10-03
[bridge]: check Ethereum transaction status

Jaguar0625:
problem: ethereum processing assumes only successful transactions are returned
solution: check ethereum transaction status and create wrap_error entry for failures


cryptoBeliever (approved):
Works correctly :+1: Failed transactions goes into wrap_errors table with error message "ethereum transaction failed on ethereum blockchain"

Wayonb (approved):
👍🏾

PR #1694, opened 2025-10-06
[bridge]: update strategy mode names

Jaguar0625:
problem: strategy names are misleading
solution: rename to { 'stake', 'wrap', 'swap' } and rename section to 'global'


Wayonb (approved):
👍🏾

PR #1747, opened 2025-10-28
[explorer/puller] feat: add puller coverage report

AnthonyLaw:
Problems: Explorer puller coverage report missing.
Solution: Added coverage report setup.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -1,3 +1,4 @@
+import asyncio
 import configparser
 import threading
 from collections import namedtuple
@@ -6,6 +7,7 @@ from queue import Queue
 from symbolchain.facade.NemFacade import NemFacade
 from symbolchain.nem.Network import Network
 from symbollightapi.connector.NemConnector import NemConnector
+from symbollightapi.model.Exceptions import NodeException
 from zenlog import log
 
 from puller.db.NemDatabase import NemDatabase
@@ -101,6 +103,7 @@ class NemPuller:
 			# Check for sentinel value (None) to stop processing
 			if block is None:
 				log.info('Received stop signal, ending DB writer thread')
+				block_queue.task_done()
 				break
 
 			# Process and insert block
@@ -119,6 +122,21 @@ class NemPuller:
 
 		log.info(f'Database thread: {processed} blocks inserted')
 
+	async def _retry_get_blocks_after(self, height, retries=3, delay=2):
+		"""Retries fetching blocks after a given height with exponential backoff."""
+
+		for attempt in range(1, retries + 1):
+			try:
+				return await self.nem_connector.get_blocks_after(height)
+			except NodeException as error:
+				if attempt < retries:
+					wait_time = delay * (2 ** (attempt - 1))  # Exponential backoff: 2s, 4s, 8s
+					log.warning(f'Error fetching blocks after height {height} (attempt {attempt}/{retries}): {error}. Retrying in {wait_time}s...')
+					await asyncio.sleep(wait_time)
+				else:
+					log.error(f'Failed to fetch blocks after height {height} after {retries} attempts: {error}')
+					raise
+
 	async def sync_blocks(self, db_height, chain_height, queue_size=200, batch_size=50):
 		"""sync blocks from NEM network."""
 
@@ -136,7 +154,7 @@ class NemPuller:
 
 		try:
 			while chain_height > db_height:
-				blocks = await self.nem_connector.get_blocks_after(db_height)
+				blocks = await self._retry_get_blocks_after(db_height)
 
 				for block in blocks:
 					block_queue.put(block)
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -1,5 +1,7 @@
 import configparser
+import threading
 from collections import namedtuple
+from queue import Empty, Queue
 
 from symbolchain.facade.NemFacade import NemFacade
 from symbolchain.nem.Network import Network
@@ -44,6 +46,17 @@ class NemPuller:
 
 		return self.nem_facade.network.datetime_converter.to_datetime(timestamp).strftime('%Y-%m-%d %H:%M:%S+00:00')
 
+	def _commit_blocks(self, message=None):
+		"""Commit blocks to database with error handling."""
+
+		try:
+			self.nem_db.connection.commit()
+			if message:
+				log.info(message)
+		except Exception as error:
+			log.error(f'Commit error: {error}')
+			raise
+
 	def _process_block(self, cursor, block_data):
 		"""Process block data."""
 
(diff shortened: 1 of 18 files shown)


explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,106 @@
...
+
+		# commit changes
+		self.nem_db.connection.commit()
+
+		log.info('added block from height 1')
+
+	async def sync_blocks(self, db_height, chain_height):
+		"""Sync network blocks."""
+
+		# sync network blocks in database
```
Wayonb: For sync blocks, the code below will work.

To improve performance, have you tried overlapping I/O? i.e.
- one thread to read blocks from the NEM node
- one thread writing to the database

explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -0,0 +1,195 @@
...
+		self.assertEqual(str(puller.nem_facade.network), expected_network_str)
+
+	def test_create_testnet_puller_instance(self):
+		# Act:
+		puller = NemPuller('http://localhost:7890', self.config_ini, 'testnet')
+
+		# Assert
+		self._assert_puller_instance(puller, 'testnet')
+
+	def test_create_mainnet_puller_instance(self):
```
Wayonb: Create an instance with default ``network_type``

Wayonb (changes requested):

explorer/puller/scripts/ci/test.sh
```diff
@@ -3,4 +3,4 @@
 set -ex
 
 TEST_RUNNER=$([ "$1" = "code-coverage" ] && echo "coverage run --append" || echo "python3")
-${TEST_RUNNER} -m pytest --asyncio-mode=auto -v
+PYTHONPATH=.:../../lightapi/python ${TEST_RUNNER} -m pytest --asyncio-mode=auto -v
```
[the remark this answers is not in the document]
AnthonyLaw: same reason.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+class NemPuller:
+	"""Facade for pulling data from NEM network."""
+
+	def __init__(self, node_url, config_file, network_type='mainnet'):
+		"""Creates a facade object."""
+
+		config = configparser.ConfigParser()
+		config.read(config_file)
```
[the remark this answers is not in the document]
AnthonyLaw: it's db config. https://github.com/symbol/product/blob/fecff8c4ddd79e711c4a65cc058442e772a8416a/explorer/puller/tests/facade/test_NemPuller.py#L69

sample:
```
[nem_db]
database = nem_db
user = usename
password = password
host = localhost
port = 5432
```

I think I'll create `tests/resource/db_config.ini` and read the file from there, instead create file during test

explorer/puller/scripts/ci/test.sh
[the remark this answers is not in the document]
Wayonb: The ``PYTHONPATH`` can be removed since the lightapi is published

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,167 @@
...
+		cursor = self.nem_db.connection.cursor()
+		processed = 0
+
+		log.info('DB writer thread started')
+
+		try:
+			while not stop_event.is_set():
+				try:
+					# retrieve block with timeout
+					block = block_queue.get(timeout=1.0)
```
Wayonb: Good Job making this sync multi-threaded 👍🏾 
 
Instead of having a timeout to wake up to check if the sync is complete, you can push ``None`` into the queue to mark it complete. When get() returns ``None`` exit the writer.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,167 @@
...
+					block_queue.put(block)
+
+				last_block_height = blocks[-1].height
+				first_block_height = blocks[0].height
+
+				db_height = last_block_height
+
+				log.info(f'fetch block from height {first_block_height} - {last_block_height}')
+
+			# Wait for all blocks to be processed
```
Wayonb, line 169: Here add ``block_queue.put(None)`` to mark complete after the join.  
the ``stop_event`` should not be needed.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,167 @@
...
+
+				block_queue.task_done()
+
+			# Final commit for remaining blocks
+			if processed % batch_size != 0:
+				self._commit_blocks()
+
+			log.info(f'Database thread: {processed} blocks inserted')
+
+		except Exception as error:
```
Wayonb: Is there any error that can be retried? Or is all error fatal? 🤔

Wayonb (changes requested):

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: I considered handling potential runtime database errors, such as duplicate block inserts. However, based on the current logic, it looks like those cases shouldn’t occur.

Do you think this error-handling code is redundant?

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Thanks, that is a good idea.
I'll improve it.

explorer/puller/puller/facade/NemPuller.py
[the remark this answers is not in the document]
AnthonyLaw: I kept node_url outside the config because if a node goes down, we can quickly switch to another URL and keep syncing. The DB configuration is stable and rarely needs changes, so I didn’t include it in the same way.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,163 @@
...
+			# Wait for all blocks to be processed
+			block_queue.join()
+
+			# Send stop signal to db_writer thread
+			block_queue.put(None)
+			db_thread.join(timeout=10)
+
+			log.info('Block sync complete')
+
+		except Exception as error:
```
Wayonb, line 178: Same for this exception.  Should there be tries for certain HTTP exceptions? 🤔

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -0,0 +1,163 @@
...
+
+		log.info('DB writer thread started')
+
+		while True:
+			# retrieve block from queue
+			block = block_queue.get()
+
+			# Check for sentinel value (None) to stop processing
+			if block is None:
+				log.info('Received stop signal, ending DB writer thread')
```
Wayonb, line 105: Just to be safe, add ``block_queue.task_done()`` here also.

Wayonb (changes requested):

explorer/puller/puller/facade/NemPuller.py
Wayonb: If the exception is not recoverable, the only reason to catch it is to log a message or convert it to a more user-friendly/custom exception.

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: I assumed the puller is triggered every minute via cron, so I initially thought retries might not be necessary since the next cron run would automatically retry the sync.

However, adding retries inside the puller is probably worth it, because it can recover from temporary connection timeouts within a few seconds instead of waiting for the next scheduled run.

Wayonb (approved):
👍🏾

PR #1805, opened 2025-11-21
[wallet/common] feat: bridge, transaction bundle, minor fixes

OlegMakarenko:
## Problem
- Wallet lacks bridge functionality.

Minor bugs:
- Incorrect mosaic formatting.
- Error when switching the network with no accounts in Wallet Controller.
- Missing listener cleanup after wallet clear.
- Websocket url helper for Ethereum uses URL class which is not fully supported by React-Native.

## Solution
**Core:**
1. Created Bridge Manager that  handles API communication with the bridge:
 -  Fetches config.
 -  Fetches estimation.
 -  Fetches request and error history.
  The Bridge Manager accepts 2 Wallet Controllers that handle wallets in 2 chains. Once `BridgeManager.load()` is called the config is fetched and registered in each Wallet Controller (native and wrapped). That allows Wallet Controllers to output the bridge tokens.

2. Created Bridge Module (pluggable to Wallet Controller). The module is configurable with chain-related Bridge Helper. The module:
 - Stores bridge configs.
 - Return the list of tokens that participated in the bridge swaps.
 - Creates bridge swap transaction (forward it to Bridge Helper that is implemented for a specific chain).
 
**Symbol:**
3. Created Bridge Helper that creates the bridge transfer transaction on Symbol chain and fetches token info from Symbol node.

**Ethereum:**
4. Created Bridge Helper that creates the bridge ERC-20 transaction on Ethereum chain and fetches token info from Ethereum node.

Example:

```js
const bridgeManager = new BridgeManager({
        id: 'symbol-ethereum-xym-wxym',
        nativeWalletController: symbolWalletController,
        wrappedWalletController: ethereumWalletController,
        bridgeUrls: {
            testnet: 'https://bridge.symbol.tools/testnet/ethereum-wrapped',
            mainnet: 'https://bridge.symbol.tools/ethereum-wrapped',
        }
});

await bridgeManager.load(); // fetches config and registers it in symbolWalletController and ethereumWalletController
const history = bridgeManager.fetchRecentHistory(10); // fetches recent requests and errors
const tokens = symbolWalletController.modules.bridge.tokens; // the list of tokens that used in all bridges in which controller is registered.

// creates bridge transfer transaction.
const transaction = await symbolWalletController.modules.bridge.createTransaction({ 
  recipientAddress: myAddressOnEthereumNetwork,
  amount: '10',
  bridgeId: bridgeManager.id
}); 
```

5. Fixed bugs.
6. Added Transaction Bundle entity for the consistency across chains. Now this entity is expected for signing and announcing transactions. The entity in most of the cases is just a wrapper for the transaction object. Since the multisig transfer on Symbol requires hash lock + aggregate pair in this case the bundle contains 2 transactions. 
7. Make consistency in fee and fee tier objects across chains.


wallet/common/core/src/lib/bridge/BridgeApi.js
```diff
@@ -0,0 +1,70 @@
+export class BridgeApi {
+	/**
+     * @param {object} options
+     * @param {object} options.bridgeUrls
+     * @param {Function} options.makeRequest
+     * @param {string} options.networkIdentifier
+     */
+	constructor({ bridgeUrls, makeRequest, networkIdentifier }) {
```
Wayonb: Do you need the space before and after {}?

wallet/common/core/src/lib/bridge/BridgeManager.js
```diff
@@ -0,0 +1,558 @@
...
+	get config() {
+		return this.#config;
+	}
+
+	/**
+     * Get native network bridge token info.
+     * @returns {TokenInfo|null} - Token info of the native network, or null if not loaded.
+     */
+	get nativeTokenInfo() {
+		return this.#config ? this.#config.nativeNetwork.tokenInfo : null;
```
Wayonb: Is the config not loaded at startup?

wallet/common/core/src/lib/bridge/BridgeManager.js
```diff
@@ -0,0 +1,558 @@
...
+		this.#bridgeApi.setNetworkIdentifier(nativeNetworkIdentifier);
+
+		// Fetch config from bridge API
+		const config = await this.#bridgeApi.fetchConfig();
+
+		// Validate config
+		const isNativeChainMatch = config.nativeNetwork.blockchain === this.#nativeWalletController.chainName;
+		const isWrappedChainMatch = config.wrappedNetwork.blockchain === this.#wrappedWalletController.chainName;
+
+		if (!isNativeChainMatch || !isWrappedChainMatch)
```
Wayonb: I guess mainnet or testnet does not matter?

Wayonb (changes requested):

wallet/common/core/src/lib/bridge/BridgeManager.js
OlegMakarenko: Yes the config should be loaded first. But until the loading completed or when the config is re-loading during network change, this guards from getting error.

wallet/common/core/src/lib/modules/BridgeModule.js
```diff
@@ -0,0 +1,195 @@
...
+
+	/**
+	 * Check if the specific bridge config is added to the module.
+	 * @param {BridgeNetworkConfig} configToTest - Config to check.
+	 * @returns {boolean} - True if the config is already added, false otherwise.
+	 */
+	hasConfig = configToTest => {
+		const config = this.getConfig(configToTest.bridgeAddress);
+
+		return config !== null && JSON.stringify(config) === JSON.stringify(configToTest);
```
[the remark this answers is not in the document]
OlegMakarenko: I expected the same order, but if it changes this will fail. Thanks, fixed!

wallet/common/core/src/lib/bridge/BridgeManager.js
OlegMakarenko: Added check.

wallet/common/core/src/lib/modules/BridgeModule.js
```diff
@@ -0,0 +1,195 @@
...
+	 * @param {BridgeNetworkConfig} options.config - bridge source network configuration
+	 * @param {number} [options.fee] - transaction fee
+	 * @returns {Promise<object>} - transaction object
+	 */
+	createTransaction = async options => {
+		const { recipientAddress, token, config, fee } = options;
+		const { currentAccount, networkProperties, networkIdentifier } = this.#walletController;
+
+		// Verify bridge is in the same chain and network as the wallet controller
+		if (config.blockchain !== this.#walletController.chainName)
```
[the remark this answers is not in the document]
OlegMakarenko: It makes sense. Removed.

I changed the way how the config is stored in the module, now that allows to simplify this method.

Wayonb (approved):
👍🏾

PR #1830, opened 2025-12-04
[bridge] Add initial openapi specification for bridge API

cryptoBeliever:
Problem: There is no openapi specification for bridge
Solution: Added initial openapi specification


bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+    description: Local development server
+tags:
+  - name: Bridge
+    description: Provides access to bridge configuration.
+  - name: Wrap
+    description: Endpoints related to wrapping tokens from the native chain into the wrapped chain.
+  - name: Unwrap
+    description: |
+      Endpoints related to unwrapping tokens from the wrapped chain back into the native chain.
+      Unwrap operations are supported only in wrap and stake bridge modes (please check the bridge README for more details).
```
segfaultxavi: This comment about reading the README should really be at the top, in `info.description`, and include a link to the file.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/requests:
+    get:
+      summary: Returns wrap requests information.
+      description: |
+        Retrieves a paginated list of wrap requests detected on the native network.
+        These represent user-initiated deposits that trigger wrapping.
```
segfaultxavi:
```suggestion
        Retrieves a paginated list of the ongoing wrap requests detected on the native network.
        These represent user-initiated deposits that trigger wrapping.
        Once complete, the requests disappear from this list.
```
Am I correct? The list only includes requests currently being processed?

Things that might be useful to mention here too:
- How can I recover a request that has already been completed, either failed or succeeded?
- Which endpoint do I use to make such a request?
- Is there any other way to monitor requests, like websockets?

In general, it's always good to include lots of links to other parts of the documentation, so if this section is not exactly what the reader was looking for, at least they have a thread to pull and continue looking. Otherwise, they have no choice but to read the whole spec.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        - Wrap
+      operationId: getWrapRequests
+      parameters:
+        - $ref: '#/components/parameters/Offset'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Sort'
+        - $ref: '#/components/parameters/PayoutStatus'
+      responses:
+        200:
+          description: List of wrap requests recorded by the bridge.
```
segfaultxavi, line 68:
```suggestion
          description: List of ongoing wrap requests recorded by the bridge.
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        500:
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/requests:
+    get:
+      summary: Returns wrap requests information.
```
segfaultxavi, line 53:
```suggestion
      summary: Returns information about ongoing wrap requests.
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/requests/{address}:
+    get:
+      summary: Returns wrap requests filtered by sender or destination address.
+      description: |
+        Retrieves wrap requests associated with the provided address.
+        This includes both source (sender) and destination addresses.
```
segfaultxavi: Same comments as before: clarify if the requests are still ongoing or not, and include as many links as possible to related topics.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        200:
+          description: Wrap requests linked to the specified address.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/WrapRequest'
+        400:
+          description: The provided address is invalid for filtering wrap requests.
```
segfaultxavi: Is this the only reason why you can get a 400?
Because the previous endpoint says: Invalid request parameters (e.g., invalid pagination or filters).

Also, since sender and receiver addresses are on different networks, their formats can be different, no? This deserves a comment.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/requests/hash/{transaction_hash}:
+    get:
+      summary: Returns wrap requests by request or payout transaction hash.
+      description: Retrieves wrap requests matched by either the request hash or associated payout hash.
```
segfaultxavi: You need to explain what are these hashes and how to obtain them.
_I guess_ one is the hash of the transaction where the initial deposit was made that initiated the request?
And the other one is the hash of the transaction on the wrapped network where the wrapped token was transferred to the user? Wouldn't the request be finished then?
As you can see, I am confused, and I shouldn't need to guess anything :)

Also, if the transactions happen on different networks they probably have different formats, so that's worth mentioning.

Also, if we're still talking about requests being processed you need to say it. You can use the word "ongoing".

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        200:
+          description: Wrap requests corresponding to the specified transaction hash.
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/WrapRequest'
+        400:
+          description: Invalid transaction hash format.
```
segfaultxavi: Worth explaining what the correct format is, because if multiple networks are involved, this can be confusing.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        500:
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/errors:
+    get:
+      summary: Returns wrap errors.
```
segfaultxavi, line 175: Is this the list of ALL errors since the bridge was started? Or only recent ones? This needs to be clarified.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+      description: Retrieves a paginated list of errors encountered while processing wrap requests.
```
segfaultxavi: Worth mentioning something like:

> If a request errors out you will not find it in `/wrap/requests`. In that case use this endpoint to retrieve it.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/errors/{address}:
+    get:
+      summary: Returns wrap errors for a specific sender address.
+      description: Retrieves wrap errors linked to the specified request sender address.
```
segfaultxavi: Same comments as above.

The similar `/wrap/requests/{address}` endpoint can filter by both sender and receiver addresses, but this one only filters by sender. This is a bit surprising so it deserves a comment. Or better yet: make them behave the same way, so it's easier to understand and to explain :D

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        - Wrap
+      operationId: getWrapErrorsByAddress
+      parameters:
+        - $ref: '#/components/parameters/Address'
+        - $ref: '#/components/parameters/Offset'
+        - $ref: '#/components/parameters/Limit'
+        - $ref: '#/components/parameters/Sort'
+      responses:
+        200:
+          description: Wrap errors associated with the provided address.
```
segfaultxavi:
```suggestion
          description: List of wrapping errors associated with the provided address.
```

Because "wrap" is also a verb, so the sentence can be a bit weird. There are multiple instances of this issue.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          content:
+            application/json:
+              schema:
+                type: array
+                items:
+                  $ref: '#/components/schemas/WrapError'
+        400:
+          description: Invalid address format.
```
segfaultxavi: Same comments as for `/wrap/requests/{address}`.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/errors/hash/{transaction_hash}:
+    get:
+      summary: Returns wrap errors for a specific transaction hash.
+      description: Retrieves wrap errors generated during processing of a transaction with the given hash.
```
segfaultxavi: Same comments as for `/wrap/requests/hash/{transaction_hash}` regarding the hash.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        500:
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/prepare:
```
segfaultxavi, line 286: The name "prepare" is a bit confusing. According to the description this endpoint returns information about a request without actually executing it, but it really does not prepare anything, no? There is no change of state in the bridge, and the endpoint does not return any information that is required to perform the actual request, so it does not look like anything is being "prepared".
Maybe "estimate", "preview" or "evaluate" are clearer?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+    post:
+      summary: Prepares a wrap transaction.
```
segfaultxavi:
```suggestion
      summary: Estimates the cost of a wrap transaction.
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+      description: |
+        Estimates the outcome of a wrap operation, including conversion results and fees.
+        This endpoint simulates how many wrapped tokens the user will receive based on the bridge’s
+        current configuration, fee structure, and mode.
```
segfaultxavi: This description is good! And it does not use the word "prepare", see? :D

Only thing missing is pointers to more information, for example "If the cost is acceptable, submit the actual wrap request by transferring the desired amount to the bridge account as explained in the README <link>".

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+        500:
+          description: Unexpected internal bridge error during preparation.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /unwrap/requests:
```
segfaultxavi, line 322: The wrap and unwrap endpoints are pretty symmetrical, so all my previous comments apply to unwrap operations too.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+      name: address
+      in: path
+      required: true
+      description: |
+        Address used for filtering wrap or unwrap requests and errors.
+        The exact format of addresses depends on the configured network (e.g. Ethereum, Symbol, NEM).
+        Clients should validate addresses according to the blockchain in use.
+      schema:
+        type: string
+        example: 0x12ab34cd56ef7890ab12cd34ef567890ab12cd34
```
segfaultxavi: Is it possible to provide examples for each different chain?
[Using the `examples` tag instead of `example`](https://learn.openapis.org/specification/docs.html#adding-examples)?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+
+    TransactionHash:
+      name: transaction_hash
+      in: path
+      required: true
+      description: Transaction hash used to filter wrap or unwrap requests and errors.
+      schema:
+        type: string
+        pattern: '^[0-9A-Fa-f]{64}$'
+        example: 3f5b8f9a2bb2f6a9c3d2b4f0a1e2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a90
```
segfaultxavi: Hashes have the same format on all networks?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        type: integer
+        minimum: 0
+        default: 25
+
+    Sort:
+      name: sort
+      in: query
+      required: false
+      description: |
+        Sort order for results by request transaction height. Values: `1` (ascending) or `0` (descending).
```
segfaultxavi:
```suggestion
        Results are always sorted by the block height of the transaction that initiated the request.
        This field selects the direction of the sort: `1` (ascending) or `0` (descending).
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        enum:
+          - 1
+          - 0
+        default: 1
+
+    PayoutStatus:
+      name: payout_status
+      in: query
+      required: false
+      description: Filter by payout processing status (0 – unprocessed, 1 - sent, 2 - completed, 3 - failed).
```
segfaultxavi:
```suggestion
      description: "Filter by payout processing status: `0` (unprocessed), `1` (sent), `2` (completed) or `3` (failed)."
```

To be consistent with the previous field.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          example: ethereum
+        network:
+          description: Network environment (testnet or mainnet).
+          type: string
+          enum:
+            - testnet
+            - mainnet
+          example: testnet
+        bridgeAddress:
+          description: Address monitored for incoming wrap or unwrap requests.
```
segfaultxavi:
```suggestion
          description: Address monitored for incoming wrap (when on the native network) or unwrap (when on the wrapped network) requests.
```

Otherwise it looks like the same address can take care of both wrap and unwrap requests.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          example: 0x12ab34cd56ef7890ab12cd34ef567890ab12cd34
+        tokenId:
+          description: Token identifier associated with this network configuration.
```
segfaultxavi: This needs more information. What's this, exactly?
The ERC-20 contract address for wrapped tokens on Ethereum?
A MosaicID on Symbol?
0x0000 for the native token?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          example: "0x0000000000000000000000000000000000000000"
+        defaultNodeUrl:
+          description: Default API or RPC node used for blockchain access.
+          type: string
+          format: uri
+          example: https://rpc.testnet.example
+        explorerUrl:
+          description: URL of a block explorer for viewing transactions and accounts.
```
segfaultxavi: The bridge doesn't actually use this, right?
It would be good to make it explicit to avoid confusion. Something like "The bridge itself does not use the Explorer. This information is provided for convenience."

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          example: 0x12ab34cd56ef7890ab12cd34ef567890ab12cd34
+        tokenId:
+          description: Token identifier associated with this network configuration.
+          type: string
+          example: "0x0000000000000000000000000000000000000000"
+        defaultNodeUrl:
+          description: Default API or RPC node used for blockchain access.
+          type: string
+          format: uri
+          example: https://rpc.testnet.example
```
segfaultxavi: Can we provide the URL of one of our working nodes, like https://reference.symboltest.net:3001?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        explorerUrl:
+          description: URL of a block explorer for viewing transactions and accounts.
+          type: string
+          format: uri
+          example: https://explorer.testnet.example
```
segfaultxavi: Same thing, can we use our explorer?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        requestTransactionHash:
+          description: Hash of the transaction submitting the request.
+          type: string
+          example: 3f5b8f9a2bb2f6a9c3d2b4f0a1e2c3d4e5f6a7b8c9d0e1f2a3b4c5d6e7f8a90
+        requestTransactionSubindex:
+          description: Subindex for Symbol aggregate transactions or -1 for others.
+          type: integer
+          example: -1
+        senderAddress:
+          description: Address of the user initiating the request.
```
segfaultxavi: Worth mentioning that the format will depend on the network.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          example: 0x12ab34cd56ef7890ab12cd34ef567890ab12cd34
+        requestAmount:
+          description: |
```
segfaultxavi, line 806: The amount is in atomic units? This has to be explained, and use the right word (we call it "atomic unit" in Symbol, but I don't know what Ethereum calls it).
Also, if these are atomic units, why are decimal strings needed?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+            Amount of native tokens deposited for wrapping or unwrapping.
+            Amounts are returned as strings to preserve precision (decimal or integer string).
+          type: string
+          example: "1000000000000000000"
+        destinationAddress:
+          description: Address on the network that will receive the tokens after swap.
```
segfaultxavi: Same comment about formats on different networks.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          example: 0xE7DefF030632d08DA0ff3c5e9650459Dc8f746b7
+        payoutStatus:
+          description: Current payout processing status of this request (0 – unprocessed, 1 - sent, 2 - completed, 3 - failed).
```
segfaultxavi:
```suggestion
          description: "Current payout processing status: `0` (unprocessed), `1` (sent), `2` (completed) or `3` (failed)."
```

Let's be consistent!

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: integer
+          enum:
+            - 0
+            - 1
+            - 2
+            - 3
+          example: 0
+        payoutTransactionHash:
+          description: Hash of the payout transaction, if already initiated.
```
segfaultxavi: And what if it has not been already initiated?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          nullable: true
+          example: 03D848C2455CADBC26AB5F940D2232D168EFA6EA770D6F583E030C160769CB2E
+        requestTimestamp:
+          description: Timestamp (Unix) when the request transaction occurred.
```
segfaultxavi:
```suggestion
          description: Timestamp (in milliseconds since the Unix epoch) when the request transaction occurred.
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: integer
+          example: 1759781792
+        payoutTransactionHeight:
+          description: Block height of the payout transaction, if available (string-encoded integer).
```
segfaultxavi: Again, what do you get here if not available?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          nullable: true
+          example: "12345700"
+        payoutNetAmount:
+          description: Net payout amount sent after deducting fees (string-encoded integer).
```
segfaultxavi: Must explain the units (whole, fractional, atomic...) and note that this is on the destination network, using the configured token.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: string
+          nullable: true
+          example: "995000000000000000"
+        payoutTotalFee:
+          description: Total fees deducted from the wrap payout (string-encoded integer).
```
segfaultxavi: Units.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+            Conversion rate expressed in parts-per-million (PPM), defining
+            how many wrapped tokens correspond to one native token.
+            - 1,000,000 PPM = 1.0 (1:1 ratio)
+            - 2,000,000 PPM = 2.0 (1 native converts to 2 wrapped)
+            - 500,000 PPM   = 0.5 (1 native converts to 0.5 wrapped)
+          type: string
+          nullable: true
+          example: "1000000"
+        payoutSentTimestamp:
+          description: Timestamp when the payout transaction was broadcast.
```
segfaultxavi: In milliseconds since Unix epoch, I presume?
All timestamp fields should clarify this.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: integer
+          nullable: true
+          example: 1759781792
+        payoutTimestamp:
+          description: Timestamp when the payout transaction was finalized.
```
segfaultxavi: Same.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+          type: integer
+          nullable: true
+          example: 1759981792
+        errorMessage:
+          description: Error message associated with this wrap request, if any.
```
segfaultxavi: Is this a free-form text, or an enum we can give more information?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+    UnwrapRequest:
+      allOf:
+        - $ref: '#/components/schemas/WrapRequest'
+      description: |
+        Represents an unwrap request detected on the wrapped network.
+        Unwrap operations are available only in wrap and stake modes.
+
+    WrapError:
+      type: object
+      description: Error entry generated during request processing.
```
segfaultxavi:
```suggestion
      description: Error entry generated while processing a request.
```

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+        requestTransactionSubindex:
+          description: Aggregate transaction subindex for Symbol, or -1 otherwise.
+          type: integer
+          example: -1
+        senderAddress:
+          description: Address that submitted the invalid request.
+          type: string
+          example: 0x12ab34cd56ef7890ab12cd34ef567890ab12cd34
+        errorMessage:
+          description: Reason explaining why the request was considered invalid.
```
segfaultxavi: Free-form string or enum?

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+      required:
+        - recipientAddress
+        - amount
+      properties:
+        recipientAddress:
+          description: Address that would receive the payout of the simulated operation.
+          type: string
+          example: 0xabcdefabcdefabcdefabcdefabcdefabcdefabcd
+        amount:
+          description: |
```
segfaultxavi, line 992: Clarify the units. Whole or atomic? Are decimals accepted?
All fields involving amounts should clarify this, and the currency (or token) of the amount.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,919 @@
...
+      properties:
+        error:
+          type: string
+          description: Human-readable explanation of the error.
+          example: "Invalid address format"
+        errorCode:
+          type: string
+          description: Categorized error code when applicable.
+          nullable: true
+          enum:
```
segfaultxavi, line 1107: Are these really the only possible values?
I am missing things like "Invalid request parameters", for example.

segfaultxavi (commented):
SORRY! 😅

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Changed.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Yes, that is the complete list of error codes. For invalid request parameters, errorCode is set to null. I’ve added this to the list of planned API improvements, which I will address later.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Added separate examples for Ethereum, Symbol and NEM.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Yes. Hash format is same across all supported networks. Added information + remark for Ethereum to use hash without 0x prefix.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Added more information.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Right, changed descrption. It's always integer, changed that as well.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: I will add it to potential changes.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: API returns all requests / errors. Not sure if it's worth to mention. I provided some changes to specification, hopefully it better now.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: You are right, invalid filters/pagination can cause also 400. Changed that comment.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Errors doesn't have informaion regarding request destination address. I added comment regarding that.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: That's correct. Also for Ethereum we don't use 0x prefix for hashes. Added comment regarding that.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Yes, it's atomic and in source network. Added to description.

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Yes. Those are two possible values (for invalid params code is null at this moment). I plan to provide some small fixes to API and ill introduce enum for INVALID_REQUEST_PARAMS. will change api specification then too.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,1111 @@
...
+tags:
+  - name: Bridge
+    description: Provides access to bridge configuration.
+  - name: Wrap
+    description: Endpoints related to wrapping tokens from the native chain into the wrapped chain.
+  - name: Unwrap
+    description: |
+      Endpoints related to unwrapping tokens from the wrapped chain back into the native chain.
+      Unwrap operations are supported only in wrap mode and stake mode. This is the authoritative statement;
+      individual endpoints do not repeat this availability text.
```
segfaultxavi: This comment about the authoritative statement is a bit weird.
Also not very useful unless you read it first and remember it every time you check an endpoint.
I would remove it.

bridge/bridge/api/static/openapi/openapi.yaml
segfaultxavi: Got it!

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,1111 @@
...
+          description: Unexpected internal error.
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/requests:
+    get:
+      summary: Returns wrap requests information.
+      description: |
```
segfaultxavi, line 54: I still think it is worth saying that these endpoints returns ALL requests since the beginning of the bridge operation.
It confused me, so I guess it can confuse other people.

bridge/bridge/api/static/openapi/openapi.yaml
segfaultxavi: As before, I still think it's worth saying that this returns an ever-growing list of all errors ever detected.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,1111 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /wrap/errors/{address}:
+    get:
+      summary: Returns wrap errors for a specific sender address.
+      description: |
+        Retrieves wrap errors associated with the provided address.
+        The address filter matches the **request address on the native network** (sender).
+        Addresses from native network is valid and will be used to filter wrap request errors accordingly.
```
segfaultxavi:
```suggestion
        Addresses from the native network are valid and will be used to filter wrap request errors accordingly.
```

bridge/bridge/api/static/openapi/openapi.yaml
segfaultxavi: If this endpoint cannot be renamed, you can ignore this comment and resolve it.

bridge/bridge/api/static/openapi/openapi.yaml
```diff
@@ -0,0 +1,1111 @@
...
+              schema:
+                $ref: '#/components/schemas/ErrorResponse'
+
+  /unwrap/errors/{address}:
+    get:
+      summary: Returns unwrap errors for a specific address.
+      description: |
+        Retrieves unwrap errors associated with the provided address.
+        The address filter matches the **request address on the wrapped network** (sender). 
+        Addresses from wrapped network is valid and will be used to filter unwrap request errors accordingly.
```
segfaultxavi:
```suggestion
        Addresses from the wrapped network are valid and will be used to filter unwrap request errors accordingly.
```

segfaultxavi (changes requested):
Thanks for considering all the comments :)

I still think we need to talk about ALL transactions being returned.
Besides that, only a couple of typos!

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Ok, added

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Ok. I left only `Unwrap operations are supported only in wrap mode and stake mode.`. It’s placed in the Unwrap section, so it’s more likely users will notice it.

image

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: Added that informaion

bridge/bridge/api/static/openapi/openapi.yaml
cryptoBeliever: ok, added :+1:

segfaultxavi (approved):
🎉

PR #1854, opened 2025-12-18
[lightapi/python] feat: added query account mosics 

AnthonyLaw:
Problem: `account_info` was missing multisig information, local harvesting status, and did not query owned account mosaics.
Solution: 
1. Extended account_info to include status, min_cosignatories, cosignatories, and cosignatory_of.
2. Added a query to retrieve owned account mosaics.


Wayonb (approved):
👍🏾

PR #1864, opened 2025-12-29
[bridge]: resolve OpenAPI spec URL dynamically based on current path

cryptoBeliever:
## Problem

Swagger UI was failing to load the OpenAPI spec when served under prefixed paths (e.g. /testnet/ethereum-wrapped/openapi).
Relative paths like ../static/... were resolved incorrectly, resulting in 404 Not Found.

## Solution

The OpenAPI spec URL is now derived dynamically from window.location.pathname by removing the /openapi suffix and appending the expected static spec location.

This allows the same index.html to work correctly under:
- /openapi
- /testnet/ethereum-wrapped/openapi

without any environment-specific configuration.


Wayonb (approved):
👍🏾

PR #1873, opened 2026-01-04
[explorer/puller] feat: process accounts from the blocks

AnthonyLaw:
Problems: Explorer puller missing process accounts from blocks.
Solution: 
- Added database account tables and upsert account.
- Implemented account info and mosaic query.
- Extract the address from transactions and block signer.
- Query account information and owned mosaics.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -28,13 +28,6 @@ class NemDatabase(DatabaseConnection):
 
 		cursor.execute(
 			'''
-			CREATE INDEX IF NOT EXISTS idx_accounts_importance
-				ON accounts(importance DESC);
-			'''
-		)
-
-		cursor.execute(
-			'''
 			CREATE INDEX IF NOT EXISTS idx_accounts_public_key
 				ON accounts (public_key)
 			'''
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -1,3 +1,4 @@
+import json
 from binascii import unhexlify
 
 from .DatabaseConnection import DatabaseConnection
@@ -10,6 +11,43 @@ class NemDatabase(DatabaseConnection):
 	def _create_table_indexes(cursor):
 		"""Creates table indexes."""
 
+		# Create indexes for accounts table
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_address
+				ON accounts (address)
+			'''
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_balance
+				ON accounts(balance DESC);
+			'''
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_importance
+				ON accounts(importance DESC);
+			'''
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_public_key
+				ON accounts (public_key)
+			'''
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_mosaics_gin
+				ON accounts USING GIN (mosaics)
+			'''
+		)
+
+		# Create indexes for blocks table
 		cursor.execute(
 			'''
 			CREATE INDEX IF NOT EXISTS idx_blocks_height_desc ON blocks(height DESC);
@@ -17,9 +55,36 @@ class NemDatabase(DatabaseConnection):
 		)
 
 	def create_tables(self):
-		"""Creates blocks database tables."""
+		"""Creates database tables."""
 
 		cursor = self.connection.cursor()
+
+		# Create accounts table
+		cursor.execute(
+			'''
+			CREATE TABLE IF NOT EXISTS accounts (
+				id serial PRIMARY KEY,
+				address bytea NOT NULL UNIQUE,
+				public_key bytea,
+				remote_address bytea,
+				importance decimal(20, 10) DEFAULT 0,
+				balance bigint DEFAULT 0,
+				vested_balance bigint DEFAULT 0,
+				mosaics jsonb DEFAULT '[]'::jsonb,
+				harvested_fees bigint DEFAULT 0,
+				harvested_blocks bigint DEFAULT 0,
+				status varchar(8) DEFAULT NULL,
+				remote_status varchar(12) DEFAULT NULL,
+				last_harvested_height bigint DEFAULT 0,
+				min_cosignatories int DEFAULT NULL,
+				cosignatory_of bytea[] DEFAULT NULL,
+				cosignatories bytea[] DEFAULT NULL,
+				updated_at timestamp DEFAULT CURRENT_TIMESTAMP
+			)
+			'''
+		)
+
+		# Create blocks table
 		cursor.execute(
 			'''
 			CREATE TABLE IF NOT EXISTS blocks (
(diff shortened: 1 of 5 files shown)


explorer/puller/puller/db/NemDatabase.py
```diff
@@ -10,16 +11,80 @@ class NemDatabase(DatabaseConnection):
...
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_accounts_balance
+				ON accounts(balance DESC);
+			'''
+		)
+
+		cursor.execute(
```
Wayonb: What is the use case for creating an index for importance?

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -73,3 +138,55 @@ def get_current_height(self):
...
+				harvested_blocks = EXCLUDED.harvested_blocks,
+				status = EXCLUDED.status,
+				remote_status = EXCLUDED.remote_status,
+				min_cosignatories = EXCLUDED.min_cosignatories,
+				cosignatory_of = EXCLUDED.cosignatory_of,
+				cosignatories = EXCLUDED.cosignatories,
+				updated_at = CURRENT_TIMESTAMP
+			''', (
+				account_info.address.bytes,
+				account_info.public_key.bytes if account_info.public_key else None,
```
Wayonb, line 174: Just wondering why the ``.bytes`` is needed? 🤔

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -43,11 +59,118 @@ def __init__(self, node_url, config_file, network_type='mainnet'):
...
+
+			# Handle levy recipient
+			levy = getattr(transaction, 'levy', None)
+			if levy:
+				addresses.add(str(Address(levy.recipient)))
+
+			# Handle multisig signatures
+			if hasattr(transaction, 'signatures'):
+				for signature in transaction.signatures:
+					other_account = getattr(signature, 'other_account', None)
```
Wayonb: Should this be ``otherAccount`` instead of ``other_account``?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -43,11 +59,118 @@ def __init__(self, node_url, config_file, network_type='mainnet'):
...
+					sender = getattr(signature, 'sender', None)
```
Wayonb: should it be ``signer`` instead of ``sender``?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -103,15 +288,27 @@ def _db_writer(self, block_queue, batch_size=50):
...
 			self._process_block(cursor, block)
 			processed += 1
 
+			# Extract addresses for account processing
+			addresses = self._extract_addresses_from_block(block)
+			pending_addresses.update(addresses)
+
 			# Batch commits
 			if processed % batch_size == 0:
+				asyncio.run(self._process_account_batch(cursor, pending_addresses))
```
Wayonb, line 302: How is performance when processing account information during the sync vs. at the end of syncing all blocks? 🤔

explorer/puller/scripts/ci/lint.sh
```diff
@@ -14,6 +14,6 @@ find . -type f -name "*.py" -print0 | PYTHONPATH=. xargs -0 python3 -m isort \
 	--check-only
 find . -type f -name "*.py" -print0 | PYTHONPATH=. xargs -0 python3 -m pycodestyle \
 	--config="$(git rev-parse --show-toplevel)/linters/python/.pycodestyle"
-find . -type f -name "*.py" -print0 | PYTHONPATH=. xargs -0 python3 -m pylint \
+find . -type f -name "*.py" -print0 | PYTHONPATH=.:../../lightapi/python xargs -0 python3 -m pylint \
```
Wayonb: We can also just build lightapi and push an internal package instead of adding the lightapi to PYTHONPATH

Wayonb (changes requested):

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I added this because we can list accounts with high importance.

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: Because in DB, the address and public key field type is bytea.
So I just converted to bytes.

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: nope, because this field already processed in `lightapi/python`

https://github.com/symbol/product/blob/9357e10cfcbf818e5ccfb0ca13faf2d561040473/lightapi/python/symbollightapi/model/Transaction.py#L476

https://github.com/symbol/product/blob/9357e10cfcbf818e5ccfb0ca13faf2d561040473/lightapi/python/symbollightapi/model/Transaction.py#L235

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Same reason, it's already processed, all signers will name it as the sender in the Transaction object.

https://github.com/symbol/product/blob/9357e10cfcbf818e5ccfb0ca13faf2d561040473/lightapi/python/symbollightapi/model/Transaction.py#L223

explorer/puller/scripts/ci/lint.sh
AnthonyLaw: That will be great. 🙏🏼

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: IMO, it’s better to process this at the end of block syncing. During block sync, duplicated addresses can appear, causing repeated requests for the same address. If we do it at the end, each address is requested only once.

explorer/puller/puller/db/NemDatabase.py
Wayonb: ok.  Usually, importance should be the same as balance unless the account is not fully vested.

explorer/puller/puller/db/NemDatabase.py
Wayonb: Ok.  I'm wondering if the public_key can be None? 🤔 
If you are using the PublicKey/PriviateKey from the SDK then ``.bytes`` should be there unless another problem.

explorer/puller/puller/facade/NemPuller.py
Wayonb: OK, Why use the getattr since the property should be present in CosignSignatureTransaction object?

explorer/puller/puller/facade/NemPuller.py
Wayonb: same as above, is the getattr needed?

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Yeah, it makes sense. I will update it.

explorer/puller/scripts/ci/lint.sh
Wayonb: Try removing this and Jenkins should be fine.

If you want to install locally - 
pip install -i https://registry.symbolsyndicate.us/repository/pypi-group/simple symbol-lightapi

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: In this case, I will drop this first. 
It will be added when we really need it.

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: > Ok. I'm wondering if the public_key can be None? 🤔

I think if the account has never done any transaction, the public key does not appear on the chain, iirc.

Wayonb (approved):
👍🏾

PR #1901, opened 2026-01-19
[lightapi/python] feat: added total fee and beneficiary properties

AnthonyLaw:
Problem: I noticed that the block info included `beneficiary` and `totalFee` properties, but they are not mentioned in the NEM docs.
Solution: Added `beneficiary` and `totalFee`, which makes it possible to calculate harvested fees directly instead of calculating from the transactions.


Wayonb (approved):
👍🏾

PR #1907, opened 2026-01-27
[explorer/rest] feat: Added account endpoint

AnthonyLaw:
Problem: Missing account endpoint in the rest service.
Solution: Added a new account endpoint that can query account information from the database by address or public key.

### DIFF
--- a/explorer/rest/rest/__init__.py
+++ b/explorer/rest/rest/__init__.py
@@ -4,6 +4,7 @@ from pathlib import Path
 
 from flask import Flask, abort, jsonify, request
 from flask_cors import CORS
+from symbolchain.CryptoTypes import PublicKey
 from zenlog import log
 
 from rest.facade.NemRestFacade import NemRestFacade
@@ -78,6 +79,32 @@ def setup_nem_routes(app, nem_api_facade):
 			abort(400)
 
 		return jsonify(nem_api_facade.get_blocks(limit=limit, offset=offset, min_height=min_height, sort=sort))
+
+	@app.route('/api/nem/account')
+	def api_get_nem_account():
+		address = request.args.get('address', '').strip() or None
+		public_key = request.args.get('publicKey', '').strip() or None
+
+		# Validate that exactly one of address or public_key is provided
+		if bool(address) == bool(public_key):
+			abort(400)
+
+		try:
+			if address and not nem_api_facade.nem_db.network.is_valid_address_string(address):
+				abort(400)
+
+			if public_key:
+				PublicKey(public_key)
+
+		except ValueError:
+			abort(400)
+
+		result = nem_api_facade.get_account(address=address, public_key=public_key)
+
+		if not result:
+			abort(404)
+
+		return jsonify(result)
 
 
 def setup_error_handlers(app):
--- a/explorer/rest/rest/__init__.py
+++ b/explorer/rest/rest/__init__.py
@@ -89,17 +89,19 @@ def setup_nem_routes(app, nem_api_facade):
 		if bool(address) == bool(public_key):
 			abort(400)
 
-		try:
-			if address and not nem_api_facade.nem_db.network.is_valid_address_string(address):
-				abort(400)
-
-			if public_key:
-				PublicKey(public_key)
-
-		except ValueError:
+		if address and not nem_api_facade.nem_db.network.is_valid_address_string(address):
 			abort(400)
 
-		result = nem_api_facade.get_account(address=address, public_key=public_key)
+		if public_key:
+			try:
+				PublicKey(public_key)
+			except ValueError:
+				abort(400)
+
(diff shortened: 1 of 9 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -38,6 +38,70 @@ def _create_block_view(self, result):
...
+			status,
+			remote_status,
+			last_harvested_height,
+			min_cosignatories,
+			cosignatory_of,
+			cosignatories
+		) = result
+
+		return AccountView(
+			address=str(Address(address)),
```
Wayonb, line 61: What format is the incoming address? I guess maybe bytes from the database?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -67,3 +131,22 @@ def get_blocks(self, limit, offset, min_height, sort):
 			results = cursor.fetchall()
 
 			return [self._create_block_view(result) for result in results]
+
+	def get_account(self, address=None, public_key=None):
```
Wayonb: might be better to split this into two functions.  One gets an account by address, and another by public_key.  This will require a value to be passed into the function.

With the current function, calling - get_account(), without passing a value for either address or public_key.  This will cause a runtime error.

explorer/rest/rest/facade/NemRestFacade.py
```diff
@@ -22,3 +27,13 @@ def get_blocks(self, limit, offset, min_height, sort):
 		blocks = self.nem_db.get_blocks(limit, offset, min_height, sort)
 
 		return [block.to_dict() for block in blocks]
+
+	def get_account(self, address=None, public_key=None):
+		"""Gets account by address or public key."""
```
Wayonb: Same here. The function can be called with no argument or with both address and public_key.

explorer/rest/rest/__init__.py
```diff
@@ -79,6 +80,32 @@ def api_get_nem_blocks():
...
+			if address and not nem_api_facade.nem_db.network.is_valid_address_string(address):
+				abort(400)
+
+			if public_key:
+				PublicKey(public_key)
+
+		except ValueError:
+			abort(400)
+
+		result = nem_api_facade.get_account(address=address, public_key=public_key)
```
Wayonb: I see the public_key and address are validated here. 👍🏾 I am leaving the comments above since ``get_account`` is part of the API and can be called elsewhere.

explorer/rest/tests/db/test_NemDatabase.py
```diff
@@ -66,3 +70,29 @@ def test_can_query_blocks_sorted_by_height_asc(self):
...
+
+		# Assert:
+		self.assertEqual(EXPECTED_ACCOUNT_VIEW_1, account_view)
+
+	def test_can_query_account_by_public_key(self):
+		# Arrange:
+		nem_db = NemDatabase(self.db_config, self.network)
+
+		# Act:
+		account_view = nem_db.get_account(public_key=ACCOUNTS[0].public_key)
```
Wayonb: Should also have test for nem_db.get_account() since this is valid.

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Yes, bytes from DB.
Or do you think the formatting better happen under AccoutView's to_dict method?

AccountView(Address(address))

to_dict -> str(self.address )

Wayonb: No, here is fine.

At first look I thought this ``str(Address(address))`` was converting str_address -> Address->back to string but then I notice its coming from the DB😅

AnthonyLaw:
I've split into two functions.
`get_account_by_address` and `get_account_by_public_key`

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -67,3 +131,31 @@ def get_blocks(self, limit, offset, min_height, sort):
 			results = cursor.fetchall()
 
 			return [self._create_block_view(result) for result in results]
+
+	def get_account_by_address(self, address):
+		"""Gets account by address."""
+
+		where_clause = 'address = %s'
```
Wayonb, line 150: There is a common code that can be refactored between the two functions


```
def _get_account(self, where_clause, query_bytes):

		sql = self._generate_account_query(where_clause)

		with self.connection() as connection:
			cursor = connection.cursor()
			cursor.execute(sql, (query_bytes))
			result = cursor.fetchone()

			return self._create_account_view(result) if result else None
```

Wayonb (changes requested):

Wayonb (approved):
👍🏾

PR #1921, opened 2026-02-02
[explorer/puller] feat: Added beneficiary field in block sync

AnthonyLaw:
Problem: We added `total fee` and `beneficiary` in symbolLightAPI, update in puller is needed.
Solution: Added beneficiary field in DB and puller logic, updated unit test, also bumped symbolLightAPI to 0.0.9


Wayonb (approved):
👍🏾

PR #1924, opened 2026-02-03
[explorer/puller] feat: Added calculate harvesting logic

AnthonyLaw:
Problem: missing harvesting fees calculation.
Solution: Extracted block fee and beneficiary from the block and update harvested fees.


Wayonb (approved):
👍🏾

PR #1927, opened 2026-02-07
[explorer/rest] feat: Added beneficiary properties in block query

AnthonyLaw:
Problem: missing `beneficiary` field in query.
Solution: added `beneficiary` and renamed `total_fee`

### DIFF
--- a/explorer/rest/rest/db/NemDatabase.py
+++ b/explorer/rest/rest/db/NemDatabase.py
@@ -25,17 +25,30 @@ class NemDatabase(DatabaseConnectionPool):
 		self.network = network
 
 	def _create_block_view(self, result):
-		harvest_public_key = PublicKey(_format_bytes(result[7]))
+		(
+			height,
+			timestamp,
+			total_fee,
+			total_transactions,
+			difficulty,
+			block_hash,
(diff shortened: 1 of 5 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -117,29 +151,37 @@ def _get_account(self, where_clause, query_bytes):
...
+		order_condition = f' ORDER BY id {sort}'
+		limit_condition = ' LIMIT %s OFFSET %s'
+
+		sql = self._generate_block_query(
+			where_condition=where_condition,
+			order_condition=order_condition,
+			limit_condition=limit_condition
+		)
+
+		params.extend([min_height, limit, offset])
```
Wayonb: Can the param be initialized here instead of extend()?

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Yes, `extend` is not needed.

Wayonb (approved):
👍🏾

PR #1933, opened 2026-02-12
[wallet/symbol/mobile] feat: add account selector and history screens, add custom Passcode component

OlegMakarenko:
## Problem
1. Outdated dependencies caused Android build errors due to incompatibility with the current React Native version.
2. The History screen (account transactions) is missing.
3. Account selection screens are missing (cannot add seed accounts or switch between them).
4. Header and navigation menu (footer) are missing.
5. A black screen overlaps the UI when quickly switching between modals.
6. A broken `MainActivity.kt` file prevents Android builds.

## Solution
1. Removed outdated dependencies and replaced them with an in-house implementation:
 - Removed `rn-fetch-blob` (file writing to device storage). It is no longer needed after removing the Paper Wallet feature.
 - Removed `@haskkor/react-native-pincode` (Passcode UI + PIN pad + passcode storage/verification), along with its sub-dependencies `react-native-svg` and `react-native-touch-id`.
 - Implemented a custom Passcode component with a PIN pad and animations.
 - Extended `PasscodeManager` with create/verify/store APIs and a lockout feature (locks for a period of time after the attempt limit is reached).
 
2. Created History screen:
  - Added screen component.
  - Added `useHistoryData` hook to manage screen logic.
  - Added `useTransactionHistory` and `useReceiptHistory` hooks with pagination via the Transactions and Statements APIs, plus helper utilities.
  - Added `TransactionListItem` and `ReceiptListItem` components (with dedicated hooks) for rendering items on the History screen.
  - Added `HistoryWidget` component intended for the Home screen.
  - Created a `FilteredListScreen` template component that handles layout and basic interactions for screens with a top filter and a section list below.

3. Created account screens:
  - `AccountList` screen to switch between wallet accounts, reorder them via long-press drag-and-drop, and remove accounts from the list. Includes a button to navigate to `AddSeedAccount`.
  - `AddSeedAccount` screen to add seed accounts from the list so they appear in `AccountList`.
  - `AccountCard` component used in these lists to render account name, balance, and address.
  - Added `useAccountBalances` and `useSeedAccountBalances` hooks to provide balances to the corresponding lists.

4. Added Header and navigation menu (footer) components:
  - Header displays the current account; tapping it navigates to `AccountList`.
  - Header includes a gear icon to open Settings.
  - Navigation menu currently navigates between Home and History.
  - Refactored the Router module.

5. Tuned delay values in the modal-transition workaround.

6. Fixed the broken `MainActivity.kt` file that prevented Android builds.

**Additionally**

7. Created additional Fixture Builder classes used in tests.

8. Added more tests.

9. Fixed TouchableNative animation on Android.


wallet/symbol/mobile/__fixtures__/local/AggregateTransactionFixtureBuilde.js
```diff
@@ -0,0 +1,184 @@
+import { TransactionFixtureBuilder } from '__fixtures__/local/TransactionFixtureBuilder';
+import { TransactionType } from 'wallet-common-symbol/src/constants';
+
+const EMPTY_AGGREGATE_FIXTURE = {
+	type: TransactionType.AGGREGATE_COMPLETE,
+	amount: '0',
```
Wayonb, line 6: missing fee?

wallet/symbol/mobile/__fixtures__/local/AggregateTransactionFixtureBuilde.js
```diff
@@ -0,0 +1,184 @@
...
+	 * @param {object} data - full transaction data.
+	 * @returns {AggregateTransactionFixtureBuilder}
+	 */
+	static createWithData = data => {
+		return new AggregateTransactionFixtureBuilder({
+			innerTransactions: data.innerTransactions ?? [],
+			cosignatures: data.cosignatures ?? [],
+			receivedCosignatures: data.receivedCosignatures ?? [],
+			amount: data.amount ?? '0',
+			...data
```
Wayonb, line 92: Should the spread operator for ``data`` be first?

wallet/symbol/mobile/__fixtures__/local/TransferTransactionFixtureBuilder.js
```diff
@@ -0,0 +1,212 @@
...
+};
+
+export class TransferTransactionFixtureBuilder extends TransactionFixtureBuilder {
+	/**
+	 * Creates a transfer transaction fixture.
+	 * 
+	 * @param {object} data - combined base + transfer data.
+	 */
+	constructor(data) {
+		super(data);
```
Wayonb, line 22: Should there be properties in the transfer class that need to be assigned, such as message and recipient?

wallet/symbol/mobile/src/components/features/Passcode/hooks/usePasscodeInput.js
```diff
@@ -0,0 +1,44 @@
...
+ * @returns {PasscodeInput} - Passcode input state and handlers.
+ */
+export const usePasscodeInput = ({ length, onComplete }) => {
+	const [value, setValue] = useState('');
+
+	const inputKey = key => {
+		if (value.length >= length) 
+			return;
+
+		const newValue = value + key;
```
Wayonb, line 25: I assume this means it only appends to the end of the string... so there is no insert in the middle of the passcode?

wallet/symbol/mobile/src/components/features/Filter.jsx
```diff
@@ -0,0 +1,232 @@
...
+	// Filter state helpers
+	const isFilterActive = name => !!value[name];
+	const isFilterAvailable = name => 
+		(Object.keys(value).length === 0 || Object.prototype.hasOwnProperty.call(value, name)) && !isDisabled;
+
+	// Handlers
+	const handleClear = () => Object.keys(value).length !== 0 && onChange({});
+
+	const handleFilterPress = filter => {
+		if (!isFilterAvailable(filter.name))
```
Wayonb: In what case can the user press the filter, but it is not available? Should the button be disabled🤔

wallet/symbol/mobile/src/hooks/usePagination.js
```diff
@@ -0,0 +1,109 @@
...
+			setPageNumber(prev => prev + 1);
+			setData(prevData => dataUpdater(shouldReplaceData ? [] : prevData, pageData));
+
+			const isPageEmpty = pageData.length === 0;
+			const isPageIncomplete = pageSize !== null && pageData.length < pageSize;
+
+			if (isPageEmpty || isPageIncomplete)
+				setIsLastPage(true);
+		} catch {
+			// Error handled by asyncManager.onError
```
Wayonb: I guess fetchPage does not need to return whether is succeed or not? 🤔

wallet/symbol/mobile/src/lib/passcode/PasscodeManager.js
```diff
@@ -1,14 +1,217 @@
...
+
+	return { hash, salt };
+};
+
+const safeCompareStrings = (left, right) => {
+	if (typeof left !== 'string' || typeof right !== 'string')
+		return false;
+
+	const maxLength = Math.max(left.length, right.length);
+	let mismatch = left.length ^ right.length;
```
Wayonb, line 38: If the length does not match, you should be able to skip the check below?

wallet/symbol/mobile/src/lib/passcode/PasscodeManager.js
```diff
@@ -1,14 +1,217 @@
...
+	resetFailedAttempts = async () => {
+		await this.storage.removeItem(STORAGE_KEY_FAILED_ATTEMPTS);
+	};
+
+	/**
+	 * Set lockout timestamp.
+	 * @returns {Promise<void>}
+	 */
+	setLockout = async () => {
+		const lockoutUntil = Date.now() + PASSCODE_LOCKOUT_DURATION_MS;
```
Wayonb: Should the lock duration increase for constective failures?

wallet/symbol/mobile/src/screens/account/components/AccountCard.jsx
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * AnimatedBalance component. Displays balance with fade animation only when value changes.
+ *
+ * @param {object} props - Component props
+ * @param {string} props.value - Balance value to display.
+ * @param {string} props.ticker - Currency ticker symbol.
+ *
+ * @returns {React.ReactNode} AnimatedBalance component
+ */
+// Note: This component is temporarily not used due to performance issues with React Native v0.82,
```
Wayonb: Should this code be remove until there is a fix?

wallet/symbol/mobile/src/screens/account/hooks/useAccountBalances.js
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * @typedef {Object} AccountBalance
+ * @property {string} balance - Current account balance. Relative amount (as string).
+ * @property {string} balanceChange - Difference between current and cached balance. Relative amount (as string).
+ * @property {boolean} isLoading - Whether the balance is currently being fetched.
+ */
+
+/**
+ * Calculates balance change between current and cached values.
```
Wayonb, line 12: empty line after function description?  Just to be consistent with below.

wallet/symbol/mobile/src/screens/history/utils/receipt-icon.js
```diff
@@ -0,0 +1,25 @@
+/**
+ * Receipt type enum for categorizing different types of receipts.
+ * @enum {string}
+ */
+export const ReceiptType = {
```
Wayonb, line 5: What about the other receipt types, like LockHash?

Wayonb (changes requested):
Thanks for fixing the tests' memory usage.  Looks good, just a few comments 👍🏾

wallet/symbol/mobile/src/components/features/Passcode/hooks/usePasscodeInput.js
OlegMakarenko: Yes, it only appends new digit to the end of the string.

wallet/symbol/mobile/src/components/features/Filter.jsx
OlegMakarenko: Yes, this is redundant. A child FilterChip component will not fire event when isDisabled is true. Thanks, removed.

wallet/symbol/mobile/src/screens/history/utils/receipt-icon.js
OlegMakarenko: Currently we are not presenting anything than harvesting rewards. Not sure wether it make sense to add more types here, since this enum is only used in presentation components. I keep it as enum in case in future we would need to present more types, so it easier to add.

wallet/symbol/mobile/__fixtures__/local/TransferTransactionFixtureBuilder.js
OlegMakarenko: Yes, the are members of the data object and are assigned in constructor of the TransactionFixtureBuilder.

wallet/symbol/mobile/src/hooks/usePagination.js
OlegMakarenko: I just checked useAsyncManager's call() function and yes - it returns and throws. Added it to usePagination as well for the consistency.

wallet/symbol/mobile/src/lib/passcode/PasscodeManager.js
OlegMakarenko: This is done for constant time comparison. If we skip it that could introduce an early exit timing difference.

wallet/symbol/mobile/src/screens/account/components/AccountCard.jsx
OlegMakarenko: Thanks, moved here: https://github.com/symbol/product/issues/2003

wallet/symbol/mobile/src/lib/passcode/PasscodeManager.js
OlegMakarenko: Yes, it makes sense. Now on every new lockout the duration is `default duration * number of lockouts since last success`. Also display on UI how many minutes remaining until lockout expires.

Wayonb (approved):
looks good...make sure to resolve conflicts 👍🏾

PR #1944, opened 2026-02-17
[wallet/common/core] fix: Listener Connection Attempts When no Listen Accounts Provided

OlegMakarenko:
## Problem

- The listener throws errors during `NetworkManager` node connection cycles when no listening accounts are provided.

## Solution

- Allow the account to be unset. Prevent the listener from starting when no listening account address is provided.


Wayonb (approved):
👍🏾

PR #1961, opened 2026-02-26
[explorer/rest] feat: added nem health endpoint

AnthonyLaw:
Problem: Frontend requires a health check endpoint for the backend service.
Solution: Added /api/nem/health endpoint that returns backend service health status, including last synced block information and synchronization errors.

### DIFF
--- a/explorer/rest/rest/facade/NemRestFacade.py
+++ b/explorer/rest/rest/facade/NemRestFacade.py
@@ -11,12 +11,13 @@ from rest.model.common import Pagination
 class NemRestFacade:
 	"""Nem Rest Facade."""
 
-	def __init__(self, node_url, db_config, network_name):
+	def __init__(self, db_config, rest_config):
 		"""Creates a facade object."""
 
-		self.network = NetworkLocator.find_by_name(Network.NETWORKS, network_name)
+		self.network = NetworkLocator.find_by_name(Network.NETWORKS, rest_config.network_name)
 		self.nem_db = NemDatabase(db_config, self.network)
-		self.nem_connector = NemConnector(node_url, self.network)
+		self.nem_connector = NemConnector(rest_config.node_url, self.network)
+		self.max_lag_blocks = rest_config.max_lag_blocks
 
 	def get_block(self, height):
 		"""Gets block by height."""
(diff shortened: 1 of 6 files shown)


explorer/rest/rest/facade/NemRestFacade.py
```diff
@@ -48,3 +52,33 @@ def get_accounts(self, pagination, sorting, is_harvesting):
...
+
+		latest_block = self.nem_db.get_blocks(
+			pagination=Pagination(1, 0),
+			min_height=1,
+			sort='DESC'
+		)[0]
+
+		last_synced_at = latest_block.timestamp
+		last_block_height = latest_block.height
+		is_healthy = True
```
Wayonb: Is the ``is_healthy`` flag for the NEM node or the rest backend?  Can the NEM node be up but the backend not syncing?  Should there be a status if the backend is not synced with the NEM node?

Wayonb (changes requested):

explorer/rest/rest/facade/NemRestFacade.py
AnthonyLaw: > Is the is_healthy flag for the NEM node or the rest backend? 

The purpose of this endpoint is for the frontend explorer application check on the backend service status.

> Can the NEM node be up but the backend not syncing? 
>  Should there be a status if the backend is not synced with the NEM node?

I'll add more fields:
- nodeUp: false/true
- nodeHeight: query chain height from the node.
- backendSynced: compare the database height with the chain height. If the chain height is more than 2 blocks ahead of the database height, it is considered false.

What do you think?

explorer/rest/rest/facade/NemRestFacade.py
```diff
@@ -48,3 +52,54 @@ def get_accounts(self, pagination, sorting, is_harvesting):
...
+
+		latest_block = self.nem_db.get_blocks(
+			pagination=Pagination(1, 0),
+			min_height=1,
+			sort='DESC'
+		)[0]
+
+		last_db_synced_at = latest_block.timestamp
+		last_db_height = latest_block.height
+		max_lag_blocks = 2  # Allow up to 2 blocks of lag between node height and database height
```
Wayonb: Should the ``max_lag_blocks`` be in the config?

Wayonb (approved):
👍🏾

PR #1964, opened 2026-02-28
[lightapi/python] fix: update transaction field.

AnthonyLaw:
Problem: `signer` and `recipient` fields are still in plain text, to be consistent, it need  convert to objects.
Solution: Wrap signer and recipient into Publickey and Address objects.


Wayonb:
> `signer` and `recipient` fields are still in plain text

Sorry, I don't understand why these need to be objects?

AnthonyLaw:
> > `signer` and `recipient` fields are still in plain text
> 
> Sorry, I don't understand why these need to be objects?

https://github.com/symbol/product/blob/e0aaae7ca7b36fc6bf7eb079223b83dcc9558d3a/lightapi/python/symbollightapi/connector/NemConnector.py#L364-L365

To be consistent with the other implementations, in the block query both `signer` and `beneficiary` are wrapped as Address and PublicKey.

The advantage of wrapping them as Address and PublicKey is that it simplifies explorer backend queries, as they can be used directly without converting the types again.

Wayonb (approved):
👍🏾

PR #1987, opened 2026-03-04
[lightapi/python] fix: wait when transaction hash is not found for Symbol

Wayonb:
problem: try_wait_for_announced_transaction does not sleep when the transaction hash is not found
         On some nodes, it takes a while for the transaction status to go from not found after submission
         to either unconfirmed or failed.  This causes the retry limit to hit before sleeping.
solution: change the try_wait_for_announced_transaction to always sleep after each transaction status call.


AnthonyLaw (approved):
👍🏼 

@Wayonb 
FYI: This PR is mixed with other commits, so you might need to rebase

PR #2001, opened 2026-03-10
[bridge] fix: use the release version of lightapi

Wayonb:
problem: Bridge is using the symbol-lightapi from dev branch
solution: update the Bridge to use release version of symbol-lightapi


Jaguar0625 (approved):
👍

PR #2045, opened 2026-03-27
[explorer/rest] fix: Unit test refactor

AnthonyLaw:
Problem: I found that those are repeated patterns in the unit test, such as pagination, sort, and data non exist.
Solution: I extracted and group them into `test_invalid_pagination_params`, `test_invalid_sort_params` and `test_data_not_found`


Wayonb (approved):
👍🏾

PR #2059, opened 2026-03-31
[explorer/puller] feat: process mosaic from blocks

AnthonyLaw:
Problems: Missing the process mosaic from blocks.
Solution:

- Added database mosaic tables.
- Implemented upsert mosaic and update mosaic supply change.


Wayonb (approved):
👍🏾

PR #2071, opened 2026-04-05
[explorer/rest] feat: Added mosaic rich list endpoint

AnthonyLaw:
Problem: Missing mosaic rich list and filter by namespace name.
Solution: 
- Added  `/mosaic/rich/list` endpoint for query by namespace name

### DIFF
--- a/explorer/rest/rest/db/NemDatabase.py
+++ b/explorer/rest/rest/db/NemDatabase.py
@@ -5,7 +5,7 @@ from symbolchain.nem.Network import Address
 
 from rest.model.Account import AccountView
 from rest.model.Block import BlockView
-from rest.model.Mosaic import MosaicView
+from rest.model.Mosaic import MosaicRichListView, MosaicView
 from rest.model.Namespace import NamespaceView
 
 from .DatabaseConnection import DatabaseConnectionPool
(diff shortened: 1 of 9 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -419,3 +434,41 @@ def get_mosaics(self, pagination, sort):
...
+					SELECT
+						a.address,
+						ar.remarks,
+						(mosaic->>'quantity')::bigint as balance
+					FROM
+						accounts a
+						LEFT JOIN account_remarks ar
+							ON ar.address = a.address
+						CROSS JOIN LATERAL jsonb_array_elements(a.mosaics) AS mosaic
+						WHERE (mosaic->>'namespace') = %s AND (mosaic->>'quantity')::bigint > 0
```
Wayonb, line 452: What value of ``quantity`` is considered to be rich?  should it be greater 10k and not zero? 🤔

Wayonb: I see support for other mosaics other than``nem.xem``.  So not sure if this is a rich list or just a sorted list of owners of a mosaic

Wayonb (approved):
👍🏾

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: > What value of quantity is considered to be rich? should it be greater 10k and not zero? 🤔

It sorts the mosaic `quantity`  from highest to lowest.
So the highest quantity is considered "rich." 

It supports other mosaics as well, so we can know who is the top holder for the specified mosaic.

Wayonb (commented):
👍🏾

PR #2084, opened 2026-04-08
[explorer/puller] feat: process transactions from blocks

AnthonyLaw:
Problems: Missing the process transactions from blocks.
Solution:
- Added database tables.
    - `transactions`: used to store all transactions from the blocks.
    - `transactions_mosaic`:  used to store extracted mosaic transactions, making it easier to filter in queries.
- Implemented process transaction logic.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -107,13 +107,6 @@ class NemDatabase(DatabaseConnection):
 			'''
 		)
 
-		cursor.execute(
-			'''
-			CREATE INDEX IF NOT EXISTS idx_tx_inner_transaction_id
-				ON transactions(inner_transaction_id)
-			'''
-		)
-
 		# Create indexes for transactions_mosaic table
 		cursor.execute(
 			'''
@@ -232,9 +225,7 @@ class NemDatabase(DatabaseConnection):
 				signature bytea,
 				amount bigint,
 				is_inner boolean DEFAULT false,
-				inner_transaction_id int,
-				payload jsonb,
-				FOREIGN KEY (inner_transaction_id) REFERENCES transactions(id)
+				payload jsonb
 			)
 			'''
 		)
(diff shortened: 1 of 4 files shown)


explorer/puller/puller/db/NemDatabase.py
```diff
@@ -63,6 +63,79 @@ def _create_table_indexes(cursor):
...
+			'''
+			CREATE INDEX IF NOT EXISTS idx_tx_sender_address
+				ON transactions(sender_address)
+			'''
+		)
+
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS idx_tx_sender_public_key
+				ON transactions(sender_public_key)
```
Wayonb: Just wondering if you need both the sender address and the public key as an index? Since public key can be converted to address.

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -149,6 +222,44 @@ def create_tables(self):
...
+				is_inner boolean DEFAULT false,
+				inner_transaction_id int,
+				payload jsonb,
+				FOREIGN KEY (inner_transaction_id) REFERENCES transactions(id)
+			)
+			'''
+		)
+
+		# Create Transactions mosaic table
+		cursor.execute(
```
Wayonb, line 234: Why a seperate table for transfer v2 transactions? 🤔

Wayonb (changes requested):

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I was thinking when doing a transaction list query about `address`, whether it'll filter the `sender address` or `recipient address`.

We are unable to convert the address to a public key.

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I'm not sure this is the most effective approach.

Previously, I stored a JSON array in a table column (`jsonb`).
However, I found it difficult to filter using queries, and indexing on the  `JSONB` type is not very effective. 

Another reason is related to the `mosaic/transfers` endpoint, which needs to query all transfers for a specified mosaic.

explorer/puller/puller/db/NemDatabase.py
Wayonb: yes, so the index would be on the address since the public key can always be converted to an address.  It's fine if you need both, but each index requires more resources.

AnthonyLaw: Actually, you're right, sender address as index is good enough for the query.

Wayonb (approved):
👍🏾

Issue #2047, opened 2026-03-27
Add separate command to Shoestring to create linking-transactions.dat


segfaultxavi:
The `linking-transactions.dat` file is generated with the `setup` command, and it contains transactions that enable harvesting.

These transactions have a deadline of 2h, so they must be signed and announced right after setting up the node.
If one decides to enable harvesting at a later stage, the transactions have expired, and I am not aware of any other mean to generate them.

A command separate from `setup` that regenerates them would be helpful in this situation.

PR #2086, opened 2026-04-08
[tools/shoestring]: add --output-transaction-only flag to setup command

Jaguar0625:
problem: linking-transactions.dat may expire before user is able to send it
solution: allow rerunning setup with --output-transaction-only to regenerate transaction

   issue: #2047


Wayonb (approved):
👍🏾

segfaultxavi (approved):
Works as advertised. I was able to recreate the linking transactions for a node I set up last week.

PR #2104, opened 2026-04-15
[explorer/puller] fix: refactor and bug fix

AnthonyLaw:
Problem: 
- Found that `sender_address` didn't convert to an address.
- `inner_transaction_id` is not required

Solution:
- convert sender public key to address.
- Removed inner_transaction_id, as it already has an index for transaction hash, when looking up for the multisig inner transaction, It can just query by transaction hash


Wayonb (approved):
👍🏾

Issue #2129, opened 2026-04-24
[tools/shoestring] renew-certificates fails to find CA certificate file


segfaultxavi:
The following command fails to find `ca.key.pem` even though it exists:

```sh
$ python3 -m shoestring renew-certificates --config config.ini --directory . --ca-key-path ca.key.pem
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "/home/xartigas/repos/tss/symbol/mkdocs/.venv/lib/python3.12/site-packages/shoestring/__main__.py", line 54, in <module>
    asyncio.run(main(sys.argv[1:]))
  File "/usr/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/home/xartigas/repos/tss/symbol/mkdocs/.venv/lib/python3.12/site-packages/shoestring/__main__.py", line 50, in main
    await possible_task
  File "/home/xartigas/repos/tss/symbol/mkdocs/.venv/lib/python3.12/site-packages/shoestring/commands/renew_certificates.py", line 24, in run_main
    factory.reuse_ca_certificate(config.node.ca_common_name, directories.certificates)
  File "/home/xartigas/repos/tss/symbol/mkdocs/.venv/lib/python3.12/site-packages/shoestring/internal/CertificateFactory.py", line 115, in reuse_ca_certificate
    shutil.copy(ca_cert_path / 'ca.crt.pem', '.')
  File "/usr/lib/python3.12/shutil.py", line 435, in copy
    copyfile(src, dst, follow_symlinks=follow_symlinks)
  File "/usr/lib/python3.12/shutil.py", line 260, in copyfile
    with open(src, 'rb') as fsrc:
         ^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: 'keys/cert/ca.crt.pem'

$ ll keys/cert/

-r-------- 1 xartigas xartigas  485 Apr 23 12:45 ca.crt.pem
```

[...]

PR #2135, opened 2026-04-29
[tools/shoestring] fix: renew-certificates fails with relative path

Wayonb:
problem: renew-certificates fails to find the CA certificate file when a relative path is used
solution: convert the relative path to an absolute one, which allows finding the certs folder

[tools/shoestring] fix: file move fails
problem: shutil.move() uses copy2() function for cross-device files.
         copy fails to open the file as 'rw' since they are readonly
solution: remove the destination file if it exist before the move()

### DIFF
--- a/tools/shoestring/shoestring/internal/CertificateFactory.py
+++ b/tools/shoestring/shoestring/internal/CertificateFactory.py
@@ -196,13 +196,21 @@ class CertificateFactory:
 		full_crt += _read_file('ca.crt.pem')
 		_write_file('node.full.crt.pem', full_crt)
 
+	_CERTIFICATE_FILENAMES = ['node.crt.pem', 'node.key.pem', 'ca.pubkey.pem', 'ca.crt.pem', 'node.full.crt.pem']
+
 	@staticmethod
 	def package(output_directory, package_filter=''):
 		"""Creates a package of final files required for node deployment in the specifed output directory."""
 
-		CertificateFactory._package(output_directory, package_filter, [
-			'node.crt.pem', 'node.key.pem', 'ca.pubkey.pem', 'ca.crt.pem', 'node.full.crt.pem'
-		])
+		CertificateFactory._package(output_directory, package_filter, CertificateFactory._CERTIFICATE_FILENAMES)
+
+	@staticmethod
+	def remove_package_files(output_directory, package_filter=''):
+		"""Removes existing package certificate files."""
+
+		for filename in CertificateFactory._CERTIFICATE_FILENAMES:
+			if filename.startswith(package_filter):
+				(Path(output_directory) / filename).unlink(missing_ok=True)
 
 	def export_ca(self):
 		"""Exports the CA private key."""
@@ -216,8 +224,5 @@ class CertificateFactory:
 				continue
 
 			destination_path = Path(output_directory) / filename
-
-			# rm dest file if exist or move will fail to open
-			Path(destination_path).unlink(missing_ok=True)
 			shutil.move(filename, destination_path)
 			destination_path.chmod(0o400)
--- a/tools/shoestring/shoestring/internal/CertificateFactory.py
+++ b/tools/shoestring/shoestring/internal/CertificateFactory.py
@@ -16,6 +16,7 @@ class CertificateFactory:
 
 		self.temp_directory = None
 		self.original_working_directory = None
+		self._certificate_filenames = ['node.crt.pem', 'node.key.pem', 'ca.pubkey.pem', 'ca.crt.pem', 'node.full.crt.pem']
 
 	def __enter__(self):
 		self.temp_directory = tempfile.TemporaryDirectory()
@@ -196,19 +197,15 @@ class CertificateFactory:
 		full_crt += _read_file('ca.crt.pem')
 		_write_file('node.full.crt.pem', full_crt)
 
-	_CERTIFICATE_FILENAMES = ['node.crt.pem', 'node.key.pem', 'ca.pubkey.pem', 'ca.crt.pem', 'node.full.crt.pem']
-
-	@staticmethod
-	def package(output_directory, package_filter=''):
+	def package(self, output_directory, package_filter=''):
 		"""Creates a package of final files required for node deployment in the specifed output directory."""
 
-		CertificateFactory._package(output_directory, package_filter, CertificateFactory._CERTIFICATE_FILENAMES)
+		CertificateFactory._package(output_directory, package_filter, self._certificate_filenames)
 
-	@staticmethod
-	def remove_package_files(output_directory, package_filter=''):
+	def remove_package_files(self, output_directory, package_filter=''):
 		"""Removes existing package certificate files."""
 
-		for filename in CertificateFactory._CERTIFICATE_FILENAMES:
+		for filename in self._certificate_filenames:
 			if filename.startswith(package_filter):
 				(Path(output_directory) / filename).unlink(missing_ok=True)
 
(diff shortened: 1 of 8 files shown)


Wayonb:
fix #2129

segfaultxavi (approved):
It works as expected now with `--directory .`

tools/shoestring/shoestring/internal/CertificateFactory.py
```diff
@@ -216,5 +216,8 @@ def _package(output_directory, package_filter, package_filenames):
 				continue
 
 			destination_path = Path(output_directory) / filename
+
+			# rm dest file if exist or move will fail to open
+			Path(destination_path).unlink(missing_ok=True)
```
Jaguar0625: why do we need this? is this previous (expired) certificate?

tools/shoestring/tests/commands/test_renew_certificates.py
```diff
@@ -71,6 +71,8 @@ async def _assert_can_renew_node_certificate(ca_password=None, retain_key=False)
 		node_key_path = preparer.directories.certificates / 'node.key.pem'
 		node_key_data = _load_binary_file_data(node_key_path)
 
+		node_path = output_directory if not use_relative_path else str(Path(output_directory).relative_to(os.getcwd()))
```
Jaguar0625: [minor] given naming would probably make variable Path and use str below

tools/shoestring/shoestring/internal/CertificateFactory.py
Wayonb: Yeah, the previous certs are read-only.  For cross-device move, shutil.move() cannot use rename so the file has to be copied with copy2(default).  Copy2() will fail to open the destination files in "wb" mode because the files are read-only.

I can update the permissions also, if this is better?

```
-r-------- 1 wayonb wayonb  509 Mar  2 14:15 ca.crt.pem
-r-------- 1 wayonb wayonb  113 Mar  2 14:15 ca.pubkey.pem
-r-------- 1 wayonb wayonb  493 Mar  2 14:15 node.crt.pem
-r-------- 1 wayonb wayonb 1002 Mar  2 14:15 node.full.crt.pem
-r-------- 1 wayonb wayonb  119 Mar  2 14:15 node.key.pem
```

```
shoestring/internal/CertificateFactory.py:203: in package
    CertificateFactory._package(output_directory, package_filter, [
shoestring/internal/CertificateFactory.py:224: in _package
    shutil.move(filename, destination_path)
/usr/lib/python3.13/shutil.py:876: in move
    copy_function(src, real_dst)
/usr/lib/python3.13/shutil.py:468: in copy2
    copyfile(src, dst, follow_symlinks=follow_symlinks)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

src = 'node.crt.pem'
dst = PosixPath('/home/wayonb/work/code/symbol/product3/tools/shoestring/tmpzhhw5gjq/keys/cert/node.crt.pem')

    def copyfile(src, dst, *, follow_symlinks=True):
        """Copy data from src to dst in the most efficient way possible.
    
        If follow_symlinks is not set and src is a symbolic link, a new
        symlink will be created instead of copying the file it points to.
    
        """
        sys.audit("shutil.copyfile", src, dst)
    
        if _samefile(src, dst):
            raise SameFileError("{!r} and {!r} are the same file".format(src, dst))
    
        file_size = 0
        for i, fn in enumerate([src, dst]):
            try:
                st = _stat(fn)
            except OSError:
                # File most likely does not exist
                pass
            else:
                # XXX What about other special files? (sockets, devices...)
                if stat.S_ISFIFO(st.st_mode):
                    fn = fn.path if isinstance(fn, os.DirEntry) else fn
                    raise SpecialFileError("`%s` is a named pipe" % fn)
                if _WINDOWS and i == 0:
                    file_size = st.st_size
    
        if not follow_symlinks and _islink(src):
            os.symlink(os.readlink(src), dst)
        else:
            with open(src, 'rb') as fsrc:
                try:
>                   with open(dst, 'wb') as fdst:
                         ^^^^^^^^^^^^^^^
E                   PermissionError: [Errno 13] Permission denied: '/home/wayonb/work/code/symbol/product3/tools/shoestring/tmpzhhw5gjq/keys/cert/node.crt.pem'

/usr/lib/python3.13/shutil.py:262: PermissionError
```

Jaguar0625: i think i'd prefer the unlink/delete upstream in renew_certificates, potentially with --force flag (or pass a flag to package). i'd prefer to not have it default behavior in package.

tools/shoestring/tests/internal/test_CertificateFactory.py
```diff
@@ -314,6 +314,93 @@ def _load_binary_file_data(filename):
...
+			for filename in ['node.crt.pem', 'node.key.pem', 'node.full.crt.pem']:
+				self.assertTrue((Path(cert_directory) / filename).exists(), f'{filename} should still exist')
+
+	def test_remove_package_files_succeeds_when_files_do_not_exist(self):
+		# Arrange:
+		with tempfile.TemporaryDirectory() as cert_directory:
+			with tempfile.TemporaryDirectory() as ca_directory:
+				ca_key_path = self._create_ca_private_key(ca_directory)
+				with CertificateFactory(self._create_executor(), ca_key_path) as factory:
+
```
Jaguar0625: extra newline

tools/shoestring/tests/internal/test_CertificateFactory.py
```diff
@@ -314,6 +314,93 @@ def _load_binary_file_data(filename):
...
+	def test_remove_package_files_does_not_remove_non_certificate_files(self):
+		# Arrange:
+		with tempfile.TemporaryDirectory() as cert_directory:
+			self._create_cert_package(cert_directory)
+
+			extra_file = Path(cert_directory) / 'other.txt'
+			extra_file.write_text('keep me', encoding='utf8')
+
+			# Act:
+			self._remove_cert_package(cert_directory)
```
Jaguar0625, line 400: might want sanity check (e.g. number of files in dir) to prove that cert files exist

tools/shoestring/shoestring/internal/CertificateFactory.py
```diff
@@ -16,6 +16,7 @@ def __init__(self, openssl_executor, ca_key_path, ca_password=None):
 
 		self.temp_directory = None
 		self.original_working_directory = None
+		self._certificate_filenames = ['node.crt.pem', 'node.key.pem', 'ca.pubkey.pem', 'ca.crt.pem', 'node.full.crt.pem']
```
Jaguar0625: can be class field

tools/shoestring/tests/commands/test_renew_certificates.py
```diff
@@ -114,6 +117,10 @@ async def test_can_renew_node_certificate_with_ca_password():
 async def test_can_renew_node_certificate_with_retain_key():
 	await _assert_can_renew_node_certificate(retain_key=True)
 
+
+async def test_can_renew_node_certificate_with_relative_path():
+	await _assert_can_renew_node_certificate(use_relative_path=True, force=True)
```
Jaguar0625, line 135: do you want separate tests that individually set test relative_path and force?

Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #2136, opened 2026-04-30
[explorer/rest] feat: Added transaction list endpoint

AnthonyLaw:
Problem: Missing transaction list endpoint.
Solution: 
- Added  `/transactions` endpoint, it support `pagination` and `sort by height`. 
- It support filters `height`, `transaction_types`, `address`, `sender address`, `recipient address`, `sender public key` and `mosaic`.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -243,7 +243,7 @@ class NemDatabase(DatabaseConnection):
 				id serial PRIMARY KEY,
 				transaction_id int NOT NULL,
 				namespace_name varchar(146),
-				quantity numeric,
+				quantity bigint,
 				FOREIGN KEY (transaction_id) REFERENCES transactions(id)
 				ON DELETE CASCADE
 			)
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -230,7 +230,6 @@ class NemDatabase(DatabaseConnection):
 				timestamp timestamp NOT NULL,
 				deadline timestamp NOT NULL,
 				signature bytea,
-				amount bigint,
 				is_inner boolean DEFAULT false,
 				payload jsonb
 			)
@@ -244,7 +243,7 @@ class NemDatabase(DatabaseConnection):
 				id serial PRIMARY KEY,
 				transaction_id int NOT NULL,
 				namespace_name varchar(146),
-				quantity bigint,
+				quantity numeric,
 				FOREIGN KEY (transaction_id) REFERENCES transactions(id)
 				ON DELETE CASCADE
 			)
@@ -493,11 +492,10 @@ class NemDatabase(DatabaseConnection):
 				timestamp,
 				deadline,
 				signature,
-				amount,
 				is_inner,
 				payload
 			)
-			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
 			RETURNING id
 			''',
 			(
@@ -511,7 +509,6 @@ class NemDatabase(DatabaseConnection):
 				transaction.timestamp,
 				transaction.deadline,
 				unhexlify(transaction.signature) if transaction.signature else None,
-				transaction.amount,
 				transaction.is_inner,
 				json.dumps(transaction.payload) if transaction.payload else None
 			)
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -72,7 +72,6 @@ TransactionRecord = namedtuple('TransactionRecord', [
 	'timestamp',
 	'deadline',
 	'signature',
-	'amount',
 	'transaction_type',
 	'is_inner',
 	'sender_address',
@@ -369,7 +368,6 @@ class NemPuller:
 
 		payload = None
 		recipient_address = None
-		amount = None
 		if transaction.transaction_type == TransactionType.TRANSFER.value:
 			payload = {
 				'message': {
@@ -378,7 +376,6 @@ class NemPuller:
 				} if transaction.message else None
 			}
 			recipient_address = transaction.recipient
-			amount = transaction.amount
 		elif transaction.transaction_type == TransactionType.ACCOUNT_KEY_LINK.value:
 			payload = {
 				'mode': transaction.mode,
@@ -452,7 +449,6 @@ class NemPuller:
 			fee=transaction.fee,
 			timestamp=self._convert_timestamp_to_datetime(transaction.timestamp),
 			deadline=self._convert_timestamp_to_datetime(transaction.deadline),
-			amount=amount,
 			signature=transaction.signature,
 			transaction_type=transaction.transaction_type,
 			is_inner=is_inner,
@@ -469,9 +465,21 @@ class NemPuller:
 		transaction_id = self.nem_db.insert_transaction(cursor, transaction_record)
 
 		if transaction.transaction_type == TransactionType.TRANSFER.value:
-			mosaics = transaction.mosaics or [Mosaic('nem.xem', 1000000)]
-			for mosaic in mosaics:
-				self.nem_db.insert_transaction_mosaic(cursor, transaction_id, mosaic)
+			if transaction.mosaics:
+				for mosaic in transaction.mosaics:
+					mosaic_amount = mosaic.quantity * transaction.amount / (10 ** 6)
+					self.nem_db.insert_transaction_mosaic(
+						cursor,
+						transaction_id,
+						Mosaic(mosaic.namespace_name, mosaic_amount)
+					)
+			else:
+				# handle v1 transfer transaction
+				self.nem_db.insert_transaction_mosaic(
+					cursor,
+					transaction_id,
+					Mosaic('nem.xem', transaction.amount)
+				)
 		elif transaction.transaction_type == TransactionType.NAMESPACE_REGISTRATION.value:
 			self._process_namespace(cursor, transaction, block_height)
 		elif transaction.transaction_type == TransactionType.MOSAIC_DEFINITION.value:
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -10,6 +10,7 @@ from symbolchain.nc import TransactionType
 from symbolchain.nem.Network import Address, Network
 from symbollightapi.connector.NemConnector import NemConnector
 from symbollightapi.model.Exceptions import NodeException
+from symbollightapi.model.Transaction import Mosaic
 from zenlog import log
 
 from puller.db.NemDatabase import NemDatabase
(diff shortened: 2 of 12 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -638,3 +642,114 @@ def get_transaction_by_hash(self, transaction_hash, is_inner=False):
...
+			filter_params.append(tuple(transaction_query.transaction_types))
+
+		if transaction_query.address:
+			where_condition += ' AND (t.sender_address = %s OR t.recipient_address = %s)'
+			filter_params.extend([transaction_query.address.bytes, transaction_query.address.bytes])
+		else:
+			if transaction_query.sender_address:
+				where_condition += ' AND t.sender_address = %s'
+				filter_params.append(transaction_query.sender_address.bytes)
+			elif transaction_query.sender:
```
Wayonb: Is the sender and the sender_address the same?  Are both needed?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -638,3 +642,114 @@ def get_transaction_by_hash(self, transaction_hash, is_inner=False):
...
+
+		if transaction_query.mosaic:
+			where_condition += ' AND t.transaction_type = %s'
+
+			if transaction_query.mosaic.upper() == 'NEM.XEM':
+				where_condition += (
+					' AND ( '
+					'	EXISTS (SELECT 1 FROM transactions_mosaic tm WHERE tm.transaction_id = t.id AND tm.namespace_name = %s)'
+					'	OR'
+					'	NOT EXISTS (SELECT 1 FROM transactions_mosaic tm WHERE tm.transaction_id = t.id)'
```
Wayonb: Why this?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -638,3 +642,114 @@ def get_transaction_by_hash(self, transaction_hash, is_inner=False):
...
+			params,
+			where_condition=where_condition,
+			order_condition=order_condition,
+			limit_condition=limit_condition
+		)
+
+		inner_transaction_hashes = [
+			transaction.payload['inner_hash']
+			for transaction in transactions
+			if transaction.transaction_type == TransactionType.MULTISIG.value and transaction.payload.get('inner_hash')
```
Wayonb: Is there a case where transcation is MUTLISIG but not have ``inner_hash``?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -638,3 +642,114 @@ def get_transaction_by_hash(self, transaction_hash, is_inner=False):
...
+		if inner_transaction_hashes:
+			inner_transactions = self._get_inner_transactions(inner_transaction_hashes)
+			inner_transaction_map = {
+				_format_bytes(inner_transaction.transaction_hash).lower(): inner_transaction for inner_transaction in inner_transactions
+			}
+
+			return [
+				self._create_transaction_view(
+					transaction,
+					inner_transaction_map.get(transaction.payload['inner_hash']) if transaction.payload.get('inner_hash') else None
```
Wayonb: Doesn't the get() returns None if not found? I dont think the if else is needed 🤔

explorer/rest/rest/__init__.py
```diff
@@ -245,6 +247,87 @@ def api_get_nem_transaction_by_hash(transaction_hash):
...
+					TransactionType.MULTISIG.name,
+					TransactionType.NAMESPACE_REGISTRATION.name,
+					TransactionType.MOSAIC_DEFINITION.name,
+					TransactionType.MOSAIC_SUPPLY_CHANGE.name
+				}
+
+				transaction_types = [tx_type.upper() for tx_type in transaction_types.split(',')]
+
+				for tx_type in transaction_types:
+					if tx_type not in valid_transaction_types:
```
Wayonb: should be able to do ``if tx_type not in TransactionType.__members__:``

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: `sender` is the public key, and I converted it into an address.

User provided sender public key, when processing the query, I converted it into an address, and searched in DB, because I added Address index, instead of public key.

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Because v1 transfer transactions are not treated as mosaic, v1 transfer txs are only stored in the `transaction` table and are not inserted into the `transactions_mosaic` table

This case only happens in `nem.xem`.

1. filter all the `nem.xem` mosaic from the `transactions_mosaic` table. (v2 transfer transaction)

2. It returns only transfer tx with no mosaic row, which represents `nem.xem` transfer

With this approach, I can get the full list of `nem.xem` transactions (both v1 and v2 transfer)

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: nope, I will remove `transaction.payload.get('inner_hash')` logic.

explorer/rest/rest/__init__.py
AnthonyLaw: Good to know this keyword `__members__`

explorer/rest/rest/db/NemDatabase.py
Wayonb: So the sender and sender_address should be the same? 🤔

explorer/rest/rest/facade/NemRestFacade.py
```diff
@@ -145,3 +145,17 @@ def get_transaction_by_hash(self, transaction_hash):
...
 		return transaction.to_dict() if transaction else None
+
+	def get_transactions(self, pagination, sort, transaction_query):
+		"""Gets transactions pagination."""
+
+		transaction_query = transaction_query._replace(
+			address=Address(transaction_query.address) if transaction_query.address else None,
+			sender_address=Address(transaction_query.sender_address) if transaction_query.sender_address else None,
+			recipient_address=Address(transaction_query.recipient_address) if transaction_query.recipient_address else None,
+			sender=self.network.public_key_to_address(PublicKey(transaction_query.sender)) if transaction_query.sender else None
```
Wayonb: Should the api fail if the user provides both a sender_address and a sender?  If they are different accounts, the returned results might not be expected

Wayonb: Also, I think when the sender's public key is used, it can be stored as ``sender_address`` in the query param, so both checks are not needed in the DB.

explorer/rest/rest/db/NemDatabase.py
Wayonb: I prefer not to check for non-existent conditions.  What happens if the transaction type does not have a mosaic and the user created the query incorrectly? 🤔 
I would prefer on the puller side to insert an entry into the mosaic table for transfer v1.  This will allow users to easily retrieve all transactions for a specific mosaic by querying the mosaic table and joining it with the transaction table.
Let me know what you think.

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Yes, I will make some refactoring for that.
Just using `sender_address` will be enough.

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: I think that sounds good, treat all transfers as v2.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -467,8 +468,9 @@ def _process_transaction(self, cursor, transaction, block_height, is_inner):
 		transaction_record = self._build_transaction_record(transaction, is_inner)
 		transaction_id = self.nem_db.insert_transaction(cursor, transaction_record)
 
-		if transaction.transaction_type == TransactionType.TRANSFER.value and transaction.mosaics:
-			for mosaic in transaction.mosaics:
+		if transaction.transaction_type == TransactionType.TRANSFER.value:
+			mosaics = transaction.mosaics or [Mosaic('nem.xem', 1000000)]
```
Wayonb: Why 1000000 xem for transfer v1?

Wayonb (changes requested):

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: The reason is because of the mosaic amount calculation.

Currently, the calculation logic for v2 is `quality * amount` 

For a v1 transfer transaction, `amount` is always the actual transfer amount. That why I added  `1000000` to make sure the calculation is correct.

cryptoBeliever: @AnthonyLaw @Wayonb 
I wonder if it would be better to remove the `amount` field and keep only the mosaics in the database? In the puller, we could convert the v1 format and save it directly as mosaics. Do we actually need the amount field for anything? What do you think?

AnthonyLaw: > @AnthonyLaw @Wayonb I wonder if it would be better to remove the `amount` field and keep only the mosaics in the database? In the puller, we could convert the v1 format and save it directly as mosaics. Do we actually need the amount field for anything? What do you think?

Yes, we can remove the `amount`. 
The purpose of the `amount` is just used for the v1 transaction.

I can just do the calculation (puller) for the mosaic amount and store it directly in the DB.

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -237,7 +236,7 @@ def create_tables(self):
 				id serial PRIMARY KEY,
 				transaction_id int NOT NULL,
 				namespace_name varchar(146),
-				quantity bigint,
+				quantity numeric,
```
cryptoBeliever: Is there a reason you changed it to NUMERIC? I assume it’s still stored in the smallest unit without a decimal point? Or is it already adjusted for divisibility?

AnthonyLaw: > Is there a reason you changed it to NUMERIC? 

Previously, `quantity` stored the absolute value.

So right now we removed the `amount` field, which means we will store `xem relative(amount) * quantity (absolute).`
which is possible, there are decimal values.

>  I assume it’s still stored in the smallest unit without a decimal point?

I don't think we can achieve it, if `absolute xem amount * quantity` and adjust for mosaic divisibility, which i dont think the value is accurate, but let me know if you have an alternative solution.

cryptoBeliever: I think that the quantity is already in the smallest unit, so it cannot contain decimal values. Even if a calculation from a transaction produces a decimal value, the fractional part should be ignored.

I created an example transaction as a test:

/transaction/get?hash=23c173b808711ad1c7eb9b28c056bdcba3e02d7468028695a55027ae58bc8007

```
{
   "meta":{
      "innerHash":{
         
      },
      "id":3630,
      "hash":{
         "data":"23c173b808711ad1c7eb9b28c056bdcba3e02d7468028695a55027ae58bc8007"
      },
      "height":619881
   },
   "transaction":{
      "timeStamp":351540564,
      "amount":1999999,
      "signature":"15fc0314883febd316b42b631f804bb04760d4b3603d3754eec3d97282dd91ee1fc110d19fd299bcaa190b1d2281512a7b74b8f6bba289bc2d25c1499adf860a",
      "fee":200000,
      "recipient":"TCHKZYGSODJSYRY3UVRDSN267BPW7IOVAAJ3GGWN",
      "mosaics":[
         {
            "quantity":1,
            "mosaicId":{
               "namespaceId":"nem",
               "name":"xem"
            }
         }
      ],
      "type":257,
      "deadline":351623364,
      "message":{
         "payload":"54657374206d657373616765",
         "type":1
      },
      "version":-1744830462,
      "signer":"408347579b9df6f698d926272830104a887659194ec5cf7c44f823b954111618"
   }
}
```

amount = 1999999
quantity = 1

total_quantity = 1999999 / 1000000 * 1 = 1,999999 microxem. 
Because the decimal point is ignored in the balance, only 1 microxem is taken, so we should persist quantity = 1.

Please correct me if I'm wrong.

Wayonb (approved):
👍🏾

explorer/puller/puller/db/NemDatabase.py
cryptoBeliever: Looks good now :+1:

cryptoBeliever (approved):

PR #2159, opened 2026-05-11
[wallet/symbol/mobile] feat: add Uniswap integration

OlegMakarenko:
## Problem

The existing bridge implementation was designed around one-step, bridge-only flows. That worked for XYM <-> wXYM, but it does not fit XYM/ETH because the two directions are not symmetrical:

- XYM -> ETH can be handled as a single bridge step.
- ETH -> XYM requires two different execution mechanisms: first swap ETH -> wXYM on Uniswap, then bridge wXYM -> XYM to Symbol.

The previous bridge abstraction also handled both directions inside one manager, while the transaction send UI assumed a single workflow running against a single controller. That made estimation, confirmation preview, progress tracking, and activity logging difficult to extend to composed routes.

## Solution

This PR introduces a workflow-based swap architecture so that each swap step is represented by a dedicated pair manager with a shared interface for loading, estimation, history, and transaction creation. A new `UniswapPairManager` was added to mirror the `BridgePairManager` contract, which allows the UI and workflow layer to treat a Uniswap step and a bridge step the same way.

Instead of keeping a single bridge object that switches between directions internally, routes are now modeled as fixed-direction pair managers composed into a `SwapWorkflowManager`. This makes the route definition explicit and allows us to support both single-step and multi-step flows without adding route-specific logic to the screen layer.

With this change, XYM -> ETH remains a one-step bridge flow, while ETH -> XYM becomes a true two-step workflow:
1. Swap ETH -> wXYM on Ethereum through Uniswap.
2. Bridge wXYM -> XYM from Ethereum to Symbol.

The transaction send flow and UI were refactored accordingly. `TransactionScreenTemplate` now supports custom workflows and custom transaction-progress view models, which makes it possible to show multi-step confirmation previews, extended activity logs, and better status tracking for composed swaps.

### New Route Model

| Route | Workflow | Steps |
| --- | --- | --- |
| XYM -> wXYM | `BridgePairManager` | 1 |
| wXYM -> XYM | `BridgePairManager` | 1 |
| XYM -> ETH | `BridgePairManager` | 1 |
| ETH -> XYM | `UniswapPairManager` + `BridgePairManager` | 2 |

### Why The Architecture Changed

- Bridge execution is now modeled as direction-specific pair managers instead of one manager that internally toggles between wrap and unwrap behavior.
- `UniswapPairManager` implements the same interface as `BridgePairManager`, so estimation and transaction creation can be reused across bridge and DEX steps.
- `SwapWorkflowManager` wraps one or more pair managers for a single source -> target route, which makes route composition explicit and keeps the UI generic.
- This is what enables ETH -> XYM to be expressed as `Uniswap + Bridge` instead of forcing special-case logic into the swap screen.

### How Transaction Execution Works

- One-step workflow:
  1. Create one transaction bundle
  2. Show confirmation preview
  3. Sign and announce
  4. Poll for confirmation

- Two-step workflow:
  1. Create a transaction bundle
  2. Show confirmation preview for both steps
  3. Sign and announce step 1
  4. Wait for step 1 confirmation
  5. Sign and announce step 2 automatically
  6. Track both steps in the progress dialog and activity log

### Uniswap-Specific Changes

- Added `UniswapPairManager` as a fixed-direction pair manager for on-chain swaps.
- Added multi-network configuration support so router, quoter, WETH, token, and fee settings can be resolved per network.
- Added support for native ETH swaps.
- For native ETH input, only the swap transaction is created.
- For ERC20 input, the bundle includes `approve + swap`.
- Token/address normalization was tightened to avoid case-related issues in pair selection and contract calls.

### UI Refactor

The transaction send UI was refactored so `TransactionScreenTemplate` no longer assumes a single standard send flow. It can now work with custom workflows, custom progress models, and custom confirmation data.

This was required for dual-step swaps because the UI now needs to:
- preview more than one transaction bundle
- track more than one sign/announce/confirm cycle
- show a richer activity log for composed flows
- keep a reusable send template for non-bridge screens

This refactor also updates the bridge screens to operate on workflow-based routes instead of the previous bridge-only model.

### Main UI Areas Updated

| Area | Change |
| --- | --- |
| `TransactionScreenTemplate` | generalized to support custom workflows, custom progress view models, and custom send flows |
| Bridge Swap screen | moved to workflow-based pair selection, sequential estimation, and step-aware transaction creation |
| Bridge Swap details / activity log | extended to present multi-step progress more clearly |
| Confirmation preview | updated to support one or multiple transaction bundles |
| Other template consumers | aligned with the refactored template API so existing send flows continue to work |

### Screens/Flows Affected By The Template Refactor

- Send
- Harvesting
- Create Multisig Account
- Modify Multisig Account
- Bridge Swap

### Main Change Areas

| Area | Summary |
| --- | --- |
| core | refactor from bidirectional bridge manager to `BridgePairManager`, add `SwapWorkflowManager`, clean up bridge types and APIs |
| ethereum | add Uniswap support, native ETH handling, multi-network Uniswap config, token normalization, Sepolia-related updates |
| mobile controller/config | wire XYM/wXYM/ETH routes through pair managers and workflows, update config for Sepolia |
| mobile bridge UI | support workflow-based selection, estimation, send flow, and dual-step activity tracking |
| tests | update template, bridge, and swap tests to match the new abstractions |

### Network And Performance Updates

- Updated Ethereum testnet support to Sepolia.
- Simplified Ethereum pair/network handling to mainnet + testnet.
- Removed the Ethereum unconfirmed listener that was causing unnecessary load and slow SwapScreen readiness on Sepolia.


wallet/symbol/mobile/src/components/templates/TransactionScreenTemplate/utils/transaction-progress.js
```diff
@@ -0,0 +1,208 @@
...
+	};
+	const sendingStatus = {
+		icon: 'pending',
+		variant: 'warning',
+		title: $t('c_transactionStatus_status_sending_title'),
+		description: $t('c_transactionStatus_status_sending_description')
+	};
+
+	const infoMap = {
+		[TransactionWorkflowStatus.IDLE]: sendingStatus,
```
Wayonb, line 34: Why is IDLE status sending? 🤔

Wayonb (approved):
👍🏾

wallet/symbol/mobile/src/components/templates/TransactionScreenTemplate/utils/transaction-progress.js
OlegMakarenko: There is a small delay between the popup appearing and the transaction sending process starting. Even though it is correctly reported as IDLE, there is no reason to show this state to the user for such a brief moment on the UI. 

This part maps the logic state to the visual state.

PR #2163, opened 2026-05-12
[explorer/rest] feat: Added unconfirmed transaction endpoint

AnthonyLaw:
Problem: Missing unconfirmed transaction in the endpoint.
Solution: 
- Query unconfirmed transactions from the node connector.
- Converted node connector response to DB query result, so that it can reuse the existing method to create a transaction view.
- Unconfirmed transactions weren't stored in the DB.

### DIFF
--- a/explorer/rest/rest/db/NemDatabase.py
+++ b/explorer/rest/rest/db/NemDatabase.py
@@ -219,7 +219,7 @@ class NemDatabase(DatabaseConnectionPool):
 			value = None
 			embedded_transaction.append({
 				'initiator': _format_address_bytes_to_string(transaction.from_address),
-				'transactionHash': _format_bytes(inner_transaction.transaction_hash),
+				'transactionHash': _format_bytes(inner_transaction.transaction_hash) if inner_transaction.transaction_hash else None,
 				'transactionType': TransactionType(inner_transaction.transaction_type).name,
 				'signatures': [{
 					'fee': _format_xem_relative(signature['fee']),
@@ -237,7 +237,7 @@ class NemDatabase(DatabaseConnectionPool):
 			to_address = _format_address_bytes_to_string(inner_transaction.to_address) if inner_transaction.to_address else None
 
 		return TransactionView(
-			transaction_hash=_format_bytes(transaction.transaction_hash),
+			transaction_hash=_format_bytes(transaction.transaction_hash) if transaction.transaction_hash else None,
 			transaction_type=TransactionType(transaction.transaction_type).name,
 			from_address=from_address,
 			to_address=to_address,
@@ -247,7 +247,7 @@ class NemDatabase(DatabaseConnectionPool):
 			height=transaction.height,
 			timestamp=str(transaction.timestamp),
 			deadline=str(transaction.deadline),
-			signature=_format_bytes(transaction.signature)
+			signature=_format_bytes(transaction.signature) if transaction.signature else None
 		)
 
 	@staticmethod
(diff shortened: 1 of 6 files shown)


explorer/rest/rest/db/NemDatabase.py
```diff
@@ -196,7 +196,7 @@ def _create_transaction_view(self, transaction, inner_transaction=None):
 			value = None
 			embedded_transaction.append({
 				'initiator': _format_address_bytes_to_string(transaction.from_address),
-				'transactionHash': _format_bytes(inner_transaction.transaction_hash),
+				'transactionHash': _format_bytes(inner_transaction.transaction_hash) if inner_transaction.transaction_hash else None,
```
Wayonb, line 223: Is the inner transaction hash missing from the unconfirmed tx?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -215,7 +215,7 @@ def _create_transaction_view(self, transaction, inner_transaction=None):
 			to_address = _format_address_bytes_to_string(inner_transaction.to_address) if inner_transaction.to_address else None
 
 		return TransactionView(
-			transaction_hash=_format_bytes(transaction.transaction_hash),
+			transaction_hash=_format_bytes(transaction.transaction_hash) if transaction.transaction_hash else None,
```
Wayonb, line 241: the same for transactions? 🤔

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -742,3 +742,167 @@ def get_transactions(self, pagination, sort, transaction_query):
...
+	def _convert_timestamp_to_datetime(self, timestamp):
+		return self.network.datetime_converter.to_datetime(timestamp).strftime('%Y-%m-%d %H:%M:%S')
+
+	def _create_unconfirmed_mosaic_records(self, mosaics):
+		if not mosaics:
+			# create default nem.xem mosaic record for v1 transfer transaction
+			return [
+				{
+					'namespace_name': 'nem.xem',
+					'quantity': 1000000,
```
Wayonb: I am assuming the ``quantity`` will be removed as part of the other PR?

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -742,3 +742,167 @@ def get_transactions(self, pagination, sort, transaction_query):
...
+		amount = None
+		mosaics = None
+		if transaction.transaction_type == TransactionType.TRANSFER.value:
+			payload = {
+				'message': {
+					'payload': transaction.message.payload,
+					'is_plain': transaction.message.is_plain,
+				} if transaction.message else None
+			}
+			recipient_address = transaction.recipient.bytes if transaction.recipient else None
```
Wayonb: When is the recipient not present for transfer? 🤔

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -742,3 +742,167 @@ def get_transactions(self, pagination, sort, transaction_query):
...
+			recipient_address = transaction.creation_fee_sink
+		elif transaction.transaction_type == TransactionType.MOSAIC_SUPPLY_CHANGE.value:
+			payload = {
+				'namespace_name': transaction.namespace_name,
+				'supply_type': transaction.supply_type,
+				'delta': transaction.delta
+			}
+
+		return TransactionRecord(
+			transaction_hash=bytes.fromhex(transaction.transaction_hash) if transaction.transaction_hash else None,
```
Wayonb: Here also, why is the transaction_hash optional?

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: yes.

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Yes, I can't see any transaction hash in the unconfirmed transaction.

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Yes, I will rebase it once it is merged.

explorer/rest/rest/db/NemDatabase.py
Wayonb: ok.  is this required for the UI?

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: good catch!

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Alright, actually, the transaction hash will be `none` in this case (unconfirmed tx).
I'll update it, also rename `_build_transaction_record` -> `_build_unconfirmed_transaction_record.`

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: Nope, The UI component will check if the hash exist only render the hash component.

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -249,7 +250,7 @@ def _create_transaction_view(self, transaction, inner_transaction=None):
 			height=transaction.height,
 			timestamp=str(transaction.timestamp),
 			deadline=str(transaction.deadline),
-			signature=_format_bytes(transaction.signature)
+			signature=_format_bytes(transaction.signature) if transaction.signature else None
```
Wayonb: Miss this before, shouldn't the ``signature`` be part of the transaction?

explorer/rest/rest/db/NemDatabase.py
Wayonb: Will search by transaction hash not be supported for unconfirmed transactions? Discoverability of the transaction is unlikely to be a big issue, since traffic on NEM is low.
The only other concern would be tools.  And how will this work with Faucet when you click on the transaction link?

Wayonb (changes requested):

explorer/rest/rest/db/NemDatabase.py
AnthonyLaw: You are right, I was confused with the inner transaction.

Wayonb (approved):
👍🏾

PR #2164, opened 2026-05-13
[lightapi/python] fix: throw error when `relativeChange` is missing

AnthonyLaw:
Problem: In multisig_account_modification transactions, the minCosignatories field can be either {} or contain relativeChange.

Solution: Handle missing relativeChange by defaulting min_approval_delta to 0.


Wayonb (approved):
👍🏾

PR #2165, opened 2026-05-14
[lightapi/python] fix: remove status, the status is node specific

AnthonyLaw:
Problem: The status is node-specific, which is not related to account status.
Solution: removed status


Wayonb (changes requested):
I don't think this needs to be removed from the API.  It's up to the consumer if they want to use it or not.

AnthonyLaw:
> I don't think this needs to be removed from the API. It's up to the consumer if they want to use it or not.

Sounds fair.

PR #2190, opened 2026-05-20
[explorer] fix: removed account status

AnthonyLaw:
Problem: Explorer does not use `status`.
Solution: Remove `status`


Wayonb (approved):
👍🏾

PR #2195, opened 2026-05-23
[explorer/puller] feat: Added refresh account workflow

AnthonyLaw:
Problem: Vested balance and importance are recalculated by the node on a schedule. However, the puller does not update those accounts in the database if no transaction happens.

Solution: Added a new workflow script that can update accounts where balance, vested_balance, or importance is greater than 0. We can set up a cron job to run it daily.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -299,6 +299,29 @@ class NemPuller:
 				last_height
 			)
 
+	async def refresh_accounts(self, batch_size):
+		"""Refresh vested balance and importance for stored accounts."""
+
+		cursor = self.nem_db.connection.cursor()
+		last_account_id = 0
+		total_refreshed = 0
+
+		while True:
+			accounts = self.nem_db.get_accounts_for_refresh(batch_size, last_account_id)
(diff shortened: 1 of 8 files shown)


explorer/puller/puller/facade/NemPuller.py
```diff
@@ -299,6 +299,29 @@ def _process_harvested_fees(self, cursor, accounts_harvested_fee):  # pylint: di
...
+		cursor = self.nem_db.connection.cursor()
+		last_account_id = 0
+		total_refreshed = 0
+
+		while True:
+			accounts = self.nem_db.get_accounts_for_refresh(batch_size, last_account_id)
+			if not accounts:
+				break
+
+			for account in accounts:
```
Wayonb, line 314: How fast is the account refresh on mainnet?

Wayonb (approved):
👍🏾

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Actually, I have no idea. We can try it, since @cryptoBeliever has synced db with the mainnet node.

PR #2209, opened 2026-05-26
[explorer] feat: Refactor nem explorer backend

AnthonyLaw:
Problem: old codebase is not working well, refactor and unit test is needed. 
Solution: Refactor the nem backend codebase, which included Puller and REST.


Wayonb (approved):
👍🏾

PR #2217, opened 2026-05-27
[explorer/frontend] feat: add Symbol frontend variant shell

daoka:

The explorer frontend is effectively NEM-only. Public modules under `explorer/frontend/api/*` contain the NEM API mapping directly, so adding Symbol support requires editing shared frontend code and makes the Symbol work too large to review safely.

The frontend also assumes a single default asset/config/theme path. There is no variant boundary for selecting NEM or Symbol runtime behavior from `NEXT_PUBLIC_PLATFORM` / `PLATFORM`.

## What's the issue?

Symbol Explorer work needs to be split into smaller PRs. Without a frontend variant boundary, follow-up Symbol pages and adapters would mix Symbol-specific behavior into shared code or into the existing NEM implementation.

This PR is intended to provide only the minimum frontend shell needed for Symbol startup and review:
- preserve NEM as the default behavior
- add a variant selection boundary
- add a minimal Symbol runtime shell
- keep account lists, richer block/transaction details, REST API/Puller integration, and feature-complete Symbol pages for follow-up PRs

Added frontend variant selection for `nem` and `symbol`.

- Added `explorer/frontend/variants/index.js` to resolve the active variant from `NEXT_PUBLIC_PLATFORM || PLATFORM || 'nem'`.
- Moved the current NEM API implementations into `variants/nem/api/*`.
- Kept the existing `api/*.js` modules as compatibility wrappers so existing imports continue to work.
- Verified `variants/nem/api/*` matches the corresponding `dev` `api/*` implementation files.

Added a minimal Symbol runtime shell.

- Added Symbol config, `DocumentHead`, page config, theme variables, fonts, and runtime assets.
- Added minimal Symbol API adapters for current shell rendering.
- Added a temporary Symbol node proxy for frontend-only shell validation. This is intentionally narrow and temporary; the long-term direction remains REST API / Puller parity with NEM.
- Hardened the temporary proxy with GET-only behavior, path/query allowlists, request timeout handling, and page-size clamping.
- Avoided misleading output for unsupported temporary Symbol filters by returning empty pages instead of forwarding unsupported filters as unfiltered requests.
- Added Symbol transaction type coverage for the supported Symbol transaction type list.

Added build/runtime support needed by Symbol.

- Added Symbol SDK browser build support and frontend-safe stubs for optional Node/WASM-only paths.
- Added Symbol address conversion support used by the runtime shell.
- Added local image pattern support for Symbol assets.
- Escaped inline app config bootstrap data before injecting it into the page.

This is not intended to be a complete Symbol Explorer implementation. Feature-complete Symbol list/detail behavior remains follow-up work.


explorer/frontend/package.json
```diff
@@ -33,6 +34,8 @@
     "react-toastify": "^11.0.5",
     "sass": "^1.49.0",
     "sharp": "^0.34.4",
+    "symbol-crypto-wasm-web": "^0.1.1",
+    "symbol-sdk": "^3.3.1",
```
AnthonyLaw, line 38: Why do you need symbol-sdk?

explorer/frontend/public/symbol/images/blocks/block.svg
AnthonyLaw, line 1: Some assets can be shared across both NEM and Symbol.

You could structure it like this:

```text
public/common
public/nem
public/symbol
```

For example, shared assets can go under `public/common`, while NEM-specific and Symbol-specific assets can go under their own folders.

explorer/frontend/variants/index.js
```diff
@@ -0,0 +1,74 @@
...
+		pageConfig
+	};
+};
+
+const loadVariant = () => {
+	switch (platform) {
+	case 'symbol':
+		return loadSymbolVariant();
+	case 'nem':
+	default:
```
AnthonyLaw, line 66: I think we should only allow users to select either `symbol` or `nem`, without setting a default value.

explorer/frontend/variants/index.js
```diff
@@ -0,0 +1,74 @@
...
+
+	return apiModule[exportName](...args);
+};
+
+const createApi = variantName => ({
+	fetchAccountPage: call(() => import(`./${variantName}/api/accounts`), 'fetchAccountPage'),
+	fetchAccountInfo: call(() => import(`./${variantName}/api/accounts`), 'fetchAccountInfo'),
+	fetchAccountInfoByPublicKey: call(() => import(`./${variantName}/api/accounts`), 'fetchAccountInfoByPublicKey'),
+	fetchBlockPage: call(() => import(`./${variantName}/api/blocks`), 'fetchBlockPage'),
+	fetchChainHight: call(() => import(`./${variantName}/api/blocks`), 'fetchChainHight'),
```
AnthonyLaw, line 14: typo: fetchChainHight -> fetchChainHeight

explorer/frontend/variants/index.js
AnthonyLaw, line 1: I think this mixes `import` and `require`. We should use import consistently instead of mixing both styles.

explorer/frontend/variants/index.js
```diff
@@ -0,0 +1,74 @@
+const platform = process.env.NEXT_PUBLIC_PLATFORM || process.env.PLATFORM || 'nem';
+
+const call = (loadModule, exportName) => async (...args) => {
+	const apiModule = await loadModule();
+
+	return apiModule[exportName](...args);
+};
+
+const createApi = variantName => ({
```
AnthonyLaw, line 9: We can import each variant API explicitly

```
import * as nemAPI from './nem/api';
import * as symbolAPI from './symbol/api';
```

This would make the code cleaner.

explorer/frontend/variants/symbol/api/search.js
```diff
@@ -0,0 +1,30 @@
...
+		},
+		{
+			type: 'mosaics',
+			data: null
+		},
+		{
+			type: 'namespaces',
+			data: null
+		}
+	];
```
AnthonyLaw, line 29: You’re returning a different data shape here.

Symbol search returns an array, but the shared UI consumes search results as an object, for example `result.block`.

We need to align Symbol search with the same return shape as NEM search.

explorer/frontend/variants/symbol/api/namespaces.js
```diff
@@ -0,0 +1,36 @@
+import { createSymbolNodePath, createSymbolPage, createSymbolSearchURL, fetchSymbolNode, hexToSymbolAddress } from '../utils';
+import config from '@/config';
+import { createTryFetchInfoFunction } from '@/utils/server';
+
+const namespaceInfoFromDTO = data => {
```
AnthonyLaw, line 5: I think most of the API responses use different DTO shapes.
How are you planning to handle that in the shared UI components?

explorer/frontend/variants/symbol/config/index.js
```diff
@@ -0,0 +1,12 @@
+export { default as pageConfig } from './pages';
+
+const env = name => process.env[`NEXT_PUBLIC_${name}`] || process.env[name];
+
+export const config = {
+	SYMBOL_STATISTIC_SERVICE_URL: env('SYMBOL_STATISTIC_SERVICE_URL'),
```
AnthonyLaw, line 6: We already migrated to the `nodeWatch` service.

explorer/frontend/package.json
daoka: I had intended to use this when converting from a public key to an address, but I never ended up using it and forgot to delete it. I’ve removed it in the new PR #2238

explorer/frontend/public/symbol/images/blocks/block.svg
daoka: For now, I kept the existing `/images/**` path as the shared/legacy asset location to avoid moving assets in this PR. I also added variant-specific image locations to the Next.js image config:

- common/legacy: `/images/**`
- NEM: `/nem/images/**`
- Symbol: `/symbol/images/**`

I added the NEM image directory placeholder without moving existing assets. Actual asset relocation and component path updates can be handled separately when we split shared/NEM/Symbol assets more fully.

Handled in: 1bd5b88

explorer/frontend/variants/index.js
daoka: Updated. The app-level variant selection now requires `NEXT_PUBLIC_PLATFORM` or `PLATFORM` to be explicitly set to either `nem` or `symbol`.

Handled in: ad3997c

explorer/frontend/variants/index.js
daoka: Fixed. I renamed the API to `fetchChainHeight` and updated the wrapper, NEM adapter, page imports, and related tests.

Handled in: ce23fd2

explorer/frontend/variants/index.js
daoka: Fixed. The variant registry now uses ES imports consistently instead of mixing `import` and `require`.

Handled in: 9f4db0f

explorer/frontend/variants/index.js
daoka: Updated. The variant registry now imports each variant API explicitly

Handled in: 9f4db0f

explorer/frontend/variants/symbol/api/search.js
daoka: Agreed. I’ll align Symbol search with the NEM search result.

explorer/frontend/variants/symbol/config/index.js
daoka: Agreed. I’ll use the nodeWatch service.

explorer/frontend/variants/symbol/api/namespaces.js
daoka: The plan is to keep DTO normalization inside each Symbol variant API module, so shared UI components consume the same view-model shape as NEM.

PR #2228, opened 2026-06-01
[explorer/rest] fix: add missing field mosaic

AnthonyLaw:
Problem: Account response mosaics are missing `creator` and `divisibility`.

Solution:
- Join the mosaics table when building account responses to include `creator` and `divisibility`.
- Rename account mosaic JSON field `namespace` to `namespace_name` to match the database field naming convention.


Wayonb (approved):
👍🏾

PR #2230, opened 2026-06-03
[explorer/rest] feat: add mosaics in namespace

AnthonyLaw:
Problem: missing mosaic list in namespace endpoint.
Solution: Added a subquery to retrieve mosaic info under the root namespace.


Wayonb (approved):
👍🏾

PR #2236, opened 2026-06-04
[faucet/authenicator]: migrate from Restify to Fastify

Wayonb:
problem: Faucet Authenticator build is failing since Restify does not support NodeJS 24
solution: migrate from Restify to Fastify


AnthonyLaw (approved):
👍🏼

PR #2239, opened 2026-06-06
[explorer] feat: add account height in database

AnthonyLaw:
Problem: The account table is missing account height.
Solution:
- Added `height` to the DB table.
- In the puller, it processes accounts with height.
- In the rest endpoint, the `/accounts` endpoint supports `height` in the sort field filter

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -29,7 +29,7 @@ BlockRecord = namedtuple('BlockRecord', [
 ])
 AccountRecord = namedtuple('AccountRecord', [
 	'address',
-	'registered_height',
+	'height',
 	'public_key',
 	'remote_address',
 	'importance',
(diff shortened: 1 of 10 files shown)


explorer/puller/puller/facade/NemPuller.py
```diff
@@ -539,7 +549,7 @@ async def sync_nemesis_block(self):
 		self._process_block(cursor, nemesis_block)
 
 		addresses = self._extract_addresses_from_block(nemesis_block)
-		await self._process_account_batch(cursor, addresses)
+		await self._process_account_batch(cursor, {address: nemesis_block.height for address in addresses})
```
Wayonb, line 552: Since all the addresses will have the same height, can't you just pass the height as a separate parameter?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -578,7 +588,7 @@ def _db_writer(self, block_queue, batch_size=50):
 
 			# Extract addresses for account processing
 			addresses = self._extract_addresses_from_block(block)
-			pending_addresses.update(addresses)
+			self._add_pending_addresses(pending_addresses, addresses, block.height)
```
Wayonb, line 591: same here

Wayonb (changes requested):

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: For `_process_account_batch`, I don’t think we can pass `height` as a separate parameter because the batch is not necessarily tied to one block.

In `_db_writer`, addresses are extracted from each block, but they are only processed when the batch is committed, currently every 50 blocks (batch_size). That means a single `_process_account_batch` call can include addresses first seen at different heights.

Wayonb (approved):
👍🏾

Issue #2202, opened 2026-05-25
[wallet/symbol/mobile] Address book: cannot save contact edit when name is unchanged


cryptoBeliever:
## Summary

When editing an existing address book contact, saving fails if the contact name is left unchanged. The app shows a validation error that an entry with this name already exists (validation_error_already_exists).

## Steps to reproduce

1. Add a contact to the address book (e.g. name “Alice”).
2. Open edit for that contact.
3. Keep the same name.
4. Change another field (e.g. notes) and save.

## Expected

The contact can be updated without renaming it.

## Actual

Validation blocks save because the existing name is treated as a duplicate. Workaround: change the name to something not already used in the address book or wallet accounts.

PR #2268, opened 2026-06-10
[wallet/symbol/mobile] fix: bugs

OlegMakarenko:
This pull request fixes multiple bugs of Symbol Mobile Wallet. Each bugfix is scoped to its own commit where possible, and commit messages include the related issue number as a reference when one exists. A few fixes also have follow-up commits for tests, cleanup, or comment-only corrections.

---------------------
## Problem
Related issue: #2183.
Alias unlink transactions in the history list were always labeled as link operations, so unlink actions still showed "Link ... | to ..." instead of unlink wording.

## Solution
The history description logic now checks the alias action and selects dedicated unlink localization keys when the transaction is an unlink. Address alias and mosaic alias rows now render "Unlink ... | from ..." correctly, while normal link transactions keep the existing wording.

---------------------
## Problem
Related issue: #2222.
The Send screen accepted any non-empty Ethereum recipient string as valid, which let invalid addresses proceed through fee estimation and confirmation until signing failed.

## Solution
The send form now adds a dedicated recipient validator for Ethereum transfers before the user can continue. Symbol send behavior stays unchanged because namespaces are valid recipients there, while Ethereum now requires a proper `0x` address up front.

---------------------
## Problem
Related issue: #2211.
The fixed-width social badges on the Settings About screen could clip one of the logos on narrower layouts.

## Solution
The badge layout now uses flexible width instead of a hard-coded badge width while keeping the image render mode unchanged. That lets the logos scale within the available row space without clipping.

---------------------
## Problem
Related issue: #2186.
Negative amounts below one network unit were formatted without a leading zero, producing values like `-.5` instead of `-0.5`.

## Solution
The shared absolute-to-relative amount conversion now separates the sign from the numeric magnitude before padding and slicing fractional digits. This preserves correct formatting for negative fractional values, keeps negative whole amounts intact, and normalizes negative zero back to `0`.

---------------------
## Problem
Related issue: #2203.
Several history and details views displayed raw transaction type numbers instead of readable transaction names.

## Solution
The mobile app now resolves transaction-type localization keys by chain and uses them consistently across table rows, history items, filters, aggregate descriptions, transaction graphics, and transaction details. This replaces numeric IDs with localized labels across the history UI instead of fixing only one screen.

---------------------
## Problem
Related issue: #2185.
Mosaic definition nonce values were shown in an inconsistent and confusing form, first as decimal and then still as a user-facing field that was not useful to the end user.

## Solution
The first change converted decimal mosaic noce from decimal to hex format in transaction mapper utils. The follow-up reverting back the changes and removing the nonce field from transaction details entirely, since it is not important in the transaction details (as we discussed with @cryptoBeliever).

---------------------
## Problem
Related issue: #2202.
Editing an address-book contact failed when the name stayed unchanged because duplicate-name validation treated the current contact as a conflicting entry.

## Solution
The address-book validators now accept the edited contact id and ignore that record while checking for duplicate names or addresses. Existing names can be preserved during edits, but real collisions with other contacts or wallet accounts are still blocked.

---------------------
## Problem
Related issue: #2184.
Mosaic supply decrease transactions showed "Increase by" in transaction details, and the displayed delta did not reflect the decrease direction.

## Solution
The history detail and graphic helpers now derive a signed supply delta and use the correct supply-change action metadata when rendering the transaction. Decrease transactions now show the right wording and a negative delta, and a separate follow-up commit added focused TransactionDetails coverage for supply change cases.

---------------------
## Problem
Related issue: none.
The Settings About screen pointed to an outdated Discord invite URL.

## Solution
The mobile config was updated to the current Discord invite URL.

---------------------
## Problem
Related issue: #2182.
Encrypted transfer messages could be sent from the wallet, but users could not read those messages from transaction history.

## Solution
The Symbol transfer module now correctly distinguishes encrypted and plain messages, and TransactionDetails now runs a dedicated decryption hook for transfer transactions that carry encrypted payloads. When decryption succeeds the decrypted text is injected into the displayed transaction data; when it fails the screen keeps the encrypted placeholder without surfacing an extra error popup. A separate follow-up commit added test coverage for the decryption hook.

---------------------
## Problem
Related issue: #2177.
The mnemonic import flow contained legacy opt-in account handling.

## Solution
The first change fixed the async import flow by passing the detected opt-in account explicitly into wallet creation, so a matched account could be added as an external account during import.

After the product decision to stop supporting legacy opt-in accounts in the new wallet, the obsolete secp256k1 derivation path, bundled public-key whitelist, extra loading step, and related copy were removed entirely. The import flow no longer advertises unsupported behavior and no longer carries dead opt-in-specific logic.

---------------------
## Problem
Related issue: #2219.
Multisig aggregate bonded announces could succeed on the partial endpoint and still trigger a false "required cosignature is missing" error afterward.

## Solution
`TransactionService` now returns immediately after the multisig partial-announcement path instead of falling through to the default announce flow. That prevents the same bundle from being re-announced through the full-validation endpoint, which was the source of the misleading cosignature error.

---------------------
## Problem
Related issue: #2187.
Long secret lock and secret proof hashes overflowed the transaction graphic and covered adjacent icons and the direction arrow.

## Solution
Secret captions are now truncated in the middle before rendering, and the text component used in the arrow supports single-line middle ellipsis. This keeps both the beginning and end of the hash visible while preserving the surrounding layout.

---------------------
## Problem
Related issue: #2213.
After wallet restore PIN entry, the software keyboard could briefly reappear on top of the full-screen wallet creation animation.

## Solution
The onboarding flow now dismisses the keyboard before showing the passcode and loading transition, using a platform utility that explicitly closes the focused input. A focused screen test was added so the loading animation stays keyboard-free.

---------------------
## Problem
Related issue: #2179.
Opening token details from the Bridge Account Details screen could crash because the navigation payload missed a required route parameter.

## Solution
`BridgeAccountDetails` now passes `accountAddress` together with `chainName`, `tokenId`, and the preloaded token data when routing to token details.

---------------------
## Problem
Related issue: #2291.
Ethereum websocket listeners were not torn down cleanly, so repeated connection jobs could accumulate open listeners and exhaust reverse-proxy websocket capacity.

## Solution
Listener shutdown now clears provider references, removes listeners, and destroys the underlying websocket provider, while `NetworkManager` waits for the previous chain listener to close before creating a replacement. This prevents overlapping listeners during reconnect cycles and closes stale websocket transports deterministically.

---------------------
## Problem
Related issue: #2299.
Mainnet fiat display could show `~NaN USD` because the wallet's market price integration no longer matched the response shape of the provider it was using.

## Solution
The Symbol market service was switched from the old CryptoCompare-style query to CoinGecko's `simple/price` endpoint, using the `symbol` coin id and lowercased fiat codes. The wallet now maps the nested response correctly and suppresses fiat text when the returned price is missing or non-finite, which prevents `NaN` output in the UI.

---------------------
## Problem
Related issue: none.
Android builds failed due to bumped dependencies by Dependabot .

## Solution
Reverted back the `react-native-permissions` and `react-native-vision-camera` to compatible versions.

---------------------
## Problem
Related issue: #2221.
Ethereum ERC-20 sends could estimate gas successfully but still revert on-chain because raw `eth_estimateGas` results were sometimes too low for proxy-based or internally delegated contract calls.

## Solution
The Ethereum transaction service now adds a 20% safety margin to estimated gas limits before returning them to the send flow. This only raises the execution cap, not the cost of unused gas, and it materially reduces out-of-gas reverts for ERC-20 transfers. A small follow-up commit only corrected the related test comment.


wallet/common/core/src/lib/controller/WalletController.js
```diff
@@ -311,7 +311,7 @@ export class WalletController {
 	 * @returns {boolean} - True if network connection is ready, false otherwise.
 	 */
 	get isNetworkConnectionReady() {
-		return this.networkStatus === NetworkConnectionStatus.CONNECTED;
+		return this.networkStatus === NetworkConnectionStatus.CONNECTED && !!this._state.networkProperties?.nodeUrl;
```
Wayonb: Why not set the ``networkStatus`` to connected after all the _state is set? 🤔

wallet/symbol/mobile/src/utils/network.js
```diff
@@ -8,7 +8,17 @@ import { InternalServerError, InvalidRequestError, NetworkRequestError, NotFound
  * @returns {Promise} The request response.
  */
 export const makeRequest = async (url, options) => {
-	const response = await fetch(url, options);
+	if (!url || url.startsWith('undefined') || url.startsWith('null'))
```
Wayonb: Why not check if the URL is valid?  meaning in the https://hostname:port
I can set the URL to an empty string, which will fail below?

Wayonb (changes requested):

wallet/symbol/mobile/src/utils/network.js
OlegMakarenko: I agree - added a helper for wider validation.
 
 Yes, the empty string URL will throw on this check.

wallet/common/core/src/lib/controller/WalletController.js
OlegMakarenko: Thanks, removed. The order is already fixed.

Wayonb (approved):
👍🏾

PR #2272, opened 2026-06-12
[explorer/backend] nem model subdirectories

daoka:

NEM REST routes are registered directly in `explorer/rest/rest/__init__.py`, and NEM REST model classes are placed directly under `explorer/rest/rest/model`.

The NEM account remarks resource is named `mainnet_account_remarks.json`, which does not make the chain ownership explicit.

## What's the issue?

Symbol Puller / REST API work will add Symbol-specific routes and models later. If the current structure is kept, NEM and Symbol route/model code would be mixed in shared top-level files and directories, making review boundaries and future maintenance harder.

The REST app also did not require an explicit chain selection, even though the intended runtime model is to run either NEM or Symbol, not both at the same time.

This PR prepares the backend structure for chain-specific REST implementations while preserving the existing NEM behavior.

Changes include:

- Renamed the NEM account remarks resource:
  - `mainnet_account_remarks.json`
  - `nem_mainnet_account_remarks.json`
- Moved NEM REST models under a chain-specific package:
  - `rest/model/nem/Account.py`
  - `rest/model/nem/Block.py`
  - `rest/model/nem/Mosaic.py`
  - `rest/model/nem/Namespace.py`
  - `rest/model/nem/Statistic.py`
  - `rest/model/nem/Transaction.py`
- Moved NEM route registration out of `rest/__init__.py` into:
  - `rest/routes/nem.py`
- Added explicit REST chain selection through `REST_CHAIN`.
  - Currently supported value: `nem`
  - Missing `REST_CHAIN` fails startup.
  - Unsupported chains fail startup.
- Updated tests and imports to use the new NEM-specific model paths.

This PR does not add Symbol REST routes or Symbol database logic.


explorer/rest/tests/test_app.py
```diff
@@ -0,0 +1,83 @@
...
+from flask import Flask, abort, jsonify
+
+import rest
+from rest import create_app, load_rest_config, setup_error_handlers
+
+
+def _write_rest_config(monkeypatch, tmp_path, contents):
+	config_path = tmp_path / 'app.config'
+	config_path.write_text(contents, encoding='utf8')
+	monkeypatch.setenv('EXPLORER_REST_SETTINGS', str(config_path))
```
Jaguar0625: is there reason you're using monkeypatch vs setting env directly using `os.environ['EXPLORER_REST_SETTINGS'] = str(config_path)` ?

daoka: I used monkeypatch.setenv here so pytest restores EXPLORER_REST_SETTINGS after each test. These app factory tests set different temporary config files, so using monkeypatch avoids leaking environment state between tests while still exercising Flask's from_envvar path. I can switch to os.environ for consistency with others if preferred.

Wayonb (approved):
👍🏾

Jaguar0625 (approved):
Great job! 👍

PR #2273, opened 2026-06-13
[explorer]: add seed for namespace and mosaic

AnthonyLaw:
Problem:
- Currently, `balance` and `vested_balance` are stored in the DB as relative values. After some thought, I decided that all xem amounts should be stored as absolute values for consistency.
- The network currency info is missing: `nem` namespace and `nem.xem` mosaic are not in the db table.

Solution:
- Converted `balance` and `vested_balance` to absolute values to keep them consistent.
- Added a seed for network currency info into the DB.


explorer/puller/puller/db/NemDatabase.py
```diff
@@ -269,6 +269,65 @@ def create_tables(self):
...
+			)
+			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+			ON CONFLICT (namespace_name)
+			DO NOTHING
+			''',
+			(
+				'nem',
+				'nem.xem',
+				'network currency',
+				bytes(32),  # zero public key
```
Wayonb: What about the creator?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -252,8 +252,8 @@ def _create_account_record(account_info, mosaics_json, height, remote_address=No
 			account_info.public_key,
 			remote_address,
 			account_info.importance,
-			account_info.balance,
-			account_info.vested_balance,
+			account_info.balance * (10 ** 6),  # convert from relative to absolute
```
Wayonb: Can you create a helper to do the conversion?

Wayonb (changes requested):

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I just gave it to empty bytes, which is a zero public key.
`'0' * 64`

Do you have anything in mind?

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: I added a helper in this commit: [767a2c7](https://github.com/symbol/product/pull/2273/commits/767a2c7da3eec48f54b1891033fd953686cc174f)

and I rebased with `dev`

explorer/puller/puller/db/NemDatabase.py
Wayonb: Ok  and I assume it gets updated?  Just wondering since I see this.
https://testnet.nem.fyi/mosaics/nem.xem

Wayonb (changes requested):

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I found a better way to handle this.
The creator should be the signer from the nemesis block

Wayonb (approved):
👍🏾

PR #2293, opened 2026-06-17
[explorer] fix: convert owner public key to address

AnthonyLaw:
Problem: The latest REST importance default is 0, the owner was displayed in the public key.
Solution: Update importance to 0 instead of null, and convert the public key to an address


OlegMakarenko (approved):

Wayonb (approved):
👍🏾

PR #2297, opened 2026-06-17
[explorer/backend] feat: Symbol backend core foundation

daoka:

Explorer backend REST/Puller currently has reviewed NEM implementations, but Symbol backend does not yet have a minimal core foundation for Puller, DB, REST facade, route registration, or Symbol node access validation.

`REST_CHAIN` can select chain-specific REST handlers after the NEM layout cleanup, but there is no Symbol backend route/facade implementation to register.

## What's the issue?

Upcoming Symbol Puller / Explorer REST API work needs a small backend foundation before adding real data vertical slices such as blocks, transactions, accounts, receipts, and statics.

Without this foundation, later PRs would need to introduce route registration, DB wiring, node security validation, facade boundaries, and real feature logic at the same time, making reviews larger and riskier.

Added the minimal Symbol backend core foundation:

- Added shared Symbol node access validation under `explorer/common/symbol`.
- Added Symbol Puller skeleton:
  - `SymbolPuller`
  - `SymbolDatabase`
  - Symbol-specific node config wrapper
  - validated `get_symbol_node()` / `post_symbol_node()` connector entrypoints
- Added Symbol REST skeleton:
  - `SymbolRestFacade`
  - `SymbolDatabase`
  - `setup_symbol_facade(app)`
  - `setup_symbol_routes(app, symbol_api_facade)`
  - `/api/symbol/health`
- Registered Symbol in `REST_CHAIN_HANDLERS`.
- Kept NEM and Symbol route registration separated.
- Added Symbol-specific model directories for future REST model expansion.
- Updated REST/Puller CI test coverage source to include `../common`.
- Added tests for Symbol node config validation, facade health behavior, route boundary behavior, DB connection checks, and NEM/Symbol REST chain selection.

This PR intentionally does not add Symbol block sync, block tables, transactions, accounts, receipts, statics, frontend changes, or real Symbol explorer data APIs.


explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,184 @@
...
+
+class SymbolNodeConfigError(ValueError):
+	"""Raised when Symbol node access config is invalid."""
+
+
+def _parse_bool(value):
+	if isinstance(value, bool):
+		return value
+
+	if str(value).lower() == 'true':
```
Wayonb: You should convert the value first so it not needed in each if statement.
```
normalized_value = str(value).lower()
if 'true' == normalized_value:
```

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,184 @@
...
+	except (TypeError, ValueError) as error:
+		raise SymbolNodeConfigError(f'{name} must be a positive integer') from error
+
+	if parsed_value < 1:
+		raise SymbolNodeConfigError(f'{name} must be a positive integer')
+
+	return parsed_value
+
+
+def _normalize_base_url(node_url):
```
Wayonb, line 39: Is this specific to Symbol node or can be use for NEM also?

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,184 @@
...
+	parsed_url = urlparse(node_url)
+
+	if parsed_url.scheme not in ('http', 'https'):
+		raise SymbolNodeConfigError('Symbol node URL scheme must be http or https')
+	if not parsed_url.hostname:
+		raise SymbolNodeConfigError('Symbol node URL must include a host')
+	if parsed_url.username or parsed_url.password:
```
Wayonb, line 46: Why is this needed?

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,184 @@
...
+		raise SymbolNodeConfigError('Symbol node URL must include a host')
+	if parsed_url.username or parsed_url.password:
+		raise SymbolNodeConfigError('Symbol node URL must not include userinfo')
+	if parsed_url.query or parsed_url.fragment:
+		raise SymbolNodeConfigError('Symbol node URL must not include query or fragment')
+	if parsed_url.path not in ('', '/'):
+		raise SymbolNodeConfigError('Symbol node URL must not include a path prefix')
+
+	try:
+		port = parsed_url.port or (443 if 'https' == parsed_url.scheme else 80)
```
Wayonb: Should it fail if the port is missing? 🤔

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,184 @@
...
+def _normalize_allowed_hosts(raw_allowed_hosts):
+	if not raw_allowed_hosts:
+		raise SymbolNodeConfigError('SYMBOL_NODE_ALLOWED_HOSTS is required')
+
+	allowed_hosts = set()
+	for entry in str(raw_allowed_hosts).split(','):
+		entry = entry.strip().lower()
+		if not entry:
+			continue
+		if '://' in entry or '/' in entry or '?' in entry or '#' in entry or '*' in entry:
```
Wayonb, line 73: Is this to check if entry is a url? or else fail?

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,184 @@
...
+			'SYMBOL_NODE_ALLOW_PRIVATE': kwargs.get('allow_private', False),
+			'SYMBOL_NODE_ALLOW_LOOPBACK': kwargs.get('allow_loopback', False),
+			'SYMBOL_NODE_REQUEST_TIMEOUT_SECONDS': kwargs.get('timeout_seconds', 10)
+		})
+
+	@classmethod
+	def _from_values(cls, values):
+		node_url = values.get('SYMBOL_NODE_URL')
+		scheme, host, port, base_url = _normalize_base_url(node_url)
+		allowed_hosts = _normalize_allowed_hosts(values.get('SYMBOL_NODE_ALLOWED_HOSTS'))
```
Wayonb, line 153: Why is the allowed node host list needed?

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,184 @@
...
+	@classmethod
+	def from_url(cls, node_url, **kwargs):
+		"""Creates a node config from an explicit node URL."""
+
+		_, host, port, _ = _normalize_base_url(node_url)
+		return cls._from_values({
+			'SYMBOL_NODE_URL': node_url,
+			'SYMBOL_NODE_ALLOWED_HOSTS': f'{host}:{port}',
+			'SYMBOL_NODE_ALLOW_PRIVATE': kwargs.get('allow_private', False),
+			'SYMBOL_NODE_ALLOW_LOOPBACK': kwargs.get('allow_loopback', False),
```
Wayonb: Also, why allow private or loopback?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -0,0 +1,33 @@
...
+
+	def create_tables(self):
+		"""Creates Symbol core foundation tables."""
+
+		cursor = self.connection.cursor()
+		cursor.execute(
+			'''
+			CREATE TABLE IF NOT EXISTS symbol_sync_state (
+				id int PRIMARY KEY DEFAULT 1,
+				status varchar(32) NOT NULL DEFAULT 'initialized',
```
Wayonb: Why set the default to ``initialized``

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -0,0 +1,72 @@
...
+		db_config = config['symbol_db']
+
+		network = Network.MAINNET if network_type == 'mainnet' else Network.TESTNET
+
+		self.symbol_db = SymbolDatabase(DatabaseConfig(**db_config))
+		self.node_config = node_config or SymbolNodeConfig.from_url(node_url)
+		symbol_node_endpoint = self.node_config.assert_request_allowed(self.node_config.base_url)
+		self._symbol_connector = SymbolConnector(symbol_node_endpoint)
+		self._symbol_connector.timeout_seconds = self.node_config.timeout_seconds
+		self.symbol_facade = SymbolFacade(str(network))
```
Wayonb: should not need to convert the network to a string.

explorer/puller/puller/model/symbol/NodeConfig.py
```diff
@@ -0,0 +1,11 @@
+import importlib.util
+from pathlib import Path
+
+_NODE_CONFIG_PATH = Path(__file__).resolve().parents[4] / 'common' / 'symbol' / 'NodeConfig.py'
+_NODE_CONFIG_SPEC = importlib.util.spec_from_file_location('common_symbol_node_config', _NODE_CONFIG_PATH)
+_node_config = importlib.util.module_from_spec(_NODE_CONFIG_SPEC)
+_NODE_CONFIG_SPEC.loader.exec_module(_node_config)
```
Wayonb: Why do you need to do a dynamic import? 🤔

explorer/puller/scripts/ci/lint.sh
```diff
@@ -6,14 +6,14 @@ if [ $# -gt 0 ]; then
 	PYLINT_DISABLE_COMMANDS="$1"
 fi
 
-find . -type f -name "*.sh" -print0 | xargs -0 shellcheck
-find . -type f -name "*.py" -print0 | PYTHONPATH=. xargs -0 python3 -m isort \
+find . ../common -type f -name "*.sh" -print0 | xargs -0 shellcheck
```
Wayonb, line 9: Ideally, ``common`` folder should be another project with own set of scripts.

Wayonb (changes requested):

explorer/common/symbol/NodeConfig.py
daoka: Updated. _parse_bool() now normalizes the value once before comparing it with true or false.

explorer/common/symbol/NodeConfiguration.py
daoka: Currently, it is for Symbol, but it could be expanded to NEM if necessary.

explorer/common/symbol/NodeConfig.py
daoka: Yes. I changed it to fail when the port is omitted. Symbol node endpoints are configured with an explicit API port, so silently defaulting to 80 or 443 could point to a different target than intended. Tests were updated for both REST and Puller.

explorer/common/symbol/NodeConfig.py
daoka: I removed the allowed-hosts parser entirely. The extra entry-shape validation is no longer needed because request validation now compares the request target directly with the normalized SYMBOL_NODE_URL origin.

explorer/common/symbol/NodeConfig.py
daoka: Agreed that the separate allowed host list was unnecessary. I removed SYMBOL_NODE_ALLOWED_HOSTS and the allowed_hosts state, and kept the security check as an exact match against the configured Symbol node origin.

explorer/common/symbol/NodeConfig.py
daoka: Private and loopback targets are denied by default. The flags are explicit opt-ins for local development, tests, and private-network deployments. The tests cover both the default rejection and the explicit allow cases.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Removed the database default. The sync status should be written explicitly by the sync workflow instead of being hidden in the schema. I also added a test assertion that the default is not generated.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Updated. SymbolPuller now passes the Network object directly to SymbolFacade instead of converting it to a string.

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -0,0 +1,33 @@
...
+
+	def create_tables(self):
+		"""Creates Symbol core foundation tables."""
+
+		cursor = self.connection.cursor()
+		cursor.execute(
+			'''
+			CREATE TABLE IF NOT EXISTS symbol_sync_state (
+				id int PRIMARY KEY DEFAULT 1,
+				status varchar(32) NOT NULL,
```
Jaguar0625: why varchar? couldn't this be some sort of enum?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -0,0 +1,35 @@
+from unittest import TestCase
+from unittest.mock import Mock
+
+from puller.db.SymbolDatabase import SymbolDatabase
+
+
+class TestSymbolDatabase(TestCase):
```
Jaguar0625: 1. please add Arrange/Act/Assert comments to tests
2. class naming is typically SymbolDatabaseTest 
3. for database wrappers like class under test, we typically tend towards more integration tests than unit tests because unit tests require excessive mocking and that typically will not detect db interfacing issues (e.g. wrong call ordering). for example see https://github.com/symbol/product/tree/dev/bridge/tests/db

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -0,0 +1,35 @@
...
+	def test_create_tables_creates_symbol_sync_state_only(self):
+		database = SymbolDatabase.__new__(SymbolDatabase)
```
Jaguar0625: why are you using this construct vs calling constructor ?

explorer/puller/puller/model/symbol/NodeConfig.py
```diff
@@ -0,0 +1 @@
+from common.symbol.NodeConfig import SymbolNodeConfig, SymbolNodeConfigError, socket
```
Jaguar0625: is this just a placeholder that you intend to extend?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -0,0 +1,72 @@
...
+from urllib.parse import urlparse
+
+from symbolchain.facade.SymbolFacade import SymbolFacade
+from symbolchain.symbol.Network import Network
+from symbollightapi.connector.SymbolConnector import SymbolConnector
+
+from puller.db.SymbolDatabase import SymbolDatabase
+from puller.model.symbol.NodeConfig import SymbolNodeConfig
+
+DatabaseConfig = namedtuple('DatabaseConfig', ['database', 'user', 'password', 'host', 'port'])
```
Jaguar0625: for type names, we typically prefer full words; Config => Configuration
abbreviations (config) are ok for variable names

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -0,0 +1,72 @@
...
+	"""Facade for pulling data from Symbol network."""
+
+	def __init__(self, node_url, config_file, network_type='mainnet', node_config=None):
+		"""Creates a Symbol puller facade object."""
+
+		config = configparser.ConfigParser()
+		config.read(config_file)
+
+		if 'symbol_db' not in config:
+			raise KeyError('symbol_db')
```
Jaguar0625: seems redundant since map indexer should throw same exception

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -0,0 +1,72 @@
...
+
+	async def post_symbol_node(self, url_path, request_payload, property_name=None, not_found_as_error=True):
+		"""Validates and dispatches a Symbol node POST request."""
+
+		return await self._symbol_connector.post(
+			self._validate_symbol_node_path(url_path),
+			request_payload,
+			property_name,
+			not_found_as_error
+		)
```
Jaguar0625, line 73: how do you intend to call these? from discovered node urls or with semi-trusted urls?

explorer/puller/tests/facade/test_SymbolPuller.py
```diff
@@ -0,0 +1,123 @@
...
+
+	return SymbolPuller(
+		node_url,
+		db_config_path,
+		network_type,
+		node_config
+	)
+
+
+class TestSymbolPuller(TestCase):
```
Jaguar0625: SymbolPullerTest

explorer/rest/tests/db/test_SymbolDatabase.py
```diff
@@ -0,0 +1,20 @@
+from contextlib import nullcontext
+from unittest import TestCase
+from unittest.mock import Mock
+
+from rest.db.SymbolDatabase import SymbolDatabase
+
+
+class TestSymbolDatabase(TestCase):
```
Jaguar0625: similar comments as earlier database tests

explorer/rest/rest/facade/SymbolRestFacade.py
```diff
@@ -0,0 +1,96 @@
...
+				log.error(f'Failed to initialize Symbol database: {error}')
+				self.db_error = DATABASE_UNAVAILABLE_MESSAGE
+
+		self.node_config = node_config
+		self.config_error = str(config_error) if config_error else None
+
+	def is_configured(self):
+		"""Returns whether Symbol REST dependencies are configured."""
+
+		return self.symbol_db is not None and self.node_config is not None and self.config_error is None
```
Jaguar0625: why would you want to allow creation of facade without database? shouldn't that be an exceptional failure?

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -0,0 +1,137 @@
...
+
+from psycopg2 import OperationalError
+
+from rest.facade.SymbolRestFacade import SymbolRestFacade
+from rest.model.symbol.NodeConfig import SymbolNodeConfig, SymbolNodeConfigError
+
+from ..test.PostgresTestUtils import create_unreachable_db_config
+from ..test.SymbolHealthTestUtils import create_symbol_health
+
+_NODE_URL = 'http://127.0.0.1:3000'
```
Jaguar0625: for private constants and classes, wouldn't add leading underscore. we only tend to do that for functions and private class variables.

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -0,0 +1,137 @@
...
+		self.assertEqual(create_symbol_health(
+			nodeConfigured=True,
+			errors=[{
+				'type': 'database',
+				'message': 'Symbol database is unavailable'
+			}]
+		), facade.get_health())
+		self.assertNotIn('connection refused', str(facade.get_health()))
+
+	def test_reports_configuration_error(self):
```
Jaguar0625: as mentioned before, i think configuration errors should probably prevent boot vs be reported

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -0,0 +1,137 @@
...
+		self.assertEqual(create_symbol_health(
+			dbUp=True,
+			nodeConfigured=True,
+			errors=[{
+				'type': 'configuration',
+				'message': 'bad config'
+			}]
+		), result)
+
+	def test_validates_node_request_target(self):
```
Jaguar0625: typically for test naming it's best to have something akin to a full sentence

validate_node_request_target_returns_url_when_node_is_valid
validate_node_request_target => function under test 
returns_url => action
when_node_is_valid => condition

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -0,0 +1,137 @@
...
+
+		result = facade.validate_node_request_target(_NODE_URL)
+
+		self.assertEqual(_NODE_URL, result)
+
+	def test_rejects_node_request_validation_when_node_config_is_missing(self):
+		facade = SymbolRestFacade()
+
+		with self.assertRaisesRegex(SymbolNodeConfigError, 'Symbol node URL is not configured'):
+			facade.validate_node_request_target('http://localhost:3000')
```
Jaguar0625: i am not sure why you need to expose this separately from facade since it is just delegating to node_config function?

explorer/rest/rest/routes/symbol.py
```diff
@@ -0,0 +1,47 @@
...
+			symbol_db_config['host'],
+			symbol_db_config['port']
+		)
+
+	try:
+		node_config = SymbolNodeConfig.from_app_config(app.config)
+		config_error = None
+	except SymbolNodeConfigError as error:
+		node_config = None
+		config_error = error
```
Jaguar0625: as mentioned elsewhere, it feels more natural for these to be fatal errors

explorer/rest/tests/test/SymbolHealthTestUtils.py
```diff
@@ -0,0 +1,12 @@
+def create_symbol_health(**overrides):
+	health = {
```
Jaguar0625, line 4: add function documentation

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -0,0 +1,252 @@
...
+		if include_symbol_db:
+			database_config = database_config or DatabaseConfig('symbol', 'postgres', '', '127.0.0.1', '5433')
+			db_config_file.write('[symbol_db]\n')
+			db_config_file.write(f'database = {database_config.database}\n')
+			db_config_file.write(f'user = {database_config.user}\n')
+			db_config_file.write('password = \n')
+			db_config_file.write(f'host = {database_config.host}\n')
+			db_config_file.write(f'port = {database_config.port}\n')
+
+	return db_config_path
```
Jaguar0625, line 34: consider refactoring into helpers since this or similar appears in mutiple places

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -0,0 +1,252 @@
...
+
+def test_rest_env_restores_existing():
+	with tempfile.TemporaryDirectory() as temp_directory:
+		config_path = Path(temp_directory) / 'app.config'
+
+		with _temporary_env_values({'EXPLORER_REST_SETTINGS': 'previous.config'}):
+			with _rest_settings_env(config_path):
+				assert str(config_path) == os.environ['EXPLORER_REST_SETTINGS']
+
+			assert 'previous.config' == os.environ['EXPLORER_REST_SETTINGS']
```
Jaguar0625: don't think you need to test test functions

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -0,0 +1,252 @@
...
+
+
+def test_external_postgres_config():
```
Jaguar0625: this also seems like testing of test code?

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -0,0 +1,252 @@
...
+		'EXPLORER_TEST_POSTGRES_USER': 'symbol_user',
+		'EXPLORER_TEST_POSTGRES_PORT': '15432'
+	}
+
+	with _temporary_env_values(postgres_env):
+		with PostgresTestDatabase() as db_config:
+			assert DatabaseConfig('symbol_test', 'symbol_user', '', 'postgres.example', '15432') == db_config
+
+
+def test_testing_postgresql_fallback():
```
Jaguar0625: what are you attempting to test here ?

explorer/rest/tests/test_app.py
```diff
@@ -1,29 +1,53 @@
...
+			if previous_value is None:
+				os.environ.pop(name, None)
+			else:
+				os.environ[name] = previous_value
+
+
+@contextmanager
+def _rest_settings_env(config_path):
+	with _temporary_env_values({'EXPLORER_REST_SETTINGS': str(config_path)}):
+		yield
```
Jaguar0625: consider refactoring these since you're using them in multiple files

explorer/rest/tests/test_app.py
```diff
@@ -90,20 +114,49 @@ def test_loads_envvar_config(rest_config_path):
...
+	# Arrange:
+	config_path = tmp_path / 'app.config'
+
+	# Act:
+	with _temporary_env_values({'EXPLORER_REST_SETTINGS': 'previous.config'}):
+		with _rest_settings_env(config_path):
+			assert str(config_path) == os.environ['EXPLORER_REST_SETTINGS']
+
+		# Assert:
+		assert 'previous.config' == os.environ['EXPLORER_REST_SETTINGS']
```
Jaguar0625: not sure if you need to test test functions

explorer/rest/tests/test_app.py
```diff
@@ -112,8 +165,77 @@ def bad_request():
...
+			}
+
+	app = Flask(__name__)
+	setup_symbol_routes(app, SymbolHealthFacade())
+
+	response = app.test_client().get('/api/symbol/health')
+
+	assert 200 == response.status_code
+	assert response.json['isHealthy']
+	assert [] == response.json['errors']
```
Jaguar0625: since you're using mock data, would suggest single assert testing full json

explorer/rest/tests/model/test_SymbolNodeConfig.py
```diff
@@ -0,0 +1,200 @@
+import pytest
```
Jaguar0625: isn't this same as common.symbol.NodeConfig?

in which case tests seem redundant to tests already there?

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,153 @@
+import ipaddress
+import socket
+from urllib.parse import urlparse
+
+
+class SymbolNodeConfigError(ValueError):
```
Jaguar0625: for class names, use whole word: Config => Configuration

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,153 @@
...
+	}
+
+
+def _is_forbidden_address(address):
+	return any([
+		address.is_link_local,
+		address.is_multicast,
+		address.is_unspecified,
+		str(address) == '169.254.169.254'
+	])
```
Jaguar0625, line 101: can you explain purposes of these functions and hardcoded urls?

explorer/common/symbol/NodeConfig.py
Jaguar0625: is this intended just for main node explorer is communicating with or arbitrary nodes?

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,153 @@
...
+			base_url=base_url,
+			allow_private=_parse_bool(values.get('SYMBOL_NODE_ALLOW_PRIVATE', 'false')),
+			allow_loopback=_parse_bool(values.get('SYMBOL_NODE_ALLOW_LOOPBACK', 'false')),
+			timeout_seconds=_parse_positive_int(
+				'SYMBOL_NODE_REQUEST_TIMEOUT_SECONDS',
+				values.get('SYMBOL_NODE_REQUEST_TIMEOUT_SECONDS', 10)
+			)
+		)
+
+	def assert_request_allowed(self, request_url):
```
Jaguar0625, line 175: 1. document
2. how do you expect this being called? why not make it private and call it during object creation?

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,153 @@
...
+			if _is_forbidden_address(address):
+				raise SymbolNodeConfigError('Resolved Symbol node address is not allowed')
+			if address.is_loopback and not self.allow_loopback:
+				raise SymbolNodeConfigError('Loopback Symbol node address requires SYMBOL_NODE_ALLOW_LOOPBACK=true')
+			if address.is_private and not address.is_loopback and not self.allow_private:
+				raise SymbolNodeConfigError('Private Symbol node address requires SYMBOL_NODE_ALLOW_PRIVATE=true')
+
+		return base_url
+
+	def to_dict(self):
```
Jaguar0625, line 193: please document  all public functions

explorer/puller/tests/model/test_SymbolNodeConfig.py
```diff
@@ -0,0 +1,209 @@
+import pytest
```
Jaguar0625: name of test file should match name of file under test; i.e. test_NodeConfig.py

explorer/puller/tests/model/test_SymbolNodeConfig.py
```diff
@@ -0,0 +1,209 @@
...
+
+def test_app_metadata_host_rejected():
+	with pytest.raises(SymbolNodeConfigError, match='Metadata service Symbol node host is not allowed'):
+		SymbolNodeConfig.from_app_config(_base_app_config(SYMBOL_NODE_URL='http://169.254.169.254:3000'))
+
+
+def test_accepts_boolean_config_values():
+	node_config = SymbolNodeConfig.from_app_config(_base_app_config(
+		SYMBOL_NODE_ALLOW_PRIVATE='true',
+		SYMBOL_NODE_ALLOW_LOOPBACK='true'
```
Jaguar0625: consider having one 'false', so you're testing parsing of both boolean strings

explorer/puller/tests/model/test_SymbolNodeConfig.py
```diff
@@ -0,0 +1,209 @@
...
+
+
+def test_rejects_different_target():
+	node_config = SymbolNodeConfig.from_url('http://localhost:3000', allow_loopback=True)
+
+	with pytest.raises(SymbolNodeConfigError, match='Symbol node request target does not match configured base URL'):
+		node_config.assert_request_allowed('http://localhost:3001')
+
+
+def test_direct_config_allows_matching_target():
```
Jaguar0625: what is difference between this test and test_allows_matching_request_target ?

explorer/puller/tests/model/test_SymbolNodeConfig.py
```diff
@@ -0,0 +1,209 @@
...
+	node_config = SymbolNodeConfig.from_url('http://10.0.0.5:3000', allow_private=True)
+
+	assert 'http://10.0.0.5:3000' == node_config.assert_request_allowed('http://10.0.0.5:3000')
+
+
+def test_rejects_forbidden_address():
+	node_config = SymbolNodeConfig.from_url('http://169.254.1.1:3000', allow_private=True)
+
+	with pytest.raises(SymbolNodeConfigError, match='Resolved Symbol node address is not allowed'):
+		node_config.assert_request_allowed('http://169.254.1.1:3000')
```
Jaguar0625: why is this rejected?

Jaguar0625 (commented):
👍

explorer/common/symbol/NodeConfiguration.py
daoka: This is to keep Symbol node access constrained to the configured node origin. The next work extends this with SYMBOL_NODE_ALLOWED_HOSTS, so requests cannot drift to arbitrary or discovered targets.

explorer/common/symbol/NodeConfig.py
daoka: It is intended for the configured Symbol node that Explorer communicates with.

explorer/puller/puller/model/symbol/NodeConfig.py
daoka: The wrapper was originally intended as a temporary re-export so REST and Puller could share the same Symbol node validation code. I removed the REST/Puller wrapper modules and now import `common.symbol.NodeConfig` directly.

explorer/puller/scripts/ci/lint.sh
daoka: For this PR I kept `common` within the current repository layout and made REST/Puller lint and tests resolve it consistently. Splitting `common` into its own project with dedicated scripts should be handled as a follow-up PR.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Yes, enum is better here since this is pre-launch schema work. I changed `symbol_sync_state.status` from `varchar` to a PostgreSQL enum.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Updated. I renamed the test class to `SymbolDatabaseTest`, added Arrange/Act/Assert comments, and moved the Symbol DB tests from SQL-string mocks to PostgreSQL-backed integration tests.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Originally this used `__new__` to inspect emitted SQL without opening a database connection. I changed it to call the real constructor and verify observable database behavior instead.

explorer/puller/puller/model/symbol/NodeConfig.py
daoka: It was only a temporary re-export placeholder. I removed it and switched callers to import the common Symbol node configuration directly.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Updated. The shared type is now `SymbolNodeConfiguration` and the error type is `SymbolNodeConfigurationError`. I kept `config` only as a variable abbreviation.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Updated. I removed the explicit `symbol_db` pre-check and now let `config['symbol_db']` raise the natural `KeyError`.

explorer/puller/puller/facade/SymbolPuller.py
daoka: These methods are intended only for the configured and allowlisted Symbol node. They dispatch relative paths through the configured connector, not discovered or semi-trusted absolute URLs. I added path validation and tests for absolute URL and fragment rejection.

explorer/puller/tests/facade/test_SymbolPuller.py
daoka: Updated. I renamed the test class to `SymbolPullerTest` to match the existing repository style.

explorer/rest/tests/db/test_SymbolDatabase.py
daoka: Updated the REST Symbol DB test in the same direction: renamed the class to `SymbolDatabaseTest`, added Arrange/Act/Assert comments.

explorer/rest/rest/facade/SymbolRestFacade.py
daoka: Originally I allowed the facade to exist so health could report an unconfigured state. I changed this to fail at setup time when the Symbol database configuration is missing, while keeping runtime database availability in the health response.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: Updated. I removed the leading underscores from the test constants/classes and kept underscores only for helper functions.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: I changed invalid or missing Symbol node/database configuration to fail during setup. Health now focuses on runtime availability, such as database connectivity failures.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: The tests around that facade-level delegation were removed because the delegation API itself was redundant. The remaining tests were renamed around observable setup and health behavior.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: Agreed. It was only delegating to `node_config`, so I removed the facade-level method and kept target validation on `SymbolNodeConfiguration`.

explorer/rest/rest/routes/symbol.py
daoka: Updated. `setup_symbol_facade()` now fails immediately for missing `symbol_db` or missing/invalid Symbol node configuration instead of creating a partially configured facade.

explorer/rest/tests/test/SymbolHealthTestUtils.py
daoka: Updated. I added a docstring to `create_symbol_health()`.

explorer/rest/tests/test_symbol_core_routes.py
daoka: I extracted the shared environment-variable helper into `EnvTestUtils.temporary_env_values`. The Symbol config-file helpers remain local because they are specific to the Symbol REST setup tests.

explorer/rest/tests/test_symbol_core_routes.py
daoka: Updated. I removed the helper-only tests and kept coverage on caller-observable behavior.

explorer/rest/tests/test_symbol_core_routes.py
daoka: Updated. The direct tests of test helpers were removed and the helper behavior is now covered through the tests that use it.

explorer/rest/tests/test_symbol_core_routes.py
daoka: That was intended to cover helper behavior. I removed it and kept the assertions focused on setup-time Symbol facade behavior.

explorer/rest/tests/test_app.py
daoka: Updated. I moved the duplicated environment helper to `tests/test/EnvTestUtils.py` and use it from both test files.

explorer/rest/tests/test_app.py
daoka: Updated. I removed those helper-only tests and rely on the tests that exercise the helper through real app configuration flows.

explorer/rest/tests/test_app.py
daoka: Updated. The Symbol health route test now asserts the full JSON response in a single assertion.

explorer/rest/tests/model/test_SymbolNodeConfig.py
daoka: I removed the REST/Puller duplicate NodeConfig tests and consolidated detailed behavior coverage under `explorer/common/tests/test_NodeConfig.py`.

explorer/common/symbol/NodeConfig.py
daoka: Updated. I renamed `SymbolNodeConfig` to `SymbolNodeConfiguration` and `SymbolNodeConfigError` to `SymbolNodeConfigurationError`.

explorer/common/symbol/NodeConfiguration.py
daoka: The hardcoded URLs are validation fixtures for the Symbol node origin policy, including metadata-service, private, loopback, and link-local cases. I centralized these detailed tests in common so REST/Puller only need integration smoke coverage.

explorer/common/symbol/NodeConfiguration.py
daoka: Originally this existed so callers could validate each future node request target. I kept it public because Puller validates dispatch targets before requests, and added a docstring. Object creation still validates the configured base URL separately.

explorer/common/symbol/NodeConfiguration.py
daoka: Updated. I added docstrings to the public `SymbolNodeConfiguration` factory, validation, and serialization methods.

explorer/puller/tests/model/test_SymbolNodeConfig.py
daoka: Updated. The duplicate Puller test file was removed, and the consolidated common test file is named `test_NodeConfig.py`.

explorer/puller/tests/model/test_SymbolNodeConfig.py
daoka: Updated. The consolidated boolean parsing test now covers a true value and a false value in the same behavior path.

explorer/puller/tests/model/test_SymbolNodeConfig.py
daoka: Agreed that the direct-construction test was not adding useful caller-observable coverage. I removed that duplicate and kept the behavior covered through the public factories and request validation tests.

explorer/puller/tests/model/test_SymbolNodeConfig.py
daoka: This is intentional SSRF hardening. Link-local addresses can expose host or cloud metadata services, so they remain rejected even when private addresses are explicitly allowed. The test keeps that policy visible.

explorer/common/symbol/NodeConfiguration.py
```diff
@@ -0,0 +1,193 @@
+import ipaddress
```
Jaguar0625, line 1: please rename file (and test file) to match; NodeConfig => NodeConfiguration

explorer/common/symbol/NodeConfig.py
```diff
@@ -0,0 +1,193 @@
...
+	def assert_request_allowed(self, request_url):
+		"""Validates a Symbol node request target against the configured allowlist and address policy."""
+
+		scheme, host, port, base_url = _normalize_base_url(request_url)
+		if (scheme, host, port) != (self.scheme, self.host, self.port):
+			raise SymbolNodeConfigurationError('Symbol node request target does not match configured base URL')
+		if f'{host}:{port}' not in self.allowed_hosts:
+			raise SymbolNodeConfigurationError('Symbol node request target is not allowed')
+		if _is_metadata_service_host(host):
+			raise SymbolNodeConfigurationError('Metadata service Symbol node host is not allowed')
```
Jaguar0625: you are doing same validation in _from_values. consider refactoring common checks into a helper function

explorer/rest/tests/db/test_SymbolDatabase.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+class SymbolDatabaseTest(unittest.TestCase):
+	def test_check_connection_executes_select_one(self):
+		# Arrange:
+		with PostgresTestDatabase() as db_config:
+			database = SymbolDatabase(db_config)
+
+			# Act:
+			result = database.check_connection()
```
Jaguar0625, line 53: consider adding a test when this returns False too

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -0,0 +1,106 @@
...
+			cursor.execute("INSERT INTO symbol_sync_state (id, status) VALUES (1, 'unknown')")
+
+		database.connection.rollback()
+
+	def test_check_connection_executes_select_one(self):
+		# Arrange:
+		database = self._create_database()
+
+		# Act:
+		result = database.check_connection()
```
Jaguar0625, line 103: consider adding test when this returns False

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -0,0 +1,78 @@
...
+		db_config = config['symbol_db']
+
+		network = _get_symbol_network(network_type)
+
+		self.symbol_db = SymbolDatabase(DatabaseConfig(**db_config))
+		self.node_config = node_config or SymbolNodeConfiguration.from_url(node_url)
+		symbol_node_endpoint = self.node_config.assert_request_allowed(self.node_config.base_url)
+		self._symbol_connector = SymbolConnector(symbol_node_endpoint)
+		self._symbol_connector.timeout_seconds = self.node_config.timeout_seconds
+		self.symbol_facade = SymbolFacade(str(network))
```
Jaguar0625: you should not need str here. SymbolFacade accepts string or Network.

explorer/puller/tests/conftest.py
```diff
@@ -0,0 +1,6 @@
+import sys
```
Jaguar0625: why do you need this as opposed to setting path in test script?

explorer/puller/tests/PostgresTestUtils.py
```diff
@@ -0,0 +1,40 @@
+import os
+from collections import namedtuple
+
+import testing.postgresql
+
+DatabaseConfig = namedtuple('DatabaseConfig', ['database', 'user', 'password', 'host', 'port'])
```
Jaguar0625: Config => Configuration

explorer/common/symbol/NodeConfiguration.py
Jaguar0625: what machine is 169.254.169.254 ?

explorer/common/tests/test_NodeConfig.py
```diff
@@ -0,0 +1,232 @@
...
+def _assert_rejects_allowed_hosts(allowed_hosts):
+	with pytest.raises(
+		SymbolNodeConfigurationError,
+		match='SYMBOL_NODE_ALLOWED_HOSTS entries must be exact host:port values'
+	):
+		SymbolNodeConfiguration.from_app_config(_base_config(SYMBOL_NODE_ALLOWED_HOSTS=allowed_hosts))
+
+
+def test_missing_node_url_returns_none():
+	assert SymbolNodeConfiguration.from_app_config({}) is None
```
Jaguar0625: would it be better to throw an exception ?

explorer/common/tests/test_NodeConfig.py
```diff
@@ -0,0 +1,232 @@
...
+	assert 'localhost' == node_config.host
+	assert 3000 == node_config.port
+	assert 'http://localhost:3000' == node_config.base_url
+	assert frozenset({'localhost:3000'}) == node_config.allowed_hosts
+	assert node_config.allow_private
+	assert node_config.allow_loopback
+	assert 15 == node_config.timeout_seconds
+
+
+def test_to_dict_hides_allowlist():
```
Jaguar0625: allowlist => allowed_hosts ?

explorer/common/tests/test_NodeConfig.py
```diff
@@ -0,0 +1,232 @@
...
+	)
+
+	assert 'http' == node_config.scheme
+	assert 'localhost' == node_config.host
+	assert 3000 == node_config.port
+	assert 'http://localhost:3000' == node_config.base_url
+	assert frozenset({'localhost:3000'}) == node_config.allowed_hosts
+	assert node_config.allow_private
+	assert node_config.allow_loopback
+	assert 15 == node_config.timeout_seconds
```
Jaguar0625: are these two tests testing different behavior? they seem to be both checking normalization of host and allowed host?

explorer/common/tests/test_NodeConfiguration.py
```diff
@@ -0,0 +1,232 @@
...
+
+def test_skips_empty_allowlist_entries():
+	node_config = SymbolNodeConfiguration.from_app_config(_base_config(SYMBOL_NODE_ALLOWED_HOSTS='localhost:3000,'))
+
+	assert frozenset({'localhost:3000'}) == node_config.allowed_hosts
+
+
+@pytest.mark.parametrize('allowed_hosts', [
+	'localhost',
+	'http://localhost:3000',
```
Jaguar0625, line 102: why is this one rejected?

explorer/common/tests/test_NodeConfig.py
```diff
@@ -0,0 +1,232 @@
...
+		SymbolNodeConfiguration.from_app_config(_base_config(
+			SYMBOL_NODE_URL='http://169.254.169.254:3000',
+			SYMBOL_NODE_ALLOWED_HOSTS='169.254.169.254:3000'
+		))
+
+
+def test_accepts_boolean_config_values():
+	node_config = SymbolNodeConfiguration.from_app_config(_base_config(
+		SYMBOL_NODE_ALLOW_PRIVATE=True,
+		SYMBOL_NODE_ALLOW_LOOPBACK='false'
```
Jaguar0625: consider parameterizing or elsewhere testing both parsing of strings 'true' and 'false'

explorer/common/tests/test_NodeConfig.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+def test_rejects_different_target():
+	node_config = SymbolNodeConfiguration.from_app_config(_base_config())
+
+	with pytest.raises(SymbolNodeConfigurationError, match='Symbol node request target does not match configured base URL'):
+		node_config.assert_request_allowed('http://localhost:3001')
+
+
+def test_rejects_removed_request_target():
```
Jaguar0625: removed => not allowed ?

explorer/common/symbol/NodeConfiguration.py
Jaguar0625: Who will be the callers in this case?

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -0,0 +1,123 @@
...
+		# Assert:
+		self.assertFalse(facade.is_configured())
+		self.assertEqual(create_symbol_health(
+			nodeConfigured=True,
+			errors=[{
+				'type': 'database',
+				'message': 'Symbol database is unavailable'
+			}]
+		), result)
+		self.assertNotIn('connection refused', str(result))
```
Jaguar0625: not sure if this assert adds anything since this string does not seem to ever appear ?

explorer/rest/rest/facade/SymbolRestFacade.py
```diff
@@ -0,0 +1,72 @@
...
+				log.error(f'Failed to check Symbol database health: {error}')
+				errors.append({
+					'type': 'database',
+					'message': DATABASE_UNAVAILABLE_MESSAGE
+				})
+
+		return {
+			'isHealthy': db_up and not errors,
+			'dbUp': db_up,
+			'nodeConfigured': True,
```
Jaguar0625: if this will always be True, can remove since it provides no information to caller

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -0,0 +1,162 @@
...
+	with tempfile.TemporaryDirectory() as temp_directory:
+		db_config_path = _create_config_file(temp_directory, database_config=symbol_database_config)
+		app_config_path = _create_app_config(temp_directory, db_config_path)
+		app = Flask(__name__)
+
+		with _rest_settings_env(app_config_path):
+			facade = setup_symbol_facade(app)
+
+	assert facade.is_configured()
+	assert 'http://localhost:3000' == facade.node_config.base_url
```
Jaguar0625: i believe these tests are intended to test the Flask app? in which case, everything should be routed to Flask and the result from the server should be checked, not the facade (tested in test_SymbolRestFacade).

for example, i'd expect more tests to look like test_symbol_health_with_database.

explorer/rest/tests/test/PostgresTestUtils.py
```diff
@@ -0,0 +1,45 @@
+import os
```
Jaguar0625: can you have only a single copy of PostgresTestUtils ?

Jaguar0625 (commented):
👍

explorer/common/symbol/NodeConfiguration.py
daoka: Renamed.

explorer/common/symbol/NodeConfig.py
daoka: I refactored construction so both from_app_config() and from_url() normalize the URL once and share _from_normalized_values() for the common allowed-host, metadata-host, boolean, and timeout validation.

explorer/rest/tests/db/test_SymbolDatabase.py
daoka: Added the false-path coverage for check_connection() by making the query return a value other than 1 and asserting that the method returns False.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Added the corresponding Puller database test for the False path when SELECT 1 does not return 1.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Updated this to pass the Network value directly to SymbolFacade instead of converting it with str().

explorer/puller/tests/conftest.py
daoka: The original intent was only to make the shared common package importable from puller tests after moving the Symbol node configuration there. I agree this belongs in the package test/lint scripts rather than a test-side path mutation, so I removed the conftest.py workaround and now rely on PYTHONPATH=.:.. in the CI scripts.

explorer/puller/tests/PostgresTestUtils.py
daoka: Renamed DatabaseConfig to DatabaseConfiguration to keep the naming consistent with NodeConfiguration and the rest of the configuration objects.

explorer/common/tests/test_NodeConfig.py
daoka: Yes. The original behavior treated a missing node URL like an unconfigured state, but this is required configuration for Symbol setup. I changed it to raise SymbolNodeConfigurationError at configuration creation/setup time and added tests for missing and invalid values. Also, 169.254.169.254 is not intended to be a Symbol node here; it is the well-known link-local cloud instance metadata endpoint used by cloud providers. The test covers rejecting that kind of metadata-service target so the configured node URL cannot accidentally point at an internal metadata service.

explorer/common/tests/test_NodeConfig.py
daoka: Updated the terminology to allowed_hosts / SYMBOL_NODE_ALLOWED_HOSTS throughout the implementation and tests, including the public docstring.

explorer/common/tests/test_NodeConfig.py
daoka: The original intent was to cover two public construction paths: app config loading, including the allowed-host policy, and direct URL-based construction. I kept that distinction but simplified the assertions so they focus on caller-visible behavior instead of duplicating helper-level normalization checks.

explorer/common/tests/test_NodeConfiguration.py
daoka: This is rejected intentionally because SYMBOL_NODE_ALLOWED_HOSTS is a host policy, not a URL policy. The node URL carries the scheme, while allowed hosts must be exact host:port values only. I made this explicit in the validation and tests.

explorer/common/tests/test_NodeConfig.py
daoka: Added parametrized coverage for boolean configuration parsing so both string values, true and false, are exercised explicitly.

explorer/common/tests/test_NodeConfig.py
daoka: The original intent was a defensive request-time check, but the wording and branch were not very useful. I removed that unreachable-style branch and now reject mismatched request targets explicitly as “does not match configured base URL.”

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: That assertion was not adding meaningful coverage. I changed the test to assert the stable health response JSON, including the sanitized Symbol database is unavailable message, instead of relying on an implementation-specific string.

explorer/rest/rest/facade/SymbolRestFacade.py
daoka: The original intent was to expose setup/configuration status separately from runtime health details. It is not always True now: it returns False when Symbol DB initialization fails even if the node configuration exists. I kept it and added coverage for that false case.

explorer/rest/tests/test_symbol_core_routes.py
daoka: The original tests were meant to cover route setup/error handling, but they were too close to facade behavior. I reworked the route tests to go through the Flask app/test client and assert the server response, while keeping facade-specific behavior in test_SymbolRestFacade.

explorer/rest/tests/test/PostgresTestUtils.py
daoka: Consolidated the duplicated PostgreSQL test helper into common.tests.PostgresTestUtils and updated REST/Puller tests to import the shared helper. There is now only one PostgresTestUtils.py copy.

Jaguar0625 (approved):
👍

Wayonb (approved):
👍🏾

PR #2300, opened 2026-06-17
[explorer] fix: integrate frontend with latest REST

AnthonyLaw:
Problem: The old REST structure require update. 
Solution: Updated to the latest REST structure.


explorer/frontend/__tests__/api/accounts.test.js
```diff
@@ -38,7 +38,7 @@ describe('api/accounts', () => {
 				pageSize: 123,
 				mosaic: 'custom.mosaic'
 			};
-			const expectedURL = 'https://explorer.backend/mosaic/rich/list?limit=123&offset=123&namespaceName=custom.mosaic';
+			const expectedURL = 'https://explorer.backend/mosaic/rich/list?limit=123&offset=123&namespace_name=custom.mosaic';
```
Wayonb, line 41: Why is the ``namespace_name`` snake vs the others?

Wayonb (approved):
👍🏾

explorer/frontend/__tests__/api/accounts.test.js
AnthonyLaw: Actually, I was using `snake_case` because of the Python environment, and I noticed that some of the params use `camelCase`. I think we should make them consistent.

Which one do you think better practice for url params?

PR #2301, opened 2026-06-18
[explorer/frontend] feat: dual-chain explorer: one codebase, per-variant builds (NEM + Symbol)

OlegMakarenko:
## Problem

The explorer frontend was a single-purpose **NEM** application. NEM-specific concerns — the
REST/data layer, theme colours and fonts, branding assets, document `<head>`, and a handful of
NEM-only config URLs — were hardcoded throughout the shared pages, components, and stylesheets.

Standing up an explorer for another chain (Symbol) under that shape meant forking the project: a
second repository to keep in sync, duplicated pages and components, and divergent bug fixes. That
does not scale past two chains and guarantees drift.

We want **one codebase** that compiles into a separate, fully-branded explorer per blockchain,
with no protocol details leaking into the shared layer and no runtime cost for variants that are
not being served.

### Shortly

1. Added project configuration and structure for multiple variants. Added the variant contract enforcement.
2. Moved NEM-specific parts into variants/nem.
3. Moved NEM assets (images, fonts, etc) into public/nem.
4. Added Symbol specific theme in variants/symbol.
5. Added Symbol assets (images, fonts, etc) into public/symbol.
6. Added Symbol api temporary stub to variants/symbol.
7. Other minor changes.

### Details

The explorer is now a **protocol-agnostic shell** plus a set of self-contained **variants**. All
shared behaviour (routing, pages, components, hooks, layout) lives at the top level and never names
a specific chain; everything chain-specific — data layer, theme tokens, branding assets, document
head, runtime config, and any variant-only page sections — lives under `variants/<id>/` and is
reached only through a thin **registry**. The active variant is chosen at **build time** via
`NEXT_PUBLIC_EXPLORER_VARIANT` (`nem` | `symbol`), where a build-time `@/app/active-variant/*` alias
resolves each registry import to exactly one variant. As a result only the selected variant is
compiled into the bundle and the other is physically excluded — a Symbol build carries none of NEM's
code, theme, or assets, and vice-versa. The payoff is the maintainability of a single codebase —
shared pages and components are written and fixed once and reused by every chain — kept honest by a
JSDoc **contract** and a smoke test that hold each variant's surface in lockstep. And because the
switch happens at build time, additional chains add zero bundle cost to anyone else: each variant
ships as its own lean image with its own pre-compiled theme, while all non-theme settings stay
runtime-injected via `window.appConfig`.

### Project structure

The shared shell is unchanged in shape; everything new is the `variants/` tree and the registry
modules that front it.

```plaintext
explorer/frontend/
├── api/                      # Stable per-domain interface; each module re-exports the active variant
├── components/               # Shared React components (chain-agnostic)
│   └── AdditionalSections.jsx  # Renders variant-declared page sections
├── config/index.js           # Base runtime config, merged with the variant's config
├── pages/                    # Shared routes/pages (NEM + Symbol render from the same files)
│   ├── _document.jsx           # Renders the active variant's DocumentHead
│   └── index.jsx               # Home page; pulls section list from the variant page config
├── styles/
│   ├── variables.scss          # Intentionally empty stub; real tokens injected per variant
│   └── globals.scss + *.module.scss  # Consume $tokens, no hardcoded colours/fonts
├── public/
│   ├── nem/ …                  # NEM branding assets (favicon, logos, images, manifest)
│   └── symbol/ …               # Symbol branding assets (+ local fonts)
├── utils/client.js            # createAssetURL() — prefixes the active variant id
└── variants/
    ├── contract.js            # The surface every variant must satisfy
    ├── manifest.js            # Registered variant set (test-only single source of truth)
    ├── active.js              # Build-time variant id (process.env)
    ├── api|components|configs|document|page-config|styles.js   # Per-concern registry resolvers
    ├── nem/                   # NEM: api, config, components, styles, DocumentHead
    └── symbol/                # Symbol: stubbed api + fixtures, config, components, styles, DocumentHead
```

### Code responsibilities

| Area | Lives in | Responsibility |
|------|----------|----------------|
| Shared pages & components | `pages/`, `components/` | Chain-agnostic UI and routing, written once and reused by every variant. Read variant data only through `@/app/api/*` and variant config. |
| Stable API interface | `api/<domain>.js` | Thin re-exports that delegate to the active variant (`api.<domain>`), so pages keep their existing per-domain import contract while the implementation is swapped per variant. |
| Variant data layer | `variants/<id>/api/*` | The actual per-domain fetchers for that chain (NEM: real REST; Symbol: static stub fixtures). |
| Variant page config | `variants/<id>/config/pages/*.json` | Per-page composition flags — which optional fields show (e.g. `showSupernodeCount`) and which `additionalSections` to inject. |
| Variant runtime config | `variants/<id>/config/index.js` | Chain-only runtime settings (e.g. NEM's supernode/nodelist/market URLs), merged into the base `config/index.js`. |
| Variant components | `variants/<id>/components/index.js` | Map of variant-only page sections, resolved by name from page config and rendered via `AdditionalSections`. |
| Global styles | `styles/globals.scss`, `styles/**/*.module.scss` | Token-driven styling — reference `$color-*`, `$font-*`, etc.; no hardcoded theme values. |
| Variant theme (SCSS) | `variants/<id>/styles/variables.scss` | The chain's actual `$token` values, injected into every stylesheet at build time. |
| Variant theme (JS) | `variants/<id>/styles/variables.json` | Theme tokens needed in JavaScript (chart/progress/transaction-square colours). |
| Branding assets | `public/<id>/…` | Per-variant favicon, logos, images, web manifest (+ Symbol's local fonts), reached via `createAssetURL`. |
| Document head | `variants/<id>/DocumentHead.jsx` | The chain's `<head>`: favicon, theme-color, description, manifest, and fonts. |

### Variant contract

Every registered variant exports the same surface through its barrel (`variants/<id>/index.js`).
`variants/contract.js` declares the required shape and `__tests__/variants/contract.test.js` walks
the manifest and asserts every variant satisfies it, so a missing export or a drifted theme token
fails the suite rather than the build.

| Surface | Variant export | What it must provide | Enforced by |
|---------|----------------|----------------------|-------------|
| Data API | `api.<domain>` | All 9 domains (`accounts`, `blocks`, `health`, `mosaics`, `namespaces`, `nodes`, `search`, `stats`, `transactions`), each exposing the required fetchers | `API_CONTRACT` |
| Theme tokens (JS) | `styleVariables` (`styles/variables.json`) | Exactly the chart/progress/transaction-square colour keys — same key set across variants | `STYLE_VARIABLES_CONTRACT` |
| Theme tokens (SCSS) | `styles/variables.scss` | Every `$token` the shared stylesheets consume, and an identical token set across variants | `THEME_STYLESHEET_DIRS` + token parity test |
| Page config | `pageConfig` (`config/pages`) | Per-page keys such as `home.showSupernodeCount` and `home.additionalSections` | `PAGE_CONFIG_CONTRACT` |
| Runtime config | `config` (`config/index.js`) | The variant's required runtime config keys (NEM: 4 URL keys; Symbol: none yet) | `CONFIG_CONTRACT` |
| Document head | `DocumentHead` | A React component for the variant `<head>` | contract test (is a function) |
| Section components | `components` | A name → component map for variant-only page sections | contract test (is an object) |

### How the variant switch works

The switch is a **single environment variable resolved three ways that must stay in sync**, so the
editor, the test runner, and the production build all pick the same variant:

1. **`NEXT_PUBLIC_EXPLORER_VARIANT`** (`nem` | `symbol`) selects the variant. It is build-time
   because the SCSS theme is compiled per variant; everything else stays runtime.
9. **Registry indirection.** Shared code imports per-concern resolvers (`@/app/variants/api`,
   `.../document`, `.../page-config`, `.../components`, `.../configs`, `.../styles`), each of which
   re-exports `@/app/active-variant/<concern>`. Nothing shared imports a concrete variant directly.
10. **`@/app/active-variant/*` is aliased to the selected variant** in three places that mirror each
   other:
   - **Build** — `next.config.js` `turbopack.resolveAlias` maps it to `./variants/<VARIANT>/…`
     (root-relative). Because only the active variant is on the import graph, **the inactive
     variant is excluded from the bundle** (`resolveAlias` takes precedence over the jsconfig path).
   - **Tests** — `jest.config.js` derives `moduleNameMapper` from `jsconfig.json` and adds the same
     `^@/app/active-variant/(.*)$` → active-variant mapping, pinned to a reference variant.
   - **Editor / lint** — `jsconfig.json` points `@/app/active-variant/*` at `variants/nem/*` so
     imports resolve in the IDE and the import-resolver lint rules.
11. **SCSS tokens** are injected by `next.config.js` `sassOptions.additionalData`, which `@import`s
   the active variant's `styles/variables.scss` into every stylesheet.
12. **Assets** are resolved through `createAssetURL(url)` (`utils/client.js`), which prefixes the
   active variant id, so a shared `/images/logo.png` reference loads from `/<variant>/images/logo.png`.
13. **Runtime config** (`config/index.js`) merges the variant's `config` into the base config on the
   server and reads everything from `window.appConfig` on the client — so non-theme settings need no rebuild.

> **Note:** because the variant id and the SCSS tokens are baked in at build time, switching variant
> (or editing a variant's `variables.scss`) requires `rm -rf .next` and a dev restart / rebuild —
> Turbopack does not track the injected tokens file as a dependency.

### How to build and run

The active variant is supplied per command. `package.json` ships explicit per-variant scripts
(via `cross-env`), and the `.env` `NEXT_PUBLIC_EXPLORER_VARIANT` is the default for the bare scripts.

| Task | NEM | Symbol |
|------|-----|--------|
| Dev server | `npm run dev:nem` | `npm run dev:symbol` |
| Production build | `npm run build:nem` | `npm run build:symbol` |
| Start built app | `npm run start:nem` | `npm run start:symbol` |

```bash
# Build + run a variant end-to-end
npm run build:symbol && npm run start:symbol

# Per-variant Docker image — NEXT_PUBLIC_EXPLORER_VARIANT build-arg is REQUIRED (build fails fast without it)
docker build --build-arg NEXT_PUBLIC_EXPLORER_VARIANT=nem    -t symbolplatform/explorer-frontend .
docker build --build-arg NEXT_PUBLIC_EXPLORER_VARIANT=symbol -t symbolplatform/explorer-frontend-symbol .
```

Each variant therefore produces its own build and its own image — one image per chain, with that
chain's theme already compiled in.

## Summary of changes

The commit history walks the migration in dependency order — build wiring → registry/contract →
extracting NEM into a variant → adding the Symbol variant → mechanical cleanup. Grouped by theme:

**Build & tooling**
- Wired the build-time variant switch: `next.config.js` (Turbopack `resolveAlias` + per-variant
  SCSS injection), `jsconfig.json` (`@/` → `@/app/` aliases plus `@/app/active-variant/*`), and
  `jest.config.js` (mirrors the active-variant mapping so tests resolve the same modules as the build).
- Added `cross-env` and per-variant `dev`/`build`/`start` scripts.
- Made the Docker build per-variant: `NEXT_PUBLIC_EXPLORER_VARIANT` is a required build-arg, baked
  into the image, with an expanded `.dockerignore`.

**Variant framework**
- Introduced the registry layer (per-concern resolver barrels, the `active` env accessor, and a
  test-only `manifest`), the JSDoc **contract**, and the **smoke test** that asserts every variant
  satisfies it.
- Added variant-driven composition: `DocumentHead` rendered through `pages/_document.jsx`, and
  `AdditionalSections` rendering variant-declared page sections via the variant component map.

**NEM extraction (no behavioural change)**
- Moved the NEM API bodies into `variants/nem/api/*`; reduced the shared `api/<domain>.js` modules
  to thin re-exports so pages keep their import contract.
- Moved NEM runtime config, theme tokens (SCSS + JSON), and branding assets into the variant;
  tokenized the shared stylesheets (replaced hardcoded colours/filters with `$token` references);
  relocated assets to `public/nem/` and routed every reference through `createAssetURL`.
- Mechanically migrated the `@/` import alias to `@/app/` across the remaining files.

**Symbol variant (new)**
- Added the Symbol variant behind `NEXT_PUBLIC_EXPLORER_VARIANT=symbol`: full **branding** —
  dark/purple theme tokens (SCSS + JSON), `DocumentHead`, **local fonts** (Protipo, SofiaSansCondensed,
  Carbon, JetBrainsMono, linked from `public/symbol/fonts/fonts.css`), and `public/symbol` assets —
  plus page config (including a `Finalization` stub section) and component map.

> ⚠️ **Symbol has no real data layer yet.** Its `variants/symbol/api/*` modules return **static stub
> fixtures** (shapes mirror the mapped NEM output) so the shared pages render for visual/QA
> verification of the variant switch — there is no `symbol-sdk`, no network, and no node calls. The
> stub layer (`variants/symbol/api/*`, `fixtures.js`, `stub.js`, and the `Finalization` placeholder)
> is intended to be **removed/replaced by the real Symbol data layer in a follow-up PR**. The NEM
> build is functionally untouched by this change.

### Validation
- The shared shell is reused unchanged by both variants; the variant **contract test** guards the
  surface and theme-token parity between NEM and Symbol.
- `next build` (Turbopack) compiles both `nem` and `symbol`, with only the selected variant present
  in each bundle (verified: a Symbol build contains none of NEM's chart hex values, and vice-versa).
</content>
</invoke>


explorer/frontend/config/index.js
```diff
@@ -1,14 +1,14 @@
...
 	BLOCKCHAIN_UNWIND_LIMIT: +process.env.BLOCKCHAIN_UNWIND_LIMIT,
 	REQUEST_TIMEOUT: +process.env.REQUEST_TIMEOUT,
 	API_BASE_URL: process.env.API_BASE_URL,
-	SUPERNODE_API_URL: process.env.SUPERNODE_API_URL,
-	NODELIST_URL: process.env.NODELIST_URL,
-	MARKET_DATA_URL: process.env.MARKET_DATA_URL,
-	HISTORICAL_PRICE_URL: process.env.HISTORICAL_PRICE_URL,
 	SOCIAL_URL_TWITTER: process.env.SOCIAL_URL_TWITTER,
 	SOCIAL_URL_GITHUB: process.env.SOCIAL_URL_GITHUB,
 	SOCIAL_URL_DISCORD: process.env.SOCIAL_URL_DISCORD,
```
daoka: This may not be directly related to this change, but I’d like to confirm one thing.

Several values read in `config/index.js` are later serialized into `window.appConfig` in `_app.jsx`, so their values become available to browser-side code. For example:

- `NATIVE_MOSAIC_ID`
- `NATIVE_MOSAIC_TICKER`

Next.js documents that non-`NEXT_PUBLIC_` environment variables are only available in the Node.js environment and are not directly accessible to the browser. In this implementation, these values are not accessed directly from `process.env` in the browser, but they are still intentionally exposed to the browser through `window.appConfig`.

Given that, should we rename the browser-visible configuration inputs to use the `NEXT_PUBLIC_*` prefix, for example:

- `NEXT_PUBLIC_NATIVE_MOSAIC_ID`
- `NEXT_PUBLIC_NATIVE_MOSAIC_TICKER`

This would make the public nature of these values explicit and align the env contract with Next.js conventions.

Reference:
https://nextjs.org/docs/pages/guides/environment-variables#bundling-environment-variables-for-the-browser

I do not think these values are secrets, so this is mainly about clarity and avoiding confusion between server-only env vars and values that are intentionally exposed to the client.

explorer/frontend/.dockerignore
```diff
@@ -1,3 +1,13 @@
 /node_modules
 
 /build
+/.next
+/coverage
+
+/__tests__
+
+.env
```
AnthonyLaw, line 9: Since `.env` was added to `.dockerignore`, it would be good to add `env_file` in `docker-compose.yaml` so Compose still loads the environment variables.

AnthonyLaw (commented):
Looks good overall.

1. I noticed some duplicated images such as `nem/images/avatars/*` and `symbol/images/avatars/*`.
2. I also noticed duplicated sections in `styles/variables.css` such as `Borders` , `Spacings General`, `Spacings Large` and other.

Was that intentional ?

explorer/frontend/config/index.js
OlegMakarenko: Thanks for the comment! I updated the config structure, and documentation for clarity.

explorer/frontend/.dockerignore
OlegMakarenko: Thanks, added.

daoka (approved):

daoka (commented):
👍

cryptoBeliever (approved):
I've tested the NEM testnet explorer, and the frontend still seems to be working fine 👍

AnthonyLaw (approved):
Look good 👍🏼

explorer/frontend/variants/nem/api/search.js
```diff
@@ -0,0 +1,40 @@
...
+
+export const search = async text => {
+	const query = `${text}`.trim().toUpperCase();
+	const results = {};
+	let account;
+	let transaction;
+	let namespace;
+
+	const block = await fetchBlockInfo(query);
+	if (query.length === 64)
```
Wayonb, line 15: Why pick the query type based on length? 🤔
There is no information in the text to identify the query?

explorer/frontend/variants/nem/api/stats.js
```diff
@@ -0,0 +1,143 @@
...
+		return response.data.map(item => [`${item.period}-01`, item.totalTransactions]);
+	}
+	default:
+		return (await fetchBlockPage({ pageSize: 240 })).data.map(item => [item.height, item.transactionCount]).reverse();
+	}
+};
+
+export const fetchTransactionStats = async () => {
+	const stats = await makeRequest(createApiUrl('transaction/statistics'));
+	const blocks = (await fetchBlockPage({ pageSize: 240 })).data;
```
Wayonb, line 65: Should the pagesize be in config?

Wayonb (changes requested):

explorer/frontend/variants/nem/api/stats.js
OlegMakarenko: Thanks, yes, it should be in config/constants. But I'd better do it in next PR, since this PR does not change actual NEM variant implementation. What do you think?

Wayonb (approved):
👍🏾

PR #2302, opened 2026-06-20
[explorer/fronted] fix: missing copy i18 config

AnthonyLaw:
Problem: Frontend Docker images having the error
```sh
app-1  | ⨯ Error: next-i18next was unable to find a user config at /app/next-i18next.config.js
app-1  |     at new Promise (<anonymous>)
app-1  | react-i18next:: You will need to pass in an i18next instance by using initReactI18next
```
Solution: copy `next-i18next.config.js` to images


Wayonb (approved):
👍🏾

OlegMakarenko (approved):

PR #2324, opened 2026-06-24
[explorer/frontend] fix: preserve embedded transaction value

AnthonyLaw:
Problem: The page returns a 500 error when formatting a multisig transaction whose embedded transaction value is not preserved.
Solution: 
- Preserve the embedded transaction value as flat payload data when formatting frontend transaction responses.
- Patched fixtures as well


Wayonb (approved):
👍🏾

PR #2325, opened 2026-06-24
[explorer] fix: change 0 to represent no expiration and update the page

AnthonyLaw:
Problem: 
- I just noticed, in the frontend, it uses `0` as a presentation of no expiration.
- Namespace page didn't handle an expired namespace

Solution: update `seed_network_currency`, and handle namespace page and listing.

Screenshot 2026-06-25 at 2 06 25 AM
Screenshot 2026-06-25 at 2 06 15 AM


Wayonb (approved):
👍🏾

PR #2343, opened 2026-06-26
[explorer/puller] fix: format vested balance to absolute when running refresh account method

AnthonyLaw:
Problem: Vested balance didn't convert to absolute when running refresh account method [Detail](https://github.com/symbol/product/issues/2340)

Solution: convert vested balance before update in db.


Wayonb (approved):
👍🏾

PR #2344, opened 2026-06-28
[explorer/frontend] fix: replace a new market data url

AnthonyLaw:
Problem: The existing market data URL is no longer valid. #2311 
Solution: Replace another URL.


explorer/frontend/api/stats.js
```diff
@@ -113,14 +113,14 @@ export const fetchNodeStats = async () => {
 
 export const fetchMarketData = async () => {
 	const response = await makeRequest(config.MARKET_DATA_URL);
```
Wayonb, line 115: I guess we switch to coingecko.com?

Wayonb (approved):
👍🏾

explorer/frontend/api/stats.js
AnthonyLaw: Yes, that is correct 👍🏼

PR #2349, opened 2026-06-28
[explorer] fix: convert mosaic amount to relative format.

AnthonyLaw:
Problem: Account mosaic balances are displayed without divisibility on the account details page https://github.com/symbol/product/issues/2341
Solution: Format the mosaic amount to a relative value.


Wayonb (approved):
👍🏾

PR #2360, opened 2026-06-30
[explorer/backend]: add Symbol block header backend

daoka:

Symbol backend currently has the core foundation, configuration validation, and health/status plumbing, but it does not provide the block-header synchronization and REST read path for Symbol block data.

## What's the issue?

The Explorer frontend cannot consume Symbol block list/detail data from the new backend unless the Puller can persist block headers and the REST API can read them through a stable response contract.

This also needs explicit synchronization state handling so REST can distinguish unavailable, lagging, repairing, unhealthy, and readable Symbol backend data.

Added the Symbol block-header backend slice.

* Added Symbol Puller storage for `symbol_sync_state` and `symbol_blocks`, including block type/status enums, indexes, upserts, block hash lookup, and rollback repair deletion.
* Added Symbol Puller block-header sync from the configured node, including pagination, finalized watermark handling, capped manual sync height, node error retry handling, and shallow rollback repair.
* Added a `sync_symbol_block` Puller workflow for manual/backend sync execution.
* Added Symbol REST block read support:
  * `GET /api/symbol/blocks`
  * `GET /api/symbol/block/<height>`
* Added REST response formatting for Symbol block list/detail fields.
* Extended Symbol health output with sync-state details such as finalized height, last DB height, backend sync status, and rollback/unhealthy reporting.
* Updated backend Python CI scripts so Puller/REST checks include the required Symbol/common import paths.
* Added and updated PostgreSQL-backed, facade, route, workflow, and response-model tests for the new behavior.

No intentional NEM behavior changes are included.


explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -1,44 +1,110 @@
...
+		_create_enum_type(cursor, 'symbol_block_type', BLOCK_TYPE_VALUES)
+		_create_enum_type(cursor, 'symbol_sync_state_status', SYNC_STATE_STATUS_VALUES)
+		_create_table(cursor, 'symbol_sync_state', SYMBOL_SYNC_STATE_DEFINITIONS)
+		_create_table(cursor, 'symbol_blocks', SYMBOL_BLOCK_DEFINITIONS)
+		cursor.execute('CREATE INDEX IF NOT EXISTS idx_symbol_blocks_height_desc ON symbol_blocks(height DESC)')
+		cursor.execute('CREATE INDEX IF NOT EXISTS idx_symbol_blocks_timestamp ON symbol_blocks(timestamp)')
+		cursor.execute('CREATE INDEX IF NOT EXISTS idx_symbol_blocks_signer_address ON symbol_blocks(signer_address)')
 		self.connection.commit()
 
 	def check_connection(self):
```
Jaguar0625, line 110: now that you're populating the db with real data, maybe change this to check if it can query singleton `symbol_sync_state` unless you need to call this when that does not yet exist?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
+			('symbol_block_type', 'importance'),
+			('symbol_block_type', 'normal'),
+			('symbol_sync_state_status', 'initialized'),
+			('symbol_sync_state_status', 'healthy'),
+			('symbol_sync_state_status', 'repairing'),
+			('symbol_sync_state_status', 'unhealthy')
+		], enum_values)
+		self.assertIn('idx_symbol_blocks_height_desc', indexes)
+		self.assertIn('idx_symbol_blocks_signer_address', indexes)
+		self.assertIn('idx_symbol_blocks_timestamp', indexes)
```
Jaguar0625, line 219: i would prefer you break this test up because it's testing too many things. probably into three: tables, enums, indexes.

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
+		cursor = database.connection.cursor()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError) as exception_info:
+			cursor.execute(
+				'INSERT INTO symbol_sync_state (id, status) '
+				'VALUES (2, \'initialized\')'
+			)
+
+		self.assertEqual('23514', exception_info.exception.pgcode)
```
Jaguar0625: are these codes guaranteed to not change across db + sdk versions? if not, we might not want to assert on them.

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
-			def execute(self, _statement):
-				pass
+		database = self._create_database()
+		database.upsert_blocks([_create_block(1), _create_block(2)])
 
-			@staticmethod
-			def fetchone():
-				return (0,)
+		# Act:
+		block_hashes = database.get_block_hashes(1, 2)
```
Jaguar0625, line 454: would prefer seeding a few extra blocks, otherwise an implementation that returns *all* would also pass this test

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
+			(1, b'hash 1'),
+			(2, b'hash 2')
+		], [
+			(height, bytes(hash_value))
+			for height, hash_value in block_hashes
+		])
 
-		database = SymbolDatabase(self.db_config)
-		database.connection = FalseConnection()
+	def test_can_upsert_importance_block_fields(self):
```
Jaguar0625, line 465: is there anything special about "importance block fields" that you specifically wanted to test? because they're optional?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
+			'''
+		)
+		result = cursor.fetchone()
+		(
+			voting_eligible,
+			harvesting_eligible,
+			total_voting,
+			previous_importance_hash,
+			block_type
+		) = result
```
Jaguar0625: would avoid locals and just use `result.` below

Jaguar0625: (also in other tests)

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1,130 +1,598 @@
...
+			_create_block(1),
+			_create_block(2),
+			_create_block(3)
+		])
+
+		# Act:
+		database.repair_rollback_from_height(2, _create_sync_state(
+			status='repairing',
+			last_synced_height=1,
+			last_synced_block_hash=b'hash 1'
```
Jaguar0625, line 566: do you need to explicitly pass sync_state. these seem like something the database should be able to calculate from last record?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -53,21 +77,253 @@ def _validate_symbol_node_path(self, url_path):
...
+		})
+		return height
+
+	async def _sync_block_pages(self, start_height, chain_height, epoch_adjustment_seconds):
+		last_synced_height = None
+		last_synced_block_hash = None
+		offset = start_height - 1
+
+		while offset < chain_height:
+			blocks = await self._get_block_page(offset)
```
Jaguar0625: is there any reason you don't want to parallelize these fetches?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -53,21 +77,253 @@ def _validate_symbol_node_path(self, url_path):
...
+
+		db_finalized_hash = self.symbol_db.get_block_hash(finalized_height)
+		if 0 < finalized_height <= sync_state['last_synced_height'] and not db_finalized_hash:
+			self.symbol_db.upsert_sync_state({
+				**sync_state,
+				'status': 'unhealthy',
+				'finalized_height': finalized_height,
+				'finalized_hash': finalized_hash
+			})
+			raise SymbolRollbackError('Finalized block is missing from local database')
```
Jaguar0625, line 220: is "unhealthy" unrecoverable state ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -53,21 +77,253 @@ def _validate_symbol_node_path(self, url_path):
...
+		if max_height < 1:
+			raise ValueError('max_height must be greater than or equal to 1')
+
+		return min(node_chain_height, max_height)
+
+	@staticmethod
+	def _get_finalized_watermark(chain_info, chain_height):
+		finalized_block = chain_info['latestFinalizedBlock']
+		node_finalized_height = int(finalized_block['height'])
+		is_finalization_capped = node_finalized_height > chain_height
```
Jaguar0625: imo, in this case, chain_height is effectively max finalized block

explorer/puller/tests/facade/symbol/test_Puller.py
```diff
@@ -0,0 +1,290 @@
+import asyncio
+import tempfile
+from unittest import TestCase
+from unittest.mock import AsyncMock, patch
+
+from symbollightapi.model.Exceptions import NodeException
+
+from puller.facade.SymbolPuller import SymbolPuller
+
+from .puller_test_utils import NODE_URL, _create_db_config, _create_symbol_puller, _temporary_symbol_puller
```
Jaguar0625: functions you are importing shouldn't start with "_"

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -53,21 +77,253 @@ def _validate_symbol_node_path(self, url_path):
...
+				attempt = attempt_index + 1
+				if retries == attempt:
+					log.error(f'Failed {description} after {retries} attempts: {error}')
+					raise
+
+				wait_time = delay * (2 ** attempt_index)
+				log.warning(f'Error {description} (attempt {attempt}/{retries}): {error}. Retrying in {wait_time}s...')
+				await asyncio.sleep(wait_time)
+
 	async def get_symbol_node(self, url_path, property_name=None, not_found_as_error=True):
```
Jaguar0625, line 115: is there a scenario where you have a single SymbolPuller instance that where get_symbol_node and/or post_symbol_node are called with different url_path?

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -0,0 +1,319 @@
...
+			db_config_path,
+			network_type,
+			request_timeout_seconds
+		)
+
+
+def _set_symbol_connector(puller, connector):
+	# Keep protected connector replacement in one helper so sync tests can use
+	# deterministic Symbol node responses.
+	puller._symbol_connector = connector  # pylint: disable=protected-access
```
Jaguar0625, line 92: this strongly implies you should be passing connector into ctor vs creating it in ctor

explorer/puller/puller/workflows/sync_symbol_block.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+async def main():
+	"""Synchronizes Symbol block headers."""
+
+	args = parse_args()
+	node_config = _create_node_config(args.symbol_node)
+	puller = SymbolPuller(args.symbol_node, args.db_config, args.network, node_config)
+
+	with puller.symbol_db:
```
Jaguar0625: this is quite awkward construction. either add resource semantics to puller or pass in symbol_db as ctor param

explorer/puller/tests/facade/symbol/test_Puller.py
```diff
@@ -0,0 +1,290 @@
...
+			'testnet',
+			request_timeout_seconds=15
+		) as puller:
+			# Assert:
+			self.assertEqual(NODE_URL, puller.node_config.base_url)
+			self.assertEqual('testnet', puller.symbol_facade.network.name)
+			self.assertEqual('symbol', puller.symbol_db.db_config.database)
+			self.assertIsNone(puller.symbol_db.connection)
+
+	@patch('puller.facade.SymbolPuller.SymbolConnector')
```
Jaguar0625: would strongly suggest you try to avoid patch wherever possible - another hint SymbolConnector should be passed as ctor param

explorer/puller/tests/facade/symbol/test_Puller.py
```diff
@@ -0,0 +1,290 @@
...
+		# Assert:
+		self.assertEqual({'ok': True}, result)
+		symbol_connector.post.assert_awaited_once_with(
+			'path',
+			{'payload': 1},
+			'data',
+			False
+		)
+
+	@patch('asyncio.sleep')
```
Jaguar0625: prefer passing 0 timeout seconds

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -0,0 +1,228 @@
...
+			# Assert:
+			sync_state = database.get_sync_state()
+
+		self.assertEqual(3, sync_state['last_synced_height'])
+		self.assertEqual(
+			bytes.fromhex(f'{3:064X}'),
+			bytes(sync_state['last_synced_block_hash'])
+		)
+
+	def test_sync_block_headers_verifies_unfinalized_hashes_without_rollback(
```
Jaguar0625, line 113: how is this test proving "verifies_unfinalized_hashes" ?

explorer/puller/tests/workflows/test_sync_symbol_block.py
```diff
@@ -0,0 +1,167 @@
...
+			'--symbol-node', 'http://localhost:3000',
+			'--network', 'testnet',
+			'--db-config', 'test_config.ini',
+			'--max-height', '1000')
+
+		# Assert:
+		self.assertEqual(args.symbol_node, 'http://localhost:3000')
+		self.assertEqual(args.network, 'testnet')
+		self.assertEqual(args.db_config, 'test_config.ini')
+		self.assertEqual(1000, args.max_height)
```
Jaguar0625: flip so constants are on same side of all asserts

explorer/puller/tests/workflows/test_sync_symbol_block.py
```diff
@@ -0,0 +1,167 @@
...
+		self.assertEqual((), mock_db.create_tables.call_args.args)
+		self.assertEqual({}, mock_db.create_tables.call_args.kwargs)
+		self.assertEqual(1, mock_facade.sync_block_headers.await_count)
+		self.assertEqual(
+			(3000,),
+			mock_facade.sync_block_headers.await_args.args)
+		self.assertEqual({}, mock_facade.sync_block_headers.await_args.kwargs)
+
+	@patch('asyncio.run')
+	def test_main_guard_runs_main(self, mock_asyncio_run):
```
Jaguar0625: not sure what you're attempting to test here ?

explorer/puller/tests/workflows/test_sync_symbol_block.py
```diff
@@ -0,0 +1,167 @@
...
+			frozenset({'localhost:7890'}),
+			node_config.allowed_hosts)
+		self.assertTrue(node_config.allow_loopback)
+		self.assertFalse(node_config.allow_private)
+		self.assertEqual(17, node_config.timeout_seconds)
+
+	@patch('puller.workflows.sync_symbol_block.SymbolPuller')
+	@patch('puller.workflows.sync_symbol_block.SymbolNodeConfiguration')
+	@patch('puller.workflows.sync_symbol_block.parse_args')
+	def test_main_creates_tables_and_syncs_block_headers(
```
Jaguar0625: would prefer a small integration test instead of mocking everything out

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+					cursor.execute(f'SELECT {", ".join(SYNC_STATE_COLUMNS)} FROM symbol_sync_state WHERE id = 1')
+					result = cursor.fetchone()
+				except PsycopgError as error:
+					connection.rollback()
+					if UNDEFINED_COLUMN_SQLSTATE != error.pgcode:
+						raise
+
+					cursor.execute(f'SELECT {", ".join(LEGACY_SYNC_STATE_COLUMNS)} FROM symbol_sync_state WHERE id = 1')
+					result = cursor.fetchone()
+					return _create_sync_state(LEGACY_SYNC_STATE_COLUMNS, result)
```
Jaguar0625: i would't include migration logic pre-release

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+
+				return _create_sync_state(SYNC_STATE_COLUMNS, result)
+
+	def get_block(self, height):
+		"""Gets a Symbol block by height."""
+
+		sync_state = self.get_sync_state()
+		if not _is_block_data_readable(sync_state):
+			return None
```
Jaguar0625, line 97: generally i'd expose this as member and expect caller to check it. semantics of get* is usually to return not None. try* would imply caller needs to check for None.

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+
+		head_height = sync_state['last_synced_height']
+		finalized_height = sync_state['finalized_height']
+		order = sort.upper()
+		if order not in ('ASC', 'DESC'):
+			raise ValueError('Sort must be either ASC or DESC')
```
Jaguar0625: seems like sort should be enum ?

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+		return sync_state['last_synced_height']
+
+	def get_blocks(self, limit, from_height, sort):
+		"""Gets Symbol blocks using fromHeight cursor pagination."""
+
+		sync_state = self.get_sync_state()
+		if not _is_block_data_readable(sync_state):
+			return None
+
+		head_height = sync_state['last_synced_height']
```
Jaguar0625: consider renaming to something more specific: max_height, local_height, explorer_height ?

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+		with self.connection() as connection:
+			with connection.cursor() as cursor:
+				cursor.execute(
+					f'''
+					SELECT {BLOCK_COLUMNS}
+					FROM symbol_blocks
+					WHERE height BETWEEN %s AND %s
+					ORDER BY height {order}
+					''',
+					(min(start_height, end_height), max(start_height, end_height)))
```
Jaguar0625: would prefer _calculate_height_range to always return start_height <= end_height

explorer/rest/tests/db/test_SymbolDatabase.py
```diff
@@ -1,9 +1,13 @@
...
+from psycopg2 import Error as PsycopgError
+from puller.db.SymbolDatabase import SymbolDatabase as PullerSymbolDatabase
 
 from rest.db.SymbolDatabase import SymbolDatabase
+from rest.model.common import DatabaseConfig
+
+from ..test.SymbolBlockTestUtils import create_symbol_block, create_symbol_importance_block, create_symbol_sync_state
 
 
 class FalseCursor:
```
Jaguar0625: can you remind me why you don't want to use a real database in these tests like the puller SymbolDatabase tests?

explorer/rest/rest/db/SymbolDatabase.py
```diff
@@ -11,3 +90,131 @@ def check_connection(self):
...
+	def get_block_head_height(self):
+		"""Gets the latest synced height for the block list."""
+
+		sync_state = self.get_sync_state()
+		if not _is_block_data_readable(sync_state):
+			return None
+
+		return sync_state['last_synced_height']
+
+	def get_blocks(self, limit, from_height, sort):
```
Jaguar0625: can you swap from_height and limit ? i think that is more standard / readable

explorer/rest/tests/db/test_SymbolDatabase.py
```diff
@@ -43,26 +47,399 @@ def putconn(_connection):
...
+				status='unhealthy'),
+			lambda database: database.get_block_head_height())
 
-			# Act:
-			result = database.check_connection()
+		# Assert:
+		self.assertIsNone(result)
+
+
+class SymbolDatabaseBlocksTest(TestCase):
```
Jaguar0625, line 70: this seems to be using real database, but i'm now confused purpose of SymbolDatabaseBlocksTest since those seem to be same/similar tests only without database?

explorer/rest/tests/db/test_SymbolDatabase.py
```diff
@@ -43,26 +47,399 @@ def putconn(_connection):
...
+					puller_database.upsert_sync_state(sync_state)
+				puller_database.upsert_blocks(blocks)
+
+				database = SymbolDatabase(db_config)
+				try:
+					return query_database(database)
+				finally:
+					database._pool.closeall()  # pylint: disable=protected-access
+			finally:
+				if created_tables:
```
Jaguar0625: not sure if you need check since `drop_symbol_block_tables_if_present` is already checking if tables exist?

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -74,46 +225,215 @@ def test_reports_healthy_core_when_dependencies_are_available(self):
...
 		self.assertEqual(create_symbol_health(
-			isHealthy=True,
-			dbUp=True
+			dbUp=True,
+			errors=[{
+				'type': 'database',
+				'message': 'Symbol database is unavailable'
+			}]
 		), result)
+		self.assertNotIn('database unavailable', str(result))
```
Jaguar0625: i think it would be better to avoid NotIn and search for something in results

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -74,46 +225,215 @@ def test_reports_healthy_core_when_dependencies_are_available(self):
...
+
+	def test_reports_database_error_when_sync_state_read_fails(self):
+		# Arrange:
+		facade = _create_facade_with_database(SyncStateFailingSymbolDatabase())
+
+		# Act:
+		result = facade.get_health()
+
 		# Assert:
 		self.assertEqual(create_symbol_health(
```
Jaguar0625, line 273: in some tests you check full result, in others you don't. for consistency, would prefer former unless you're only checking specific field or two.

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -74,46 +225,215 @@ def test_reports_healthy_core_when_dependencies_are_available(self):
...
+	def test_get_blocks_returns_none_when_database_is_unavailable(self):
+		# Arrange:
+		facade = SymbolRestFacade(
+			create_unreachable_db_configuration(),
+			_create_node_config())
+
+		# Act + Assert:
+		self.assertFalse(facade.is_database_available())
+		self.assertIsNone(
+			facade.get_blocks(limit=1, from_height=None, sort='DESC'))
```
Jaguar0625: do you have your editor line length set to 140? it looks like you're wrapping a bit too aggressively.

explorer/rest/tests/model/symbol/test_Block.py
```diff
@@ -0,0 +1,121 @@
...
+
+	def test_can_convert_to_detail_dict(self):
+		# Arrange:
+		block_view = self._create_block_view()
+
+		# Act:
+		result = block_view.to_detail_dict()
+
+		# Assert:
+		self.assertEqual('03', result['signature'])
```
Jaguar0625: would prefer you test full json - can have helper for list json and just ** into this object.

explorer/rest/tests/model/symbol/test_Block.py
```diff
@@ -0,0 +1,121 @@
...
+		self.assertEqual(4, result['votingEligibleAccountsCount'])
+		self.assertEqual('17', result['harvestingEligibleAccountsCount'])
+		self.assertEqual('19000235663367', result['totalVotingBalance'])
+		self.assertEqual('0A', result['previousImportanceBlockHash'])
+		self.assertEqual('nemesis', result['blockType'])
+
+	def test_can_convert_empty_state_hash_sub_cache_roots(self):
+		# Arrange:
+		block_view = self._create_block_view()
+		block_view.state_hash_sub_cache_roots = []
```
Jaguar0625, line 99: is there anything special about state_hash_sub_cache_roots that you want to test ?

explorer/rest/rest/model/symbol/Block.py
```diff
@@ -0,0 +1,103 @@
...
+def _hex(value):
+	return value.hex().upper()
+
+
+def _str_or_none(value):
+	return str(value) if value is not None else None
+
+
+def _format_total_fee(total_fee):
+	return float(Decimal(total_fee) / (Decimal(10) ** XYM_DIVISIBILITY))
```
Jaguar0625: you might want to consider refactoring these since you might also want to use them in other json object construction

explorer/rest/rest/model/symbol/Block.py
```diff
@@ -0,0 +1,103 @@
...
+
+XYM_DIVISIBILITY = 6
+
+
+def _format_timestamp(timestamp):
+	if hasattr(timestamp, 'isoformat'):
+		formatted_timestamp = timestamp.isoformat()
+		return formatted_timestamp.replace('+00:00', 'Z') if timestamp.tzinfo else f'{formatted_timestamp}Z'
+
+	return str(timestamp).replace(' ', 'T')
```
Jaguar0625: can you explain sources of timestamps that require you to support multiple formats?

explorer/rest/rest/routes/symbol.py
```diff
@@ -35,3 +38,76 @@ def setup_symbol_routes(app, symbol_api_facade):
...
+
+			result = symbol_api_facade.get_blocks(limit, from_height, sort)
+		except ValueError as error:
+			abort(400, error)
+		except PsycopgError:
+			log.error('Failed to get Symbol blocks')
+			return _service_unavailable('Symbol backend data is unavailable')
+
+		if result is None:
+			return _service_unavailable('Symbol backend data is unavailable')
```
Jaguar0625, line 77: shouldn't this be 404 instead of 503?

explorer/rest/rest/routes/symbol.py
```diff
@@ -35,3 +38,76 @@ def setup_symbol_routes(app, symbol_api_facade):
...
+		try:
+			result = symbol_api_facade.get_block(height)
+		except PsycopgError:
+			log.error('Failed to get Symbol block')
+			return _service_unavailable('Symbol backend data is unavailable')
+
+		if not result:
+			abort(404)
+
+		return jsonify(result)
```
Jaguar0625: think you can refactor general handling with previous - only difference is actual call get_blocks vs get_block

explorer/rest/tests/test_app.py
```diff
@@ -11,6 +12,10 @@ def _write_rest_config(config_path, contents):
 	config_path.write_text(contents, encoding='utf8')
 
 
+def _write_db_config(config_path, contents):
+	config_path.write_text(contents, encoding='utf8')
```
Jaguar0625: since body is same, would just have single function - _write_config

explorer/rest/rest/routes/symbol.py
```diff
@@ -35,3 +38,76 @@ def setup_symbol_routes(app, symbol_api_facade):
...
+
+		if result is None:
+			return _service_unavailable('Symbol backend data is unavailable')
+
+		return jsonify(result)
+
+	@app.route('/api/symbol/block/<height>')
+	def api_get_symbol_block_by_height(height):
+		try:
+			height = _parse_block_height(height)
```
Jaguar0625, line 84: are you validating you can parse limit and fromHeight in previous function ?

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -3,16 +3,41 @@
 
 import pytest
```
Jaguar0625, line 4: can you explain what tests you decided to put in this file vs test_routes?

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -106,21 +305,162 @@ def test_setup_rejects_bad_node_url():
...
+	health = facade.get_health()
+	assert isinstance(facade, SymbolRestFacade)
+	assert not facade.is_configured()
+	assert not health['isHealthy']
+	assert not health['dbUp']
+	assert [{
+		'type': 'database',
+		'message': 'Symbol database is unavailable'
+	}] == health['errors']
+	assert 'connection refused' not in str(health)
```
Jaguar0625: similar comment about preferring to avoid  not in asserts (unless there is another similar test that has an in assert)

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -106,21 +305,162 @@ def test_setup_rejects_bad_node_url():
...
+	# Assert:
+	assert DatabaseConfig(
+		'symbol_test',
+		'symbol_user',
+		'',
+		'postgres.example',
+		'15432') == result
+
+
+def test_testing_postgresql_fallback():
```
Jaguar0625: this test seems to be testing PostgresTestDatabase ?

explorer/rest/tests/test_symbol_core_routes.py
```diff
@@ -106,21 +305,162 @@ def test_setup_rejects_bad_node_url():
...
+
+		# Act:
+		with pytest.raises(SymbolNodeConfigurationError) as exception_info:
+			setup_symbol_facade(app)
+
+	# Assert:
+	assert str(exception_info.value) == 'Configured Symbol node host is not in SYMBOL_NODE_ALLOWED_HOSTS'
+
+
+def test_external_postgres_config():
```
Jaguar0625: this seems to be testing PostgresTestDatabase ?

Jaguar0625 (commented):
👍

explorer/puller/puller/db/SymbolDatabase.py
daoka: Fixed: changed to get_sync_state() is not None — this confirms the table exists and has been initialized, not just that the connection is open.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Fixed: split into test_create_tables_creates_block_table, test_create_tables_creates_enum_types, and test_create_tables_creates_indexes.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Fixed: now seeds 3 blocks (heights 1, 2, 3) and queries only [1, 2] — an all-returning implementation would include height 3 and fail the assertion.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes — the four importance block fields are nullable in the schema. The dedicated test verifies that non-NULL values survive the upsert round-trip, while normal block tests implicitly cover the NULL case.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: `repair_rollback_from_height` commits block deletion and sync_state update in a single transaction. The caller passes the sync_state it already holds in memory rather than re-reading inside the transaction, avoiding a read-modify-write race.

explorer/puller/tests/facade/symbol/test_Puller.py
daoka: Fixed: renamed `_create_db_config`, `_create_symbol_puller`, and `_temporary_symbol_puller` to public names (`create_db_config`, `create_symbol_puller`, `temporary_symbol_puller`).

explorer/puller/puller/facade/SymbolPuller.py
daoka: Yes — a single instance calls `get_symbol_node` with different paths each sync cycle: `/chain/info`, `/network/properties`, `/blocks/{height}` for hash verification, and `/blocks?pageSize=…&offset=…` for pagination. `_validate_symbol_node_path` runs on each call to guard against any caller-injected absolute URLs.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Fixed: added `connector` parameter to `SymbolPuller.__init__`. The `_set_symbol_connector` helper remains for `_SymbolPullerTestBase` subclasses where the connector needs to be replaced after the puller has been initialized with a live PostgreSQL database.

explorer/puller/puller/workflows/sync_symbol_block.py
daoka: Fixed: added `__enter__`/`__exit__` resource semantics to `SymbolPuller` (delegating to `symbol_db`) and changed the workflow to `with puller:`.

explorer/puller/tests/facade/symbol/test_Puller.py
daoka: Fixed: removed all `@patch('puller.facade.SymbolPuller.SymbolConnector')` decorators. Connectors are now injected via the constructor and tests pass `FakeConnector` or `ResponseConnector` directly.

explorer/puller/tests/facade/symbol/test_Puller.py
daoka: Fixed: added `_retry_delay` instance variable (default 2 s). Retry tests set `puller._retry_delay = 0` to skip the sleep without patching `asyncio.sleep`.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Fixed: added `assertEqual(['chain/info', 'network/properties', 'blocks/2', 'blocks/3'], connector.paths)` to prove that blocks 2 and 3 were actually fetched from the node for hash verification.

explorer/puller/tests/workflows/test_sync_symbol_block.py
daoka: Fixed: flipped to `assertEqual(args.max_height, 1000)` so the variable is always on the left.

explorer/puller/tests/workflows/test_sync_symbol_block.py
daoka: Removed: `test_main_guard_runs_main` was testing the `if __name__ == '__main__':` guard marked with `# pragma: no cover`. It has no real coverage value and has been deleted.

explorer/puller/tests/workflows/test_sync_symbol_block.py
daoka: Fixed: replaced the triple-patched mock test with a single `@patch('SymbolPuller')` that lets the real `parse_args` and `_create_node_config` run against `sys.argv` and `os.environ` patches.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: removed the `finalized_epoch`/`finalized_point` column-existence shim. Since this is pre-release and no production data exists, those columns are now required without a migration fallback.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: renamed to `try_get_sync_state()` to signal that the caller must check for a `None` return.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: added `SortOrder(str, Enum)` with `ASC` and `DESC` values.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: renamed to `local_height` to make clear it represents the Explorer's last synced height, not the node's chain height.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: `_calculate_height_range` now always returns either `(None, None)` or a range where `start_height <= end_height`. DESC returns `(max(1, start - limit + 1), start)` and ASC returns `(start, min(local_height, start + limit - 1))`.

explorer/rest/tests/db/test_SymbolDatabase.py
daoka: `SymbolDatabaseBlocksTest` tests pure validation paths (invalid sort, `fromHeight < 1`, cursor past head) that raise `ValueError` before any SQL executes. Running PostgreSQL for these pure-logic checks would add unnecessary overhead. `SymbolDatabasePostgresTest` covers the actual SQL execution paths.

explorer/rest/rest/db/SymbolDatabase.py
daoka: Fixed: swapped parameter order to `get_blocks(self, from_height, limit, sort)` — `from_height` first as the primary cursor, then `limit` as the page size.

explorer/rest/tests/model/symbol/test_Block.py
daoka: `state_hash_sub_cache_roots` is stored as JSONB and returned as a native Python list, unlike other fields that go through hex encoding. The test verifies that the empty-list edge case is handled correctly rather than becoming `None` or an empty string.

explorer/rest/tests/test_symbol_core_routes.py
daoka: `test_symbol_core_routes.py` contains end-to-end integration tests that spin up the full Flask app against a real PostgreSQL database, verifying the HTTP→DB stack. `tests/routes/symbol/test_routes.py` contains unit tests that mock `SymbolRestFacade` and focus on route logic in isolation (parameter parsing, error mapping, response shaping).

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Fixed: replaced tuple unpacking with `result[0]`–`result[3]` direct indexing, and other tests.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Agreed — and simplified the implementation to make this explicit. When `node_finalized_height > chain_height`, we now return `chain_height` directly as the finalized height via early return, removing the `min()` and the triple `None if is_finalization_capped else` expressions. This makes it clear at a glance that `chain_height` is the effective finalization ceiling for that sync run.

explorer/rest/tests/db/test_SymbolDatabase.py
daoka: Agreed — the split between non-PostgreSQL and PostgreSQL-backed tests for the same methods was confusing. Consolidated `SymbolDatabaseBlockTest` and `SymbolDatabasePostgresTest` into `SymbolDatabaseBlocksTest` so all `get_blocks` / `get_block` tests live in one place. Also renamed the helpers `_get_blocks_from_postgresql` → `_get_blocks` and `_get_block_from_postgresql` → `_get_block` since the PostgreSQL context is now implied by the class.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: Fixed: removed the `assertNotIn` in `test_reports_database_error_when_sync_state_read_fails` — the preceding `assertEqual` against the full health dict already makes it redundant. Applied the same cleanup to `test_reports_database_initialization_error` which had a similar `assertNotIn('connection refused', ...)` for the same reason.

explorer/rest/tests/model/symbol/test_Block.py
daoka: Fixed: added a `_expected_list_dict()` module-level helper and rewrote `test_can_convert_to_detail_dict` to use a single full-dict `assertEqual` with `**_expected_list_dict()` spread into the detail-only fields.

explorer/rest/rest/model/symbol/Block.py
daoka: Fixed: moved `_format_timestamp`, `_hex`, `_hex_or_none`, `_str_or_none`, `_format_total_fee`, and `XYM_DIVISIBILITY` out of `Block.py` into a new `rest/model/symbol/format.py` module so they can be reused by other Symbol model files.

explorer/rest/rest/routes/symbol.py
daoka: Fixed: extracted `_run_block_query(query_fn, error_log)` as an inner helper inside `setup_symbol_routes`. It centralises the `is_block_data_available()` check and `PsycopgError` handling; each route now only contains its own input validation and not-found logic.

explorer/rest/tests/test_app.py
daoka: Fixed: merged `_write_rest_config` and `_write_db_config` into a single `_write_config` function and updated all call sites.

explorer/rest/tests/test_symbol_core_routes.py
daoka: Fixed: replaced the individual health field assertions and `assertNotIn` with a single `create_symbol_health(errors=[...]) == health` full-dict comparison, which makes the `assertNotIn` redundant.

explorer/rest/tests/test_symbol_core_routes.py
daoka: Agreed — both tests verify `PostgresTestDatabase` infrastructure, not Symbol-specific logic. Moved `test_testing_postgresql_fallback`, `test_external_postgres_config`, and `FakePostgresql` to a new `explorer/rest/tests/test_postgres_utils.py`.

explorer/rest/tests/test_symbol_core_routes.py
daoka: Agreed — both tests verify `PostgresTestDatabase` infrastructure, not Symbol-specific logic. Moved `test_testing_postgresql_fallback`, `test_external_postgres_config`, and `FakePostgresql` to a new `explorer/rest/tests/test_postgres_utils.py`.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: Fixed: converted test_reports_unsynced_when_healthy_state_is_behind_chain_height, test_reports_unhealthy_rollback_state, and test_reports_repairing_sync_state_as_unhealthy from partial field assertions to full create_symbol_health() dict comparisons, consistent with the other health tests.

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: The editor is set to 140. That said, looking at the wrapped lines, the wrapping feels a bit aggressive. Will be more mindful of avoiding unnecessary line breaks going forward.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Fixed: removed the explicit pgcode assertions (22P02, 23514) and now just verify that PsycopgError is raised — the specific codes may vary across PostgreSQL versions. Also replaced the hardcoded block type value 32835 with BlockType.NEMESIS.value from the Symbol SDK so it tracks SDK changes automatically.

explorer/puller/puller/facade/SymbolPuller.py
daoka: `unhealthy` is effectively unrecoverable without operator intervention. It is set when a finalized block is missing from the local database or its hash does not match the node. On subsequent sync runs the same checks will re-trigger `SymbolRollbackError` until the operator manually recovers by dropping the database tables and restarting the puller for a full resync from height 1.

explorer/rest/rest/routes/symbol.py
daoka: 503 is intentional and appropriate here. `is_block_data_available()` returns `False` when the backend is temporarily unable to serve block data (e.g. status is `unhealthy` or `repairing`, or the DB connection failed). This is a service-availability problem, not a "resource not found" problem.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Parallelization is feasible here and I've implemented it.

Key design decisions:

- Added `BLOCK_PAGE_FETCH_CONCURRENCY = 10` and rewrote `_sync_block_pages` to
  fetch pages in batches of that size using `asyncio.gather`.
- Switched from per-page to batch-level `upsert_blocks`, reducing DB commits from
  ~56,000 to ~5,600 on a mainnet initial sync (~56,000 pages).
- **Watermark safety**: `upsert_sync_state` is called only once, at the end of
  `sync_block_headers`, so watermarks never advance mid-batch.
- **Deterministic restart**: `upsert_blocks` is idempotent, and `all_offsets` is
  computed as `range(start_height - 1, chain_height, BLOCK_PAGE_SIZE)`, which stops
  at `chain_height`. This guarantees the terminal short page is always the last
  element of `all_offsets` and its batch — no offset beyond the chain tip is ever
  requested.
- **Trade-off**: if a page fetch fails mid-batch after all retries, the entire
  batch must be re-fetched on restart (vs. the old per-page commit that preserved
  partial progress). This is acceptable given `_retry_operation` already handles
  transient errors with 3-attempt exponential backoff.

explorer/rest/rest/model/symbol/Block.py
daoka: Simplified `_format_timestamp` to `strftime('%Y-%m-%dT%H:%M:%SZ')`, dropping the dead string branch — psycopg2 always returns a `datetime` for `TIMESTAMP` columns, so the `str.replace(' ', 'T')` path was never reached in practice. The `test_can_format_string_timestamp` test was removed accordingly; `test_can_format_naive_datetime_as_utc` is kept to document that naive datetimes (the real psycopg2 output type) are handled correctly.

explorer/rest/rest/routes/symbol.py
daoka: Good catch. `get_block` was missing a `try-except PsycopgError` around the DB call, unlike `get_blocks` which already handles it. Added the guard so both methods handle DB errors consistently at the facade layer — a PsycopgError from `symbol_db.get_block()` now logs and returns `None` rather than propagating to `_run_block_query`.

explorer/rest/tests/db/test_SymbolDatabase.py
daoka: The redundant check at this line was removed as part of the test class consolidation
(the old `SymbolDatabaseBlocksTest` and `SymbolDatabasePostgresTest` were merged into
a single PostgreSQL-backed test class). The current `_query_symbol_database` helper
calls `drop_symbol_block_tables_if_present` directly — no extra guard needed since
it already uses `DROP TABLE IF EXISTS` / `DROP TYPE IF EXISTS` internally.

explorer/puller/tests/facade/symbol/test_Puller.py
```diff
@@ -0,0 +1,248 @@
...
+from unittest.mock import AsyncMock, MagicMock, patch
+
+from symbollightapi.model.Exceptions import NodeException
+
+from puller.facade.SymbolPuller import SymbolPuller
+
+from .puller_test_utils import (
+	NODE_URL,
+	ResponseConnector,
+	_set_retry_delay,
```
Jaguar0625: since imported, should rename to `set_retry_delay`

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def _set_retry_delay(puller, retry_delay):
+	# Keep retries fast in unit tests without exposing retry policy as public API.
+	puller._retry_delay = retry_delay  # pylint: disable=protected-access
```
Jaguar0625: is there any place in tests where you don't want to set this? if not, consider setting it to zero in create_symbol_puller

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -0,0 +1,224 @@
+import asyncio
+
+from puller.facade.SymbolPuller import SymbolRollbackError
+
+from .puller_test_utils import FakeConnector, _create_node_block, _create_sync_state, _set_symbol_connector, _SymbolPullerTestBase
```
Jaguar0625: remove '_' from imported functions and classes

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -0,0 +1,224 @@
...
+
+		# Assert:
+		sync_state = self.puller.symbol_db.get_sync_state()
+
+		self.assertEqual('unhealthy', sync_state['status'])
+
+	def test_sync_block_headers_marks_missing_finalized_block_unhealthy(self):
+		# Arrange:
+		connector = FakeConnector(3, {})
+		self.puller.symbol_db.create_tables()
```
Jaguar0625: i'd also suggest moving create_tables to setUp since nearly all tests require tables. for any that don't would explicitly drop them. this way the common case (tables) is less code than uncommon case.

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -0,0 +1,289 @@
...
+			connector.paths
+		)
+		self.assertIn(
+			'blocks?pageSize=100&offset=100&orderBy=height',
+			connector.paths
+		)
+		self.assertIn(
+			'blocks?pageSize=100&offset=200&orderBy=height',
+			connector.paths
+		)
```
Jaguar0625, line 125: would suggest asserting len(connector.paths) in this and following test

explorer/rest/rest/model/symbol/format.py
```diff
@@ -0,0 +1,23 @@
+from decimal import Decimal
+
+XYM_DIVISIBILITY = 6
+
+
+def _format_timestamp(timestamp):
```
Jaguar0625: please add corresponding tests for these functions are remove leading '_' since they are imported by Block.py

explorer/rest/rest/facade/SymbolRestFacade.py
```diff
@@ -61,11 +78,86 @@ def get_health(self):
...
+		"""Gets a Symbol block by height."""
+
+		if not self.is_block_data_available():
+			return None
+
+		try:
+			block = self.symbol_db.get_block(height)
+		except PsycopgError:
+			log.error('Failed to get Symbol block')
+			return None
```
Jaguar0625: do you need to expliclty handle PsycopgError error here just to log error and suppress? you already have handling at the caller's level (routes/symbol.py), so these seem to not add much value.

Jaguar0625 (commented):
👍

explorer/puller/tests/facade/symbol/test_Puller.py
daoka: Fixed: since `create_symbol_puller` now sets `_retry_delay = 0` directly (see below), all callers already get delay=0 without calling `_set_retry_delay`. With no remaining call sites, the function was removed entirely rather than just renamed.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Checked: all tests create their puller via `create_symbol_puller` or `temporary_symbol_puller` and all want delay=0. Added `puller._retry_delay = 0` directly in `create_symbol_puller`, making the four explicit `_set_retry_delay(puller, 0)` calls in the retry tests redundant. Removed those call sites and the helper function.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Fixed: renamed `_create_node_block` → `create_node_block`, `_create_sync_state` → `create_sync_state`, `_set_symbol_connector` → `set_symbol_connector`, and `_SymbolPullerTestBase`→ `SymbolPullerTestBase` in `puller_test_utils.py`. Updated all import sites.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Fixed: `moved create_tables()` into `SymbolPullerTestBase.setUp()`. Removed the eight explicit per-test calls from `test_PullerRollback.py`. No tests in the rollback suite need to skip table creation, so the common case is now the default.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Fixed: added `assertEqual(5, len(connector.paths))` to `test_sync_block_headers_fetches_multiple_pages_in_single_batch` (chain/info + network/properties + 3 block pages) and `assertEqual(13, len(connector.paths))` to `test_sync_block_headers_splits_into_multiple_batches` (chain/info + network/properties + 11 block pages).

explorer/rest/rest/model/symbol/format.py
daoka: Added `tests/model/symbol/test_format.py` with five tests covering `_format_timestamp`, `_hex`, `_hex_or_none`, `_str_or_none`, and `_format_total_fee`.

explorer/rest/rest/facade/SymbolRestFacade.py
daoka: Removed the `try-except PsycopgError` blocks from `get_blocks` and `get_block`. `PsycopgError` from the DB calls now propagates to `_run_block_query` in `routes/symbol.py`, which catches it, logs the error, and returns 503. `is_block_data_available()` retains its own guard since it is called directly from the route layer outside of `_run_block_query`.

explorer/puller/tests/facade/symbol/test_PullerSync.py
Jaguar0625: i would suggest adding comments like the above: `chain/info + network/properties + 3 block pages` so, the assert is clear for future reader of test

explorer/rest/tests/facade/test_SymbolRestFacade.py
```diff
@@ -4,6 +4,7 @@
 from common.tests.PostgresTestUtils import create_unreachable_db_configuration
```
Jaguar0625, line 4: instead of deleting the tests
test_get_blocks_returns_none_when_database_read_fails
test_get_block_returns_none_when_database_read_fails

you should be testing that that behavior raises Error

explorer/rest/tests/model/symbol/test_format.py
```diff
@@ -0,0 +1,32 @@
+from datetime import datetime, timezone
+from decimal import Decimal
+from unittest import TestCase
+
+from rest.model.symbol.format import _format_timestamp, _format_total_fee, _hex, _hex_or_none, _str_or_none
```
Jaguar0625: please remember to rename these to remove underscore

explorer/rest/tests/model/symbol/test_format.py
```diff
@@ -0,0 +1,32 @@
...
+		# Act + Assert:
+		self.assertEqual('0A0B', _hex_or_none(bytes.fromhex('0a0b')))
+		self.assertIsNone(_hex_or_none(None))
+
+	def test_can_format_optional_string(self):
+		# Act + Assert:
+		self.assertEqual('123', _str_or_none(123))
+		self.assertIsNone(_str_or_none(None))
+
+	def test_can_format_total_fee(self):
```
Jaguar0625: consider renaming: format_xym_amount ?

Jaguar0625 (commented):
👍

explorer/rest/tests/facade/test_SymbolRestFacade.py
daoka: Fixed. Replaced with test_get_blocks_raises_when_database_read_fails and test_get_block_raises_when_database_read_fails, which assert that PsycopgError propagates out of the facade.

explorer/rest/tests/model/symbol/test_format.py
daoka: Fixed.

explorer/rest/tests/model/symbol/test_format.py
daoka: Fixed. Renamed test_can_format_total_fee → test_can_format_xym_amount to match the production function name.

Jaguar0625 (approved):
👍

PR #2364, opened 2026-07-01
[explorer/rest] fix: refactor unit test, added helper method

AnthonyLaw:
Problem: I found it hard to read when using the index to view transaction information.
Solution: Added a helper, reading the hash instead of index


Wayonb (approved):
👍🏾 
Not sure if this is any easier for me to read since you switched to another set of indices 😅 
maybe having a name -> tx hash mapping might help

PR #2365, opened 2026-07-02
[explorer/frontend] fix: transaction type filter query param

AnthonyLaw:
Problem: Currently, the transaction type params are incorrect https://github.com/symbol/product/issues/2346. 
Solution: Handle the transaction type filter with the right parameter name


Wayonb (approved):
👍🏾

PR #2366, opened 2026-07-02
[explorer] fix: account and transaction detail pages return 404 instead of 500 for invalid identifiers

cryptoBeliever:
## Summary
Account and transaction detail pages returned a **500 Internal Server Error** for invalid identifiers, unlike block, mosaic, namespace, and node pages which already render **404**. This PR fixes both the REST API and the frontend so all detail pages behave consistently.

Issue description: https://github.com/symbol/product/issues/2345


AnthonyLaw (approved):
👍🏼

PR #2368, opened 2026-07-03
[wallet/symbol/mobile] feat: add multisig support to Harvesting screen

OlegMakarenko:
## Problem

The Send screen already supports sending transfer transactions from either the current account or a selected multisig account. The Harvesting screen, however, has no equivalent capability — it only allows starting/stopping harvesting transactions for the current account, with no way to select a multisig account as the sender. This inconsistency prevents users from activating or managing harvesting on their multisig accounts. Issue: https://github.com/symbol/product/issues/2201.

Additionally, the account selector used on the Send screen was a plain `Dropdown` that only displayed account names/addresses, with no balance information to help users choose the right sender.

## Solution

- Added a new, reusable `SelectTransactionSender` component that renders each account together with its balance, providing a clearer and more informative selection experience than a plain dropdown.
- Replaced the existing `Dropdown` on the Send screen with `SelectTransactionSender`, improving consistency and usability for account selection when sending transfers.
- Extended `HarvestingModule` to support fetching the harvesting summary and creating start/stop harvesting transactions for a specific account, enabling harvesting to be activated and managed via multisig accounts.
- Updated the Harvesting screen UI to include `SelectTransactionSender` for choosing the sender account, and to display the harvesting summary for the selected account — bringing the Harvesting screen's multisig support in line with the Send screen.


Wayonb (approved):
👍🏾

PR #2373, opened 2026-07-06
[Explorer] feat: transaction size and schema version.

AnthonyLaw:
Problem: The current transaction didn't have a size and schema version https://github.com/symbol/product/issues/2321
Solution: 
- Added size and version in `lightapi/python`
- Added size and version column in DB schema
- Added size and version in REST endpoint.


Wayonb (approved):
👍🏾

PR #2374, opened 2026-07-06
[explorer/frontend] fix: localize address book validation and profile modal

cryptoBeliever:
## Summary

- Fix address book validation in the profile modal: normalize input (remove separators, trim, uppercase) and verify a 40-character base32 address instead of only checking minimum length
- Store normalized addresses in local storage to avoid duplicate entries caused by formatting differences
- Replace hardcoded English strings in the profile modal (language, currency, address book, add/cancel/remove buttons, validation toasts) with i18n keys
- Add the required translation keys to `en`, `ja`, and `isv` locale files


explorer/frontend/components/Header.jsx
```diff
@@ -180,16 +182,16 @@ const Header = ({ backendStatus }) => {
...
 							<div>
-								<h4>Currency</h4>
+								<h4>{t('section_currency')}</h4>
 								<Dropdown options={currencies} value={userCurrency} onChange={setUserCurrency} />
 							</div>
 							<div>
-								<h4>Address Book</h4>
-								Give accounts names to easily identify them through the explorer.
+								<h4>{t('section_addressBook')}</h4>
+								{t('field_addressBook_description')}
```
OlegMakarenko: I'd call it `section_addressBook_description`, since keys started with `field_*` are reserved to be used with the `<Field />` component.

OlegMakarenko (changes requested):

explorer/frontend/components/Header.jsx
cryptoBeliever: You’re right. Changed 👍

OlegMakarenko (approved):

PR #2375, opened 2026-07-06
[explorer/frontend] Hide supernodes metric when SUPERNODE_API_URL is not configured

cryptoBeliever:
## Summary
- Skip the supernode statistics API call when `SUPERNODE_API_URL` is not set
- Hide the **Supernodes** metric on the home page when no supernode count is available
- Document that `SUPERNODE_API_URL` is optional

Closes #2314

## Problem

The testnet Explorer home page displayed a mainnet supernode count (e.g. **Supernodes: 77**) alongside testnet node data (e.g. **Total Nodes: 6**). Supernodes exist only on mainnet, so this mixed mainnet and testnet metrics and misrepresented the network status.

The frontend always called `SUPERNODE_API_URL/statistics`, which points to the mainnet supernode API regardless of the deployed network.

## Solution

Treat `SUPERNODE_API_URL` as optional:
- When configured → fetch supernode statistics and display the metric (mainnet behavior)
- When omitted or empty → do not call the supernode API, return `supernodes: null`, and hide the metric in the UI
For testnet deployments, omit `SUPERNODE_API_URL` from the environment configuration.

### DIFF
--- a/explorer/frontend/pages/index.jsx
+++ b/explorer/frontend/pages/index.jsx
@@ -119,7 +119,7 @@ const Home = ({
 					<div className="layout-grid-row layout-flex-fill">
 						<div className="layout-flex-col layout-flex-fill">
 							<Field title={t('field_totalNodes')}>{nodeStats.total}</Field>
-							{nodeStats.supernodes != null && (
+							{nodeStats.supernodes !== null && (
 								<Field title={t('field_supernodes')}>{nodeStats.supernodes}</Field>
 							)}
 						</div>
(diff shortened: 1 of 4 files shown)


explorer/frontend/pages/index.jsx
```diff
@@ -119,7 +119,9 @@ const Home = ({
 					<div className="layout-grid-row layout-flex-fill">
 						<div className="layout-flex-col layout-flex-fill">
 							<Field title={t('field_totalNodes')}>{nodeStats.total}</Field>
-							<Field title={t('field_supernodes')}>{nodeStats.supernodes}</Field>
+							{nodeStats.supernodes != null && (
```
OlegMakarenko: `!=` -> `!==`

OlegMakarenko (changes requested):

explorer/frontend/pages/index.jsx
cryptoBeliever: Thank you, changed :+1:

OlegMakarenko (approved):

PR #2376, opened 2026-07-06
[explorer/frontend] fix: show awaiting cosignatures for pending multisig transactions

cryptoBeliever:
## Summary
- Fix misleading `~ in X seconds` estimate shown for unconfirmed multisig transactions in the Pending Transactions section on the home page
- Show `Awaiting cosignatures` for unconfirmed `MULTISIG` transactions instead of the block time estimate
- Keep the confirmation time estimate for other unconfirmed transaction types (e.g. transfers)

Closes #2342


explorer/frontend/variants/symbol/utils/transactions.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Checks whether a transaction is waiting for cosignatures before it can confirm. A Symbol
+ * aggregate bonded transaction stays in the partial group until the required cosignatures are
+ * collected, then moves to the unconfirmed group.
+ * @param {{ type: string, group: string }} transaction - the transaction to evaluate.
+ * @returns {boolean} true when the transaction is awaiting cosignatures.
+ */
+export const isTransactionAwaitingSignatures = transaction => transaction.group === 'partial';
```
OlegMakarenko: Same here, but `SymbolTransactionGroup` in the variant scoped constants `variants/symbol/constants/`.

explorer/frontend/variants/nem/utils/transactions.js
```diff
@@ -1,7 +1,17 @@
+import { TRANSACTION_TYPE } from '@/app/constants';
 import { decodeTransactionMessage } from '@/app/utils/common';
 
 const NEM_PLAIN_MESSAGE_TYPE = 1;
```
OlegMakarenko: Could you please create a `MessageType` object in the root`frontend/constants/` and define `Plain = 'plain`, `Encrypted = 'encrypted`, `Raw = 'raw'`. Then you can reuse it across the project instead of defining inline strings.

Or in this cases, it can be numbers `Plain = 1`, etc.

OlegMakarenko (changes requested):
Looks good, just a few minor comments.

explorer/frontend/constants/index.js
```diff
@@ -8,6 +8,18 @@ export const TRANSACTION_TYPE = {
 	MULTISIG: 'MULTISIG'
 };
 
+export const MESSAGE_TYPE = {
+	PLAIN: 'plain',
+	ENCRYPTED: 'encrypted',
+	RAW: 'raw'
+};
+
+export const TRANSACTION_GROUP = {
```
OlegMakarenko, line 17: I think there can be 3 options:
**The `TRANSACTION_GROUP` is used in common components or pages:**
1. We keep the `TRANSACTION_GROUP` with all 3 values here and do not have dedicated Symbol and NEM variants. All are consuming this single object.
2. We keep the `TRANSACTION_GROUP` with only generic 2  values 'confirmed' and 'unconfirmed' which is used in common components/pages. Also Symbol and NEM implement their own `SYMBOL_TRANSACTION_GROUP` and `NEM_TRANSACTION_GROUP` - which are used only in `/variants/*`.

**The `TRANSACTION_GROUP` is not used in common parts, but only in Symbol and NEM variants:**
3. Keep only Symbol and NEM specific `SYMBOL_TRANSACTION_GROUP` and `NEM_TRANSACTION_GROUP`. Remove `TRANSACTION_GROUP` from the root constants.

I would go with 2 or 3, since the root constants should not know about Symbol and NEM specifics, we keep it generic as possible. In case this requires significant code changes we can keep option 1 temporary and update it in later PR. What do you think?

cryptoBeliever: Thanks! I went with option 2 - it turned out to be a simple change (if I understood it correctly) since the variant-specific SYMBOL_TRANSACTION_GROUP and NEM_TRANSACTION_GROUP were already in place and used throughout /variants/*.

What I changed:
- Removed PARTIAL from the root TRANSACTION_GROUP, so it now only holds the generic confirmed/unconfirmed values used by common components/pages.
- SYMBOL_TRANSACTION_GROUP now defines PARTIAL: 'partial' directly (the generic values are still reused from the root object to keep them consistent with what common pages consume).
- RecentTransactions.jsx was the only common component referencing PARTIAL - inverted its condition to group !== TRANSACTION_GROUP.CONFIRMED, so the common layer no longer needs to know about Symbol-specific groups and treats any non-confirmed transaction as pending.

OlegMakarenko (approved):

PR #2379, opened 2026-07-07
[explorer/puller] refactor: extract Symbol block model and format helpers

daoka:
## Summary

- This is the prerequisite groundwork for the receipts and
 transactions puller branches, both of which build on this
  `Block.py`/`format.py` structure
- Extract block row construction out of `SymbolDatabase.py`/`SymbolPuller.py`
  into a new `puller/model/symbol/Block.py` (`create_block_row`), and shared
  optional-value formatters into `puller/model/symbol/format.py`, mirroring
  the existing REST-side `format.py` pattern
- Narrow `symbol_blocks.size`/`fee_multiplier`/`voting_eligible_accounts_count`
  from `bigint` to `int` — Symbol network limits make it impossible for these
  values to approach the int4 range, and this also fixes the inconsistency
  where `harvesting_eligible_accounts_count` was already `int`


explorer/puller/puller/model/symbol/Block.py
```diff
@@ -0,0 +1,62 @@
...
+	32835: 'nemesis',
+	33347: 'importance',
+	33091: 'normal'
+}
+BLOCK_TYPE_VALUES = tuple(BLOCK_TYPE_LABELS.values())
+
+
+def _block_type_label(block_type):
+	try:
+		return BLOCK_TYPE_LABELS[int(block_type)]
```
Jaguar0625: use `in` to check inclusion instead of relying on exception

Jaguar0625 (approved):
👍

explorer/puller/puller/model/symbol/Block.py
daoka: Updated to check membership within instead of relying on the KeyError.

PR #2380, opened 2026-07-07
[explorer/puller] feat: sync Symbol receipts and compute block reward

daoka:
## Summary

- Add a `symbol_receipts` table and `puller/model/symbol/Receipt.py` to　persist receipts from Symbol Node's `/statements/transaction` endpoint.　Each receipt is classified into one of four groups　(`balanceChange`/`balanceTransfer`/`artifactExpiry`/`inflation`) based on its Symbol receipt type, with group-specific fields populated (mosaic/amount, sender/recipient/target address, or artifact id)
- Add `symbol_blocks.block_reward`, computed as the sum of a height's inflation-receipt amounts and written back once that height's receipts are synced
- Sync receipts as part of the existing block-sync loop: after each batch of blocks is persisted, fetch and group the corresponding receipt statements by height. Statement pages are fully accumulated before grouping, so a height whose statements span two pages isn't split
- Extend rollback to delete `symbol_receipts` rows before `symbol_blocks` (the foreign key has no cascading delete, so the deletion order matters)
- `receipt_type`/`receipt_group` are Postgres enum columns rather than freeform text/int; `amount`/`block_reward` are integer columns, not `numeric`, since Symbol raw-unit amounts fit comfortably in a 64-bit integer without the overhead of arbitrary-precision arithmetic
- Consolidate the page-size constant used for both block and statement pagination into a single `MAX_PAGE_SIZE`, and reuse a shared optional-value formatting module instead of duplicating small string/bytes conversion helpers

## Note on base branch

This PR targets `explorer/backend/symbol-block-model-refactor` instead of `dev` because it depends on that branch's block-row/format-helper extraction (`puller/model/symbol/Block.py`, `puller/model/symbol/format.py`), which hasn't merged to `dev` yet (see #2379 ). Once that PR merges, GitHub will offer to retarget this PR's base to `dev` automatically.

### DIFF
--- a/explorer/puller/puller/model/symbol/Receipt.py
+++ b/explorer/puller/puller/model/symbol/Receipt.py
@@ -41,6 +41,7 @@ RECEIPT_TYPE_GROUPS = {
 	**{receipt_type.value: 'artifactExpiry' for receipt_type in ARTIFACT_EXPIRY_RECEIPT_TYPES},
 	ReceiptType.INFLATION.value: 'inflation'
 }
+assert set(RECEIPT_TYPE_LABELS) == set(RECEIPT_TYPE_GROUPS), 'RECEIPT_TYPE_LABELS and RECEIPT_TYPE_GROUPS must cover the same receipt types'
 INFLATION_RECEIPT_TYPE = RECEIPT_TYPE_LABELS[ReceiptType.INFLATION.value]
 
 
--- a/explorer/puller/puller/model/symbol/Receipt.py
+++ b/explorer/puller/puller/model/symbol/Receipt.py
@@ -1,5 +1,7 @@
 from symbolchain.sc import ReceiptType
 
+from puller.model.symbol.format import bytes_from_hex_or_none, str_or_none
+
 BALANCE_CHANGE_RECEIPT_TYPES = (
 	ReceiptType.HARVEST_FEE,
 	ReceiptType.LOCK_HASH_CREATED,
@@ -42,14 +44,6 @@ RECEIPT_TYPE_GROUPS = {
 INFLATION_RECEIPT_TYPE = RECEIPT_TYPE_LABELS[ReceiptType.INFLATION.value]
 
 
-def _address_bytes_or_none(value):
-	return bytes.fromhex(value) if value else None
-
-
-def _str_or_none(value):
-	return str(value) if value is not None else None
-
-
 def _create_base_receipt_row(statement, source, receipt):
 	receipt_type = int(receipt['type'])
 
@@ -71,7 +65,7 @@ def _create_base_receipt_row(statement, source, receipt):
 
 
 def _add_mosaic_fields(row, receipt):
-	row['mosaic_id'] = _str_or_none(receipt.get('mosaicId'))
+	row['mosaic_id'] = str_or_none(receipt.get('mosaicId'))
 	row['amount'] = int(receipt.get('amount', 0))
 
 
@@ -89,15 +83,15 @@ def create_receipt_rows(statement_item):
 		row = _create_base_receipt_row(statement, source, receipt)
 		if 'balanceChange' == row['receipt_group']:
 			_add_mosaic_fields(row, receipt)
-			row['target_address'] = _address_bytes_or_none(receipt.get('targetAddress'))
+			row['target_address'] = bytes_from_hex_or_none(receipt.get('targetAddress'))
 		elif 'balanceTransfer' == row['receipt_group']:
 			_add_mosaic_fields(row, receipt)
-			row['sender_address'] = _address_bytes_or_none(receipt.get('senderAddress'))
-			row['recipient_address'] = _address_bytes_or_none(receipt.get('recipientAddress'))
+			row['sender_address'] = bytes_from_hex_or_none(receipt.get('senderAddress'))
+			row['recipient_address'] = bytes_from_hex_or_none(receipt.get('recipientAddress'))
 		elif 'inflation' == row['receipt_group']:
 			_add_mosaic_fields(row, receipt)
 		elif 'artifactExpiry' == row['receipt_group']:
-			row['artifact_id'] = _str_or_none(receipt.get('artifactId'))
+			row['artifact_id'] = str_or_none(receipt.get('artifactId'))
 
 		rows.append(row)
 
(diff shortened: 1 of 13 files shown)


explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -752,6 +926,82 @@ def test_rejects_duplicate_block_hash_for_different_heights(self):
 		with self.assertRaises(PsycopgError):
 			database.upsert_blocks([_create_block(2, block_hash=b'same hash')])
 
+	def test_rejects_receipt_without_existing_block(self):
```
Jaguar0625, line 929: in this case, i believe it will insert the receipt and then fail setting reward in blocks table?

explorer/puller/puller/model/symbol/Receipt.py
```diff
@@ -0,0 +1,99 @@
...
+def create_receipt_rows(statement_item):
+	"""Creates persisted Symbol receipt rows from one transaction statement item."""
+
+	statement = statement_item['statement']
+	source = statement['source']
+	rows = []
+	for receipt in statement['receipts']:
+		receipt_type = int(receipt['type'])
+		if receipt_type not in RECEIPT_TYPE_GROUPS:
+			continue
```
Jaguar0625, line 82: are there specific types you don't expect to handle?

explorer/puller/tests/model/symbol/test_Receipt.py
```diff
@@ -0,0 +1,235 @@
...
+				'type': ReceiptType.MOSAIC_EXPIRED.value,
+				'artifactId': 'ABCDEF0123456789'
+			}
+		])
+
+		# Act:
+		rows = create_receipt_rows(statement_item)
+
+		# Assert:
+		self.assertEqual([
```
Jaguar0625, line 229: would also assert they are using height from statement

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -336,6 +364,34 @@ def _fetch_importance_block_fields(database, height):
 
 		return cursor.fetchone()
 
+	@staticmethod
+	def _fetch_block_reward(database, height):
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT block_reward FROM symbol_blocks WHERE height = %s', (height,))
+
+		result = cursor.fetchone()
+		return result[0] if result and result[0] is not None else None
```
Jaguar0625: isn't this redundant? `and result[0] is not None`

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -61,13 +63,36 @@ def test_sync_block_headers_repairs_shallow_unfinalized_rollback(self):
...
+					]
+				}
+			}
 		)
 		self._seed_blocks(
 			self.puller.symbol_db,
 			[1, 2, 3],
 			{2: b'local mismatch'.hex()}
 		)
+		self.puller.symbol_db.upsert_receipts_for_height(2, [{
```
Jaguar0625, line 83: you are testing this entry is removed as part of repair (by asserting different amount) correct?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -147,8 +157,10 @@ def test_sync_block_headers_splits_into_multiple_batches(self):
 		# Assert:
 		block_paths = [path for path in connector.paths if path.startswith('blocks?')]
 		transaction_paths = [path for path in connector.paths if path.startswith('transactions/confirmed?')]
+		statement_paths = [path for path in connector.paths if path.startswith('statements/transaction?')]
 		self.assertEqual(11, len(block_paths))
 		self.assertEqual(2, len(transaction_paths))
+		self.assertEqual([statement_path(1, 1000), statement_path(1001, chain_height)], statement_paths)
```
Jaguar0625, line 163: for consistency, i'd either test all len or all list contents. only checking contents for one is odd.

Jaguar0625 (commented):
👍

explorer/puller/tests/model/symbol/test_Receipt.py
daoka: Added height to the same assertion alongside the source ids, since it's populated from the statement the same way.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: It was redundant — when result[0] is already None, returning it directly produces the same result as the explicit else None. Simplified to result[0] if result else None.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Correct — the full-list assertion on receipts after repair only passes if the stale amount=999 row is gone and replaced by amount=222. Added a comment on the seed data making that explicit instead of leaving it implicit.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Agreed, that was inconsistent with the other two. Changed transaction_paths to assert full contents ([transaction_path(1, 201)]) to match block_paths/statement_paths.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Checked this empirically — the receipt insert fails first, not the block_reward update. symbol_receipts.height has a REFERENCES symbol_blocks(height) FK, so inserting a receipt at a height with no block raises immediately:


ForeignKeyViolation: insert or update on table "symbol_receipts" violates foreign key constraint "symbol_receipts_height_fkey"
DETAIL:  Key (height)=(7) is not present in table "symbol_blocks".
The UPDATE symbol_blocks SET block_reward = ... line is never reached.

explorer/puller/puller/model/symbol/Receipt.py
daoka: Yes, three ReceiptType values are intentionally excluded:

* `ADDRESS_ALIAS_RESOLUTION` / `MOSAIC_ALIAS_RESOLUTION`: these belong to `address_resolution_statements` / `mosaic_resolution_statements` — separate arrays from `TransactionStatement.receipts[]` (what `/statements/transaction` returns). They'd need their own endpoints/schema — out of scope here.
* `TRANSACTION_GROUP`: confirmed against the node source (`TransactionStatement.cpp`) — it's only used as a hash domain-separation tag, never passed to `addReceipt()`. It structurally can't appear in `receipts[]`.

explorer/puller/tests/db/test_SymbolDatabase.py
Jaguar0625: you're right - i missed the foreign key 👍

Jaguar0625 (approved):
👍

PR #2381, opened 2026-07-07
[explorer/puller] feat: sync Symbol confirmed transactions and persist mosaic/address rows

daoka:
## Summary

* Add `symbol_transactions` / `symbol_transaction_mosaics` / `symbol_transaction_addresses` tables and `puller/model/symbol/Transaction.py` to persist confirmed transactions from Symbol Node's `/transactions/confirmed` endpoint, covering both top-level transactions identified by `hash` and embedded transactions inside aggregates identified by `aggregate_hash` + `embedded_index`.
* For each transaction type, extract the mosaics and addresses it involves into dedicated child tables so they're searchable without parsing the raw payload. For example, a transfer produces a signer/recipient address pair and a transfer-role mosaic row per mosaic sent; an aggregate produces a cosignatory address row per cosigner.　These extractions, plus `recipient_address` / `target_address` on the transaction row itself, exist specifically because account/mosaic history search needs to find a transaction by an address or mosaic id it never explicitly names as "the" signer. A plain column or join target is required for that to be indexable.
  Every other type-specific field, such as `restrictionAdditions` for an account address restriction or `newRestrictionValue` for a mosaic global restriction, has no such search requirement. It only needs to round-trip for detail-view rendering. Rather than adding a one-off column per field per transaction type, those fields are kept in a `body` JSON column: every field of the raw transaction payload except the ones already promoted to their own column, namely signer public key, max fee, deadline, size, type, and message.
* Compute `effective_fee` from the block's fee multiplier using `LEAST(max_fee, size * fee_multiplier)`, and assign a monotonic `list_sequence` to top-level transactions so the default transaction list can paginate without a SQL `OFFSET`.
* Sync transactions as part of the existing block-sync loop, and extend rollback to remove transaction/mosaic/address rows before blocks.
* `type_name`, `message_type`, and mosaic/address `role` are Postgres enum columns rather than freeform text. `size` is a plain integer column, not `bigint`, since transaction sizes fit comfortably within that range.
* Consolidated a few small duplicated helpers, including type-label lookup, public-key-to-address derivation, timestamp conversion, and a repeated rollback delete sequence, into shared functions. Also aligned the page-size constant naming with the sibling receipts branch to avoid an unnecessary future merge conflict between the two.

## ~~Note on base branch~~

~~This PR targets `explorer/backend/symbol-block-model-refactor` instead of `dev` because it depends on that branch's block-row/format-helper extraction (`puller/model/symbol/Block.py`, `puller/model/symbol/format.py`), which hasn't merged to `dev` yet (see #2379 ). Once that PR merges, GitHub will offer to retarget this PR's base to `dev` automatically.~~


explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -289,6 +297,223 @@ def test_create_tables_creates_symbol_block_columns_and_constraints(self):
...
+		], columns)
+		self.assertEqual([
+			('FOREIGN KEY', 'height'),
+			('PRIMARY KEY', 'id'),
+			('UNIQUE', 'aggregate_hash'),
+			('UNIQUE', 'embedded_index'),
+			('UNIQUE', 'hash')
+		], key_constraints)
+
+	def test_create_tables_creates_symbol_transaction_child_columns_and_constraints(self):
```
Jaguar0625: do you mean child **tables**, not columns?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -554,6 +971,12 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
 		])
+		database.upsert_transactions_for_height(2, [create_transaction_entry(
+			2,
+			'rollbacked',
+			mosaic_rows=[{'mosaic_id': '1111111111111111', 'amount': 10, 'role': 'transfer', 'position': 0}],
+			address_rows=[{'address': b'rollbacked address', 'role': 'signer'}]
+		)])
 
 		# Act:
 		database.repair_rollback_from_height(2, _create_sync_state(
```
Jaguar0625, line 1015: consider adding some additional txes at or prior to repair height, so db is not completely empty after

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -148,7 +150,7 @@ def test_sync_block_headers_splits_into_multiple_batches(self):
 		block_heights, sync_state = self._sync_with_connector(connector)
 
 		# Assert:
-		self.assertEqual(13, len(connector.paths))  # chain/info + network/properties + 11 block pages
+		self.assertEqual(15, len(connector.paths))  # chain/info + network/properties + 11 block pages + 2 transaction pages
```
Jaguar0625: why 2 pages here? 🤔

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -0,0 +1,187 @@
...
+		# Arrange:
+		connector = FakeConnector(1, {}, transactions_by_path={
+			transaction_path(1, 1): {
+				'data': [create_node_transaction(1)]
+			}
+		})
+		set_symbol_connector(self.puller, connector)
+
+		# Act:
+		rows_by_height = asyncio.run(self.puller._get_transaction_rows_by_height(1, 1, 100))  # pylint: disable=protected-access
```
Jaguar0625, line 44: can you not test this behavior using public function(s) ?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -0,0 +1,187 @@
...
+		set_symbol_connector(self.puller, connector)
+		block_rows = [{'height': 1}, {'height': 2}, {'height': 3}]
+
+		# Act:
+		asyncio.run(self.puller._sync_transactions_for_batch(block_rows, 100))  # pylint: disable=protected-access
+
+		# Assert:
+		self.assertEqual([
+			(1, [bytes.fromhex(f'{1:064X}')]),
+			(2, []),
```
Jaguar0625, line 143: why did you choose behavior to write empty rows? are you relying on it to simplify compound queries?

explorer/puller/puller/model/symbol/format.py
```diff
@@ -6,3 +11,31 @@ def int_or_none(value):
...
+
+	parts = name.lower().split('_')
+	return parts[0] + ''.join(part.capitalize() for part in parts[1:])
+
+
+def label_for_type(labels, type_value, entity_name):
+	"""Looks up a label for a Symbol type code, raising ValueError when unsupported."""
+
+	try:
+		return labels[int(type_value)]
```
Jaguar0625: please use in for checking existence

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -0,0 +1,241 @@
...
+	unique_rows = []
+	for row in rows:
+		key = (row['address'], row['role'])
+		if key in seen:
+			continue
+
+		seen.add(key)
+		unique_rows.append(row)
+
+	return unique_rows
```
Jaguar0625: i think you can simplify this with list comprehension ?

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -0,0 +1,241 @@
...
+		'signer_address': signer_address,
+		'recipient_address': bytes.fromhex(transaction['recipientAddress']) if 'recipientAddress' in transaction else None,
+		'target_address': _target_address(transaction_type, transaction),
+		'message_type': message_type,
+		'message_payload': message_payload,
+		'body': {
+			key: value
+			for key, value in transaction.items()
+			if key not in BODY_EXCLUDED_FIELDS
+		},
```
Jaguar0625, line 235: can you explain why you are excluding these specific fields? it is because they are promoted to higher level (table columns) vs object.

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -0,0 +1,241 @@
...
+		'raw_payload': item,
```
Jaguar0625, line 236: i am not sure of reason for both body and raw_payload, isn't body always a subset of raw_payload?

explorer/puller/tests/model/symbol/test_Transaction.py
```diff
@@ -0,0 +1,581 @@
...
+			}],
+			'address_rows': [
+				{'address': bytes.fromhex(SIGNER_ADDRESS), 'role': 'signer'},
+				{'address': bytes.fromhex(RECIPIENT_ADDRESS), 'role': 'recipient'}
+			]
+		}, row)
+
+	def test_sdk_enum_values_build_transaction_labels(self):
+		transactions = {
+			TransactionType.TRANSFER.value: _create_transaction(),
```
Jaguar0625: would prefer you use a three-tuple included expected label explicitly for asserting

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -0,0 +1,241 @@
...
+	return _unique_address_rows(rows)
+
+
+def _parse_message(transaction):
+	message = transaction.get('message')
+	if not message:
+		return None, None
+
+	message_type = MESSAGE_TYPE_LABELS.get(int(message[:2], 16))
+	return message_type, message[2:]
```
Jaguar0625: i am not sure if this is wanted since symbol (unlike nem) doesn't require message_type prefixed byte

explorer/puller/tests/model/symbol/test_Transaction.py
```diff
@@ -0,0 +1,581 @@
...
+	def test_create_transaction_address_rows_deduplicates_repeated_role_addresses(self):
+		self._assert_address_rows(TransactionType.ACCOUNT_ADDRESS_RESTRICTION.value, {
+			'restrictionAdditions': [TARGET_ADDRESS],
+			'restrictionDeletions': [TARGET_ADDRESS]
+		}, [
+			{'address': bytes.fromhex(SIGNER_ADDRESS), 'role': 'signer'},
+			{'address': bytes.fromhex(TARGET_ADDRESS), 'role': 'target'}
+		])
+
+	def test_create_transaction_row_leaves_scalar_target_address_empty_for_list_target_type(self):
```
Jaguar0625, line 473: can you explain what you're testing in this test?

explorer/puller/tests/model/symbol/test_Transaction.py
```diff
@@ -0,0 +1,581 @@
...
+		])
+
+	def test_create_transaction_address_rows_creates_multisig_cosignatories(self):
+		self._assert_address_rows(TransactionType.MULTISIG_ACCOUNT_MODIFICATION.value, {
+			'addressAdditions': [TARGET_ADDRESS],
+			'addressDeletions': [ALIAS_ADDRESS]
+		}, [
+			{'address': bytes.fromhex(SIGNER_ADDRESS), 'role': 'signer'},
+			{'address': bytes.fromhex(TARGET_ADDRESS), 'role': 'cosignatory'},
+			{'address': bytes.fromhex(ALIAS_ADDRESS), 'role': 'cosignatory'}
```
Jaguar0625, line 524: i am not sure if same role makes sense for both cosigner add/delete and when cosigner is being used as signer? @cryptoBeliever what is your opinion?

Jaguar0625 (commented):
👍

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -0,0 +1,241 @@
...
+		'role': 'signer'
+	}]
+
+	if transaction_type in (TransactionType.TRANSFER.value, TransactionType.SECRET_LOCK.value, TransactionType.SECRET_PROOF.value):
+		rows.append({
+			'address': _bytes_from_transaction_field(transaction, 'recipientAddress'),
+			'role': 'recipient'
+		})
+
+	if transaction_type in (
```
Jaguar0625, line 109: one more general comment about structure of this file intended more as informational. feel free to incorporate or not.

the way you split up the file, adding support for a transaction requires changes in multiple places (e.g. for extracting addresses and mosaics) in separate functions and additional branches. 

normally, it's a good practice to minimize the number of branches (more for readability and maintenance than performance nowadays given hardware advancements). an alternative design would be to have a small class / tuple with lambdas for each transaction type. instances could live in a dict and there would be one for each transaction that would be retrieved and used for normalization.

for example, something roughly like this
```py
TransactionHandler = namedtuple('TransactionHandler', ['extract_address', 'extract_mosaics'])

def create_transfer_handler():
        def extract_address(self, transaction):
            return [{
			'address': _bytes_from_transaction_field(transaction, 'recipientAddress'),
			'role': 'recipient'
            }]
        def extract_mosaics(self, transaction):
    		return [
		    	{
			    	'mosaic_id': mosaic['id'],
		    		'amount': int(mosaic['amount']),
		    		'role': 'transfer',
		    		'position': position
	    		}
		    	for position, mosaic in enumerate(transaction.get('mosaics', []))
		]

    return TransactionHandler(extract_address, extract_mosaics)

TRANSACTION_HANDLERS = {
    TRANSFER.value(): create_transfer_handler()
    # ...
}
```
again, you don't have to change to this since what you have is fine, but i just wanted to call out a different architecture.

explorer/puller/puller/model/symbol/format.py
daoka: Updated to check membership with in instead of relying on the KeyError

explorer/puller/puller/model/symbol/Transaction.py
daoka: Simplified using a dict comprehension keyed by (address, role), then .values() to get the deduplicated list back.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Right, it's the child tables (`symbol_transaction_mosaics/symbol_transaction_addresses`), not columns of some parent table. Renamed to `test_create_tables_creates_symbol_transaction_child_table_columns_and_constraints`.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Added a transaction at height 1 (before the repair height of 2) and switched the assertions from bare row counts to actual row content (height/hash/mosaic/address), so the test now verifies that pre-repair-height data survives the rollback instead of just showing the database isn't in some broken state after an all-empty repair.

explorer/puller/tests/model/symbol/test_Transaction.py
daoka: The old assertion compared row['type_name'] against TRANSACTION_TYPE_LABELS[transaction_type] itself, so a bug in the label table wouldn't have been caught by this test at all. Switched to a three-tuple with the expected label string hardcoded per case, so it's an independent check.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Switched this to drive through the public `sync_block_headers()` (via the existing `_assert_sync_rejects_node_response` helper) instead of calling _get_transaction_rows_by_height directly.

That actually surfaced a real bug: blocks were being upserted before transactions were fetched, so a malformed transaction page response would leave already-committed blocks with no matching transactions — a partial-write state. Fixed by fetching transaction rows first in a new _sync_block_batch, then writing blocks and transactions together only once the fetch succeeds. That also made the old `_sync_transactions_for_batch` unreachable in production, so I removed it and retargeted its two unit tests at `_sync_block_batch`.

explorer/puller/puller/model/symbol/Transaction.py
daoka: These fields aren't an arbitrary exclusion list — they're the ones common to every transaction type that either need to be queried/indexed directly, or need dedicated parsing logic rather than being stored verbatim.

signerPublicKey and type back the signer_public_key/signer_address and type/type_name columns, which have their own indexes (idx_symbol_transactions_signer_height, idx_symbol_transactions_signer_address, idx_symbol_transactions_type_height) so we can filter directly in SQL instead of scanning JSON. deadline and maxFee get their own columns (deadline/network_deadline, max_fee) for the same reason list views need them: every top-level transaction has them, and they're needed to render any transaction row, not just one type's detail view. size follows maxFee for the same reason (both only apply at the top level; embedded transactions get NULL for these). message is a bit different from the rest — it needs actual parsing (the marker-byte handling from the other thread) rather than just being copied verbatim, so we decode it once into message_type/message_payload instead of making every reader re-parse it out of body.

Everything else stays in body because it's type-specific — only some transaction types have it (e.g. restrictionAdditions for ACCOUNT_ADDRESS_RESTRICTION, id/duration for MOSAIC_DEFINITION), so giving each its own column would mean a wide table that's mostly NULL depending on the row's type, and none of them need to be filtered on directly — they're only read back when rendering that specific transaction's detail view.

explorer/puller/puller/model/symbol/Transaction.py
daoka: Yes, body is a subset of raw_payload — that's intentional. Transactions have dozens of types with genuinely different fields, so a JSON escape hatch is needed for the type-specific remainder.
We persist body instead of deriving it from raw_payload at read time so the "which fields are already columns" knowledge (BODY_EXCLUDED_FIELDS) lives only in the puller. Otherwise REST would need its own copy of that exclusion list, and the two could drift whenever a field gets promoted to a column later.
raw_payload stays for full-fidelity audit/reprocessing; body is the pre-filtered projection REST is meant to consume directly.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: BLOCK_PAGE_FETCH_CONCURRENCY is 10, and this test has 11 block pages, so the sync loop runs 2 batches (10 pages, then 1 short page). The transaction fetch runs once per batch, not once per block page, so 2 batches → 2 transaction-page fetches.
Rather than just explain the 15, I split the assertion so it's self-evident: block_paths/transaction_paths are now asserted separately (11 and 2), instead of relying on a combined total with a decoding comment. Same fix applied to the other batching test above it.

explorer/puller/tests/model/symbol/test_Transaction.py
daoka: target_address is a scalar column only populated for transaction types with a single target (ACCOUNT_METADATA/MOSAIC_METADATA/NAMESPACE_METADATA/MOSAIC_ADDRESS_RESTRICTION). ACCOUNT_ADDRESS_RESTRICTION has a list of targets (restrictionAdditions/restrictionDeletions), which go into symbol_transaction_addresses with role='target' instead — so this test asserts the scalar column is intentionally left NULL for that type, to make sure we don't accidentally pick one arbitrary address out of the list for it.

explorer/puller/puller/model/symbol/Transaction.py
daoka: You're right that Symbol doesn't enforce a message-type prefix at the protocol level — message is opaque bytes_array in the catbuffer schema, unlike NEM's real MessageType field.

Correction on 0x00/plain though: it's an official value in the TypeScript SDK's MessageType enum (PlainMessage = 0x00, alongside RawMessage = -1 for unmarked content) — not just wallet folklore. The Python SDK we use just doesn't expose that enum (MessageEncoder only handles encryption).

That encoder's actual markers exposed two bugs here: persistentDelegationHarvesting matched on 1 byte instead of the full 8-byte DELEGATION_MARKER, and unrecognized markers had their first byte unconditionally stripped, corrupting unprefixed messages.

Fixed both — full marker match for delegation, and unrecognized/unmarked messages are now left untouched instead of truncated.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Not for compound queries — the empty-list call to upsert_transactions_for_height doesn't actually insert any placeholder row (the loop over transaction_entries just runs zero times). Its only effect is running the DELETE statements for that height, which matters if a height is re-synced after previously having transactions (e.g. after a rollback repairs a shorter chain). Renamed the test and added test_upsert_transactions_for_height_clears_existing_rows_when_replaced_with_empty_list at the DB layer to make that concrete.

explorer/puller/puller/model/symbol/Transaction.py
daoka: Thanks for the pointer — makes sense as the codebase grows, but I'll leave the current per-field-branch structure for this phase and revisit if it gets unwieldy.

explorer/puller/tests/model/symbol/test_Transaction.py
Jaguar0625: maybe rename the test or add comment so this is clearer

Jaguar0625 (approved):
👍

explorer/puller/tests/model/symbol/test_Transaction.py
daoka: Added a comment explaining the intent directly on the test — thanks for the suggestion.

PR #2382, opened 2026-07-07
[lightapi] feat: Keep NEM transfer message semantics in lightapi by exposing the original message type

cryptoBeliever:
## Summary
- Keep NEM transfer message semantics in lightapi by exposing the original message type in Message objects instead of a derived boolean.
- Rename the transfer message field from is_plain to type in lightapi mapping (1 = plain, 2 = encrypted).
- Update lightapi tests to validate message type mapping for plain and encrypted transfer messages.
- Use `MessageType.PLAIN` instead of forwarding the raw NEM message type integer (`1` / `2`).

## Motivation
A boolean is_plain loses message-type information and makes downstream handling brittle if new NEM message types appear in the future. Keeping the raw type in lightapi preserves protocol semantics and centralizes interpretation in Explorer/UI logic while maintaining a clear migration path for already persisted data.

Issue: https://github.com/symbol/product/issues/2352

### DIFF
--- a/lightapi/python/tests/model/test_Transaction.py
+++ b/lightapi/python/tests/model/test_Transaction.py
@@ -420,19 +420,19 @@ class TransactionHandlerTest(unittest.TestCase):
 		}
 		return TransactionHandler().map[TransactionType.TRANSFER.value](tx_json)
 
-	def test_maps_plain_message_type_to_is_plain_true(self):
+	def test_maps_plain_message_type(self):
 		# Arrange + Act:
 		mapped_args = self._map_transfer({'payload': '48656C6C6F', 'type': MessageType.PLAIN.value})
 
 		# Assert:
-		self.assertEqual(Message('48656C6C6F', True), mapped_args['message'])
+		self.assertEqual(Message('48656C6C6F', MessageType.PLAIN.value), mapped_args['message'])
 
-	def test_maps_encrypted_message_type_to_is_plain_false(self):
+	def test_maps_encrypted_message_type(self):
 		# Arrange + Act:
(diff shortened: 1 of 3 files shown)


lightapi/python/tests/model/test_Transaction.py
```diff
@@ -399,3 +400,35 @@ def test_return_unknown_transaction_type(self):
 		# Arrange + Act:
 		with self.assertRaises(UnknownTransactionType):
 			TransactionFactory.create_transaction(123, COMMON_ARGS, {})
+
+
+class TransactionHandlerTest(unittest.TestCase):
```
AnthonyLaw, line 413: Maybe you can add these tests under `TransactionFactoryTest` ?

It currently has `test_create_transfer_transaction`. You could update it to test a transaction with a plain message, then add

`test_create_transfer_transaction_encrypted_message`

What do you think?

cryptoBeliever: Thanks for the suggestion. I think this specific change is better tested in TransactionHandlerTest, because the message.type -> message.is_plain mapping happens in TransactionHandler.
TransactionFactory receives already mapped specific_args (including Message), so testing this in TransactionFactoryTest would not validate the new mapping logic itself. Do you agree?

AnthonyLaw: sound good 👍🏼

AnthonyLaw (approved):
👍🏼

Wayonb (changes requested):
The fix looks fine 👍🏾 
Is there a reason not to fix this on the Explorer side?
I think api is correct and should just return the information from the node. 🤔

Wayonb (approved):
👍🏾

PR #2384, opened 2026-07-08
[lightapi/python] feat: Added transaction size and version field

AnthonyLaw:
Problem: The transaction information is missing the size and version fields. https://github.com/symbol/product/issues/2321
Solution:  Handle calculating transaction size and extract schema version information.


lightapi/python/symbollightapi/connector/NemConnector.py
```diff
@@ -384,12 +387,19 @@ def _map_to_transaction(transaction, block_height):
 			'timestamp': tx_json['timeStamp'],
 			'deadline': tx_json['deadline'],
 			'signature': tx_json['signature'],
+			'size': size,
+			'schema_version': schema_version
```
Wayonb: Maybe just ``schema_version -> version`` to keep consistent.  Not sure if there is a reason to rename it? 🤔

Wayonb (changes requested):

lightapi/python/symbollightapi/connector/NemConnector.py
AnthonyLaw: I'm bad at giving a field name :D

Wayonb (approved):
👍🏾

PR #2385, opened 2026-07-08
[explorer/puller] fix: process account updates when a transferred mosaic has levy receipient.

AnthonyLaw:
Problem: Currently, the levy recipient is missed when mosaics are transferred. https://github.com/symbol/product/issues/2354
Solution: Added logic to check the levy recipient of each transferred mosaic and extract the recipient’s address for account processing.

### DIFF
--- a/explorer/puller/tests/db/test_NemDatabase.py
+++ b/explorer/puller/tests/db/test_NemDatabase.py
@@ -592,6 +592,41 @@ class NemDatabaseTest(unittest.TestCase):  # pylint: disable=too-many-public-met
 			'TALICE6XEEEOBFJVY3ZCENZ7WBG6LB4KB7P7KMQX'
 		])
 
+	def test_can_get_mosaic_levy_recipients(self):
+		# Arrange:
+		levy_recipient = Address('TBRYCNWZINEVNITUESKUMFIENWKYCRUGNE63PMQQ')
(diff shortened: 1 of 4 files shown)


explorer/puller/tests/db/test_NemDatabase.py
```diff
@@ -591,6 +591,41 @@ def test_can_get_accounts_for_refresh(self):
...
+		mosaics = [
+			MOSAICS[0]._replace(
+				root_namespace='namespace',
+				namespace_name='namespace.test',
+				levy_type=1,
+				levy_namespace_name='nem.xem',
+				levy_fee=500,
+				levy_recipient=levy_recipient
+			),
+			MOSAICS[0]._replace(
```
Wayonb, line 607: should this be MOSAICS[1]?

Wayonb (changes requested):

explorer/puller/tests/db/test_NemDatabase.py
AnthonyLaw: I intended to use `MOSAICS[0]` here as a base fixture and override the fields needed for this case with `_replace(...)`. 

So this creates new test data derived from the sample, rather than reusing the original mosaic.

Wayonb (approved):
👍🏾

PR #2389, opened 2026-07-09
[explorer/puller] fix: widen symbol_blocks.total_fee to bigint to avoid overflow

daoka:
## Problem

`symbol_blocks.total_fee` was defined as `int` (PostgreSQL `integer`, max ~2.1B). Live-node validation against testnet (`https://401-sai-dual.symboltest.net:3001`) found a real block (height 6224) with `totalFee = 13199992608`, which exceeds that range and made `upsert_blocks` fail with `psycopg2.errors.NumericValueOutOfRange`.

`total_fee` sums fees across every transaction in a block, so unlike per-transaction fee fields it can plausibly exceed `int4` on a block with many/large-fee transactions.

## Solution

Widen `total_fee` to `bigint`, matching the existing `difficulty`/`total_voting_balance` columns.

## Test plan

- [x] `test_create_tables_creates_symbol_block_columns_and_constraints` updated to expect `int8`
- [x] New regression test `test_can_upsert_block_when_total_fee_exceeds_int4`, using the real observed value (height 6224, `13199992608`)
- [x] `cd explorer/puller && PYTHONPATH=.:.. python -m pytest tests/ -v` — 231 passed
- [x] `bash scripts/ci/lint.sh` — 10.00/10
- [x] 100% patch coverage on changed lines

### DIFF
--- a/explorer/puller/puller/db/SymbolDatabase.py
+++ b/explorer/puller/puller/db/SymbolDatabase.py
@@ -43,7 +43,7 @@ SYMBOL_BLOCK_DEFINITIONS = [
 	'previous_hash bytea NOT NULL',
 	'timestamp timestamp NOT NULL',
 	'network_timestamp bigint NOT NULL',
-	'total_fee int NOT NULL',
+	'total_fee bigint NOT NULL',
 	'transactions_count int NOT NULL',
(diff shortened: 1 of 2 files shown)


Jaguar0625 (approved):
👍

PR #2390, opened 2026-07-10
[explorer/puller] feat: add Symbol accounts core current-state persistence

daoka:
## Summary

* Add `symbol_accounts` / `symbol_account_mosaics` / `symbol_multisig` tables and `puller/model/symbol/Account.py` to persist current-state account data (current balances, importance, supplemental keys, multisig) needed by account detail and current-state enrichment.
* Introduce a dirty-key current-state account refresh: after each block-sync batch, collect the set of addresses touched by that batch and refresh their current state via a batched `POST /accounts` (chunked at 100 addresses) plus a per-address `GET /account/{address}/multisig` (no batch endpoint exists for multisig). "Touched" covers three sources:
  * Block beneficiaries.
  * Every transaction participant address (signer, recipient, target, cosignatory, mosaic owner, etc.), excluding namespace-alias pseudo-addresses (`Address.is_alias()`), which aren't resolvable accounts.
  * Receipt addresses: `balanceChange` receipts' `target_address` and `balanceTransfer` receipts' `sender_address`/`recipient_address`. This is needed because some receipts change an account's balance with no correlated transaction in that block — `LOCK_HASH_EXPIRED`/`LOCK_SECRET_EXPIRED` fire automatically at a protocol-timeout height with no transaction at all, and rental-fee (`MOSAIC_RENTAL_FEE`/`NAMESPACE_RENTAL_FEE`) recipient addresses aren't signers of the triggering transaction. `HARVEST_FEE`'s target always equals the block beneficiary, so it's already covered by dedup with no special-casing.
* Like the existing block/transaction/receipt sync, all of this batch's network fetches (transactions, receipts, dirty accounts) complete before any DB writes begin, so a mid-batch failure leaves zero partial writes.
  * `is_harvesting_active`: an address that was a block's beneficiary gets `true` when that block's timestamp is within the last 7 days; dirty-key refresh never overwrites this back to `false` (no reliable negative signal from a single batch). When an address is both a beneficiary and a transaction participant in the same batch, the beneficiary result takes priority.

## Known limitation (out of scope for this PR)

When a transaction's address field (e.g. `recipientAddress`) encodes a namespace alias instead of a real address, dirty-key refresh currently excludes that alias pseudo-address entirely rather than resolving it to the real account via `/statements/resolutions/address`. This means an account reachable only through an alias reference in a block's transactions doesn't get a dirty-key refresh. Resolving this correctly requires per-height (not range) resolution-statement fetches and tracking each transaction's block-relative source id — including embedded transactions, which need their parent aggregate's index. It also affects `symbol_transaction_addresses` (Backend4, already merged), which persists the same unresolved alias bytes for participant rows. Both call sites should be fixed together in a future phase; deferring both is intentional, not an oversight.

### DIFF
--- a/explorer/puller/puller/facade/SymbolPuller.py
+++ b/explorer/puller/puller/facade/SymbolPuller.py
@@ -307,25 +307,32 @@ class SymbolPuller:
 				last_synced_block_hash = last_row['hash']
 
 				if len(blocks) < MAX_PAGE_SIZE:
-					await self._sync_block_batch(batch_rows, epoch_adjustment_seconds)
-					await self._refresh_dirty_accounts_for_batch(batch_rows, native_mosaic_id, native_mosaic_divisibility)
+					await self._sync_block_batch(batch_rows, epoch_adjustment_seconds, native_mosaic_id, native_mosaic_divisibility)
 					return last_synced_height, last_synced_block_hash
 
-			await self._sync_block_batch(batch_rows, epoch_adjustment_seconds)
-			await self._refresh_dirty_accounts_for_batch(batch_rows, native_mosaic_id, native_mosaic_divisibility)
+			await self._sync_block_batch(batch_rows, epoch_adjustment_seconds, native_mosaic_id, native_mosaic_divisibility)
 
 		return last_synced_height, last_synced_block_hash
 
-	async def _sync_block_batch(self, batch_rows, epoch_adjustment_seconds):
+	async def _sync_block_batch(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+		self,
+		batch_rows,
+		epoch_adjustment_seconds,
+		native_mosaic_id,
+		native_mosaic_divisibility
+	):
 		transaction_rows_by_height = await self._get_transaction_rows_by_height(
 			batch_rows[0]['height'],
 			batch_rows[-1]['height'],
 			epoch_adjustment_seconds
 		)
 		receipt_rows_by_height = await self._get_receipt_rows_by_height(batch_rows[0]['height'], batch_rows[-1]['height'])
+		dirty_account_rows = await self._fetch_dirty_accounts_for_batch(batch_rows, native_mosaic_id, native_mosaic_divisibility)
+
 		self.symbol_db.upsert_blocks(batch_rows)
 		self._upsert_transactions_for_batch(batch_rows, transaction_rows_by_height)
 		self._upsert_receipts_for_batch(batch_rows, receipt_rows_by_height)
+		self._write_dirty_accounts_for_batch(dirty_account_rows)
 
 	def _upsert_transactions_for_batch(self, block_rows, rows_by_height):
 		for row in block_rows:
@@ -413,8 +420,8 @@ class SymbolPuller:
 
 		return self._native_mosaic_info
 
-	async def _refresh_dirty_accounts_for_batch(self, block_rows, native_mosaic_id, native_mosaic_divisibility):
-		"""Refreshes beneficiary current-state account rows touched by a synced block batch."""
+	async def _fetch_dirty_accounts_for_batch(self, block_rows, native_mosaic_id, native_mosaic_divisibility):
+		"""Fetches beneficiary current-state account rows touched by a synced block batch."""
 
 		latest_block_by_beneficiary = {}
 		for row in block_rows:
@@ -423,15 +430,31 @@ class SymbolPuller:
 				latest_block_by_beneficiary[row['beneficiary_address']] = row
 
 		observed_height = max(row['height'] for row in block_rows)
+		dirty_account_rows = []
 		for address, block_row in sorted(latest_block_by_beneficiary.items()):
-			await self._refresh_account_current_state(
+			dirty_account_rows.append(await self._fetch_account_current_state(
 				address,
 				observed_height=observed_height,
 				native_mosaic_id=native_mosaic_id,
 				native_mosaic_divisibility=native_mosaic_divisibility,
-				harvested_block_timestamp=block_row['timestamp'])
+				harvested_block_timestamp=block_row['timestamp']))
 
-	async def _refresh_account_current_state(  # pylint: disable=too-many-arguments,too-many-positional-arguments
+		return dirty_account_rows
+
+	def _write_dirty_accounts_for_batch(self, dirty_account_rows):
+		"""Writes fetched dirty account current-state rows after block rows are persisted."""
+
+		for dirty_account_row in dirty_account_rows:
+			self.symbol_db.upsert_account_current_state(
+				dirty_account_row['account_row'],
+				dirty_account_row['mosaic_rows'],
+				overwrite_importance_percentage=False,
+				overwrite_is_harvesting_active=dirty_account_row['overwrite_is_harvesting_active'])
+			self.symbol_db.upsert_multisig(
+				dirty_account_row['address'],
+				dirty_account_row['multisig_row'])
+
+	async def _fetch_account_current_state(  # pylint: disable=too-many-arguments,too-many-positional-arguments
 		self,
 		address,
 		observed_height,
@@ -439,7 +462,7 @@ class SymbolPuller:
 		native_mosaic_divisibility,
 		harvested_block_timestamp
 	):
-		"""Refreshes one Symbol account current-state row and its multisig row from the configured node."""
+		"""Fetches one Symbol account current-state row and its multisig row from the configured node."""
 
 		address_text = str(self.symbol_facade.network.address_class(address))
 		item = await self.get_symbol_node(f'/accounts/{address_text}')
@@ -455,14 +478,16 @@ class SymbolPuller:
 		if overwrite_is_harvesting_active:
 			account_row['is_harvesting_active'] = True
 
-		self.symbol_db.upsert_account_current_state(
-			account_row,
-			mosaic_rows,
-			overwrite_importance_percentage=False,
-			overwrite_is_harvesting_active=overwrite_is_harvesting_active)
-		self.symbol_db.upsert_multisig(
-			address,
-			None if _is_not_found_response(multisig_response) else create_multisig_row(address, multisig_response['multisig'], observed_height))
+		return {
+			'address': address,
+			'account_row': account_row,
+			'mosaic_rows': mosaic_rows,
+			'overwrite_is_harvesting_active': overwrite_is_harvesting_active,
+			'multisig_row': None if _is_not_found_response(multisig_response) else create_multisig_row(
+				address,
+				multisig_response['multisig'],
+				observed_height)
+		}
 
 	@staticmethod
 	def _is_harvested_block_within_active_window(harvested_block_timestamp):
--- a/explorer/puller/puller/facade/SymbolPuller.py
+++ b/explorer/puller/puller/facade/SymbolPuller.py
@@ -1,6 +1,7 @@
 import asyncio
 import configparser
 from collections import defaultdict, namedtuple
+from datetime import datetime, timedelta, timezone
 from urllib.parse import urlparse
 
 from common.symbol.NodeConfiguration import SymbolNodeConfiguration
@@ -12,6 +13,7 @@ from zenlog import log
 
 from puller.db.SymbolDatabase import SymbolDatabase
 from puller.facade.RequestRateLimiter import RequestRateLimiter
+from puller.model.symbol.Account import HARVESTING_ACTIVE_WINDOW_DAYS, create_account_row, create_multisig_row
 from puller.model.symbol.Block import create_block_row
 from puller.model.symbol.Receipt import INFLATION_RECEIPT_TYPE, create_receipt_rows
 from puller.model.symbol.Transaction import create_transaction_row
@@ -35,9 +37,16 @@ class SymbolRollbackError(RuntimeError):
 	"""Raised when Symbol rollback repair is outside the safe Backend2 scope."""
 
 
-def _raise_if_node_error(response):
+def _raise_if_node_error(response, allow_not_found=False):
 	if isinstance(response, dict) and 'code' in response and 'message' in response:
+		if allow_not_found and 'ResourceNotFound' == response.get('code'):
+			return
+
 		raise NodeException(f'{response["code"]}: {response["message"]}')
+
+
+def _is_not_found_response(response):
+	return isinstance(response, dict) and 'ResourceNotFound' == response.get('code')
 
 
 class SymbolPuller:
@@ -70,6 +79,7 @@ class SymbolPuller:
 		self.symbol_facade = SymbolFacade(network)
 		self._retry_delay = 2
 		self._rate_limiter = rate_limiter or RequestRateLimiter(max_requests_per_second)
+		self._native_mosaic_info = None
 
 	def __enter__(self):
 		self.symbol_db.__enter__()
@@ -90,14 +100,14 @@ class SymbolPuller:
 
 		return normalized_path
 
-	async def _retry_operation(self, operation, description, retries=3):
+	async def _retry_operation(self, operation, description, retries=3, not_found_as_error=True):
 		"""Retries a Symbol node operation with exponential backoff."""
 
 		for attempt_index in range(retries):
 			try:
 				await self._rate_limiter.wait_for_turn()
 				response = await operation()
-				_raise_if_node_error(response)
+				_raise_if_node_error(response, allow_not_found=not not_found_as_error)
 				return response
 			except NodeException as error:
 				attempt = attempt_index + 1
@@ -115,7 +125,8 @@ class SymbolPuller:
 		normalized_path = self._validate_symbol_node_path(url_path)
 		return await self._retry_operation(
 			lambda: self._symbol_connector.get(normalized_path, property_name, not_found_as_error),
-			f'fetching Symbol node path {normalized_path}'
+			f'fetching Symbol node path {normalized_path}',
+			not_found_as_error=not_found_as_error
 		)
 
 	async def post_symbol_node(self, url_path, request_payload, property_name=None, not_found_as_error=True):
@@ -124,10 +135,11 @@ class SymbolPuller:
 		normalized_path = self._validate_symbol_node_path(url_path)
 		return await self._retry_operation(
 			lambda: self._symbol_connector.post(normalized_path, request_payload, property_name, not_found_as_error),
-			f'posting Symbol node path {normalized_path}'
+			f'posting Symbol node path {normalized_path}',
+			not_found_as_error=not_found_as_error
 		)
 
-	async def sync_block_headers(self, max_height=None):
+	async def sync_block_headers(self, max_height=None):  # pylint: disable=too-many-locals
 		"""Synchronizes Symbol block headers from the configured node."""
 
 		chain_info = await self.get_symbol_node('/chain/info')
@@ -137,6 +149,7 @@ class SymbolPuller:
 			self._get_finalized_watermark(chain_info, chain_height)
 		)
 		epoch_adjustment_seconds = self._parse_epoch_adjustment(network_properties)
+		native_mosaic_id, native_mosaic_divisibility = await self._get_native_mosaic_info(network_properties)
 		sync_state = self._get_bounded_sync_state(self.symbol_db.get_sync_state(), chain_height)
 		if is_finalization_capped and sync_state and sync_state['last_synced_height'] >= finalized_height:
 			finalized_hash = self.symbol_db.get_block_hash(finalized_height)
@@ -145,7 +158,12 @@ class SymbolPuller:
 		if not start_height:
 			start_height = (sync_state['last_synced_height'] + 1) if sync_state and sync_state['last_synced_height'] else 1
 
-		last_synced_height, last_synced_block_hash = await self._sync_block_pages(start_height, chain_height, epoch_adjustment_seconds)
+		last_synced_height, last_synced_block_hash = await self._sync_block_pages(
+			start_height,
+			chain_height,
+			epoch_adjustment_seconds,
+			native_mosaic_id,
+			native_mosaic_divisibility)
 		if last_synced_height is None and sync_state:
 			last_synced_height = sync_state['last_synced_height']
 			last_synced_block_hash = sync_state['last_synced_block_hash']
@@ -250,7 +268,14 @@ class SymbolPuller:
 		})
 		return height
 
-	async def _sync_block_pages(self, start_height, chain_height, epoch_adjustment_seconds):
+	async def _sync_block_pages(  # pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals
+		self,
+		start_height,
+		chain_height,
+		epoch_adjustment_seconds,
+		native_mosaic_id,
+		native_mosaic_divisibility
+	):
 		last_synced_height = None
 		last_synced_block_hash = None
 		all_offsets = range(start_height - 1, chain_height, MAX_PAGE_SIZE)
@@ -283,9 +308,11 @@ class SymbolPuller:
 
 				if len(blocks) < MAX_PAGE_SIZE:
 					await self._sync_block_batch(batch_rows, epoch_adjustment_seconds)
+					await self._refresh_dirty_accounts_for_batch(batch_rows, native_mosaic_id, native_mosaic_divisibility)
 					return last_synced_height, last_synced_block_hash
 
 			await self._sync_block_batch(batch_rows, epoch_adjustment_seconds)
+			await self._refresh_dirty_accounts_for_batch(batch_rows, native_mosaic_id, native_mosaic_divisibility)
 
 		return last_synced_height, last_synced_block_hash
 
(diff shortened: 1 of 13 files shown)


explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -817,6 +927,231 @@ def test_rejects_non_singleton_sync_state_row(self):
...
+		self.assertEqual((
+			multisig_row['min_approval'],
+			multisig_row['min_removal'],
+			multisig_row['cosignatory_addresses'],
+			multisig_row['multisig_addresses']
+		), (
+			result[0],
+			result[1],
+			[bytes(value) for value in result[2]],
+			[bytes(value) for value in result[3]]
```
Jaguar0625: do you want to check updated_at_height too ?

explorer/puller/tests/model/symbol/test_Account.py
```diff
@@ -0,0 +1,203 @@
...
+	def test_create_account_row_rejects_balance_one_raw_unit_below_lower_boundary(self):
+		self._assert_harvesting_eligibility(HARVESTING_ELIGIBLE_MIN_NATIVE_BALANCE * 10 ** DIVISIBILITY - 1, False)
+
+	def test_create_account_row_rejects_upper_boundary_balance(self):
+		self._assert_harvesting_eligibility(HARVESTING_ELIGIBLE_MAX_NATIVE_BALANCE * 10 ** DIVISIBILITY, False)
+
+	def test_create_account_row_marks_balance_one_raw_unit_below_upper_boundary_as_eligible(self):
+		self._assert_harvesting_eligibility(HARVESTING_ELIGIBLE_MAX_NATIVE_BALANCE * 10 ** DIVISIBILITY - 1, True)
+
+	def test_create_account_row_rejects_zero_balance(self):
```
Jaguar0625: you seem to be testing that zero balance is not eligible for harvesting, not that it is rejected. perhaps rename test ?

Jaguar0625: also following test is using "rejects"

explorer/puller/tests/model/symbol/test_Account.py
```diff
@@ -0,0 +1,203 @@
...
+	account.update(account_overrides)
+
+	return {'account': account, 'id': 'search-id'}
+
+
+def _plain_value(value):
+	return value.adapted if isinstance(value, Json) else value
+
+
+class AccountTest(TestCase):
```
Jaguar0625, line 44: consider adding test for account without any mosaics

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -134,6 +147,7 @@ async def sync_block_headers(self, max_height=None):
 			self._get_finalized_watermark(chain_info, chain_height)
 		)
 		epoch_adjustment_seconds = self._parse_epoch_adjustment(network_properties)
+		native_mosaic_id, native_mosaic_divisibility = await self._get_native_mosaic_info(network_properties)
```
Jaguar0625: consider returning named tuple since you seem are passing these together everywhere

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
 		raw_epoch_adjustment = str(raw_epoch_adjustment)
 		return int(raw_epoch_adjustment[:-1] if raw_epoch_adjustment.endswith('s') else raw_epoch_adjustment)
 
+	async def _get_native_mosaic_info(self, network_properties=None):
```
Jaguar0625: does not seem to be a reason to have network_properties optional

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
...
+			dirty_addresses[address] = {
+				'is_beneficiary': True,
+				'harvested_block_timestamp': block_row['timestamp']
+			}
+
+		for transaction_rows in transaction_rows_by_height.values():
+			for transaction_row in transaction_rows:
+				for address_row in transaction_row['address_rows']:
+					address = address_row['address']
+					if address not in dirty_addresses and not Address(address).is_alias():
```
Jaguar0625: are you planning on handling aliases eventually?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -279,20 +305,44 @@ async def _sync_block_pages(self, start_height, chain_height, epoch_adjustment_s
...
 			batch_rows[0]['height'],
 			batch_rows[-1]['height'],
 			epoch_adjustment_seconds
 		)
 		receipt_rows_by_height = await self._get_receipt_rows_by_height(batch_rows[0]['height'], batch_rows[-1]['height'])
+		dirty_addresses = self._collect_dirty_addresses_for_batch(
+			batch_rows,
+			transaction_rows_by_height,
+			receipt_rows_by_height)
+		observed_height = max(row['height'] for row in batch_rows)
```
Jaguar0625, line 336: if you have two blocks: H, H+1 
account A is modified in H and 
account B is modified in H+1 

this will cause both blocks to have same observed height of H+1, which seems wrong? consider including modified height in dirty_addresses

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
...
+		native_mosaic_id,
+		native_mosaic_divisibility
+	):
+		"""Fetches current-state account and multisig rows touched by a synced block batch."""
+
+		sorted_addresses = sorted(dirty_addresses.keys())
+		address_text_by_address = {
+			address: str(self.symbol_facade.network.address_class(address))
+			for address in sorted_addresses
+		}
```
Jaguar0625: why do you need to sort the addresses?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
...
+
+		dirty_account_rows = []
+		for address in sorted_addresses:
+			address_text = address_text_by_address[address]
+			if address not in account_items_by_address:
+				raise ValueError(f'Missing Symbol accounts batch item for address {address_text}')
+
+			item = account_items_by_address[address]
+			multisig_response = await self.get_symbol_node(f'/account/{address_text}/multisig', not_found_as_error=False)
+			account_row, mosaic_rows = create_account_row(
```
Jaguar0625: is there reason you don't want to batch multisig lookups?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
...
+				item,
+				self.symbol_facade.network,
+				observed_height,
+				native_mosaic_id,
+				native_mosaic_divisibility)
+
+			dirty_info = dirty_addresses[address]
+			overwrite_is_harvesting_active = dirty_info['is_beneficiary'] and self._is_harvested_block_within_active_window(
+				dirty_info['harvested_block_timestamp'])
```
Jaguar0625: how are you handling case harvested block was found in previous sync (still within active window) but not this sync ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -370,6 +420,146 @@ def _parse_epoch_adjustment(network_properties):
...
+		return dirty_account_rows
+
+	def _write_dirty_accounts_for_batch(self, dirty_account_rows):
+		"""Writes fetched dirty account current-state rows after block rows are persisted."""
+
+		for dirty_account_row in dirty_account_rows:
+			self.symbol_db.upsert_account_current_state(
+				dirty_account_row['account_row'],
+				dirty_account_row['mosaic_rows'],
+				overwrite_importance_percentage=False,
```
Jaguar0625: where do you call with `overwrite_importance_percentage` True?

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -256,33 +262,82 @@ def create_chain_info(chain_height=1, finalized_height=1):
...
+		'importance': '100',
+		'importanceHeight': '6'
+	}
+	account.update(account_overrides)
+
+	return {'account': account, 'id': item_id}
 
 
-class FakeConnector:
+class PostPath(str):
```
Jaguar0625: i would be very cautious subclassing built in types. why don't you want to convert paths to contain str or tuple(url_path, request_payload) ?

explorer/puller/tests/facade/symbol/test_Puller.py
```diff
@@ -204,6 +204,27 @@ def test_get_symbol_node_retries_api_error_response(self):
 		)
 		self.assertEqual(2, connector.get.await_count)
 
+	def test_get_symbol_node_retries_non_not_found_api_error_when_not_found_is_allowed(self):
```
Jaguar0625, line 254: should you have matching test to show how ResourceNotFound is not retried

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -0,0 +1,449 @@
...
+
+		# Act:
+		connector = self._sync_blocks(blocks, create_account_item(importance='321'))
+
+		# Assert:
+		self.assertEqual(1, connector.paths.count('accounts'))
+		self.assertEqual([{'addresses': [address_text]}], connector.post_payloads)
+		self.assertEqual((321, 0, True, True, 2), self._fetch_account_current_state())
+
+	def test_refresh_dirty_accounts_for_batch_updates_transaction_participant_without_overwriting_harvesting_active(self):
```
Jaguar0625: is this really testing not overwrite? it seems to be testing that participant's are not set as active harvesters

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -0,0 +1,449 @@
...
+			},
+			account_by_address=self._account_by_address_text(BENEFICIARY_ADDRESS))
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual([{'addresses': [beneficiary_address_text]}], connector.post_payloads)
+
+	def test_refresh_dirty_accounts_for_batch_prefers_beneficiary_timestamp_when_address_is_also_transaction_participant(self):
```
Jaguar0625, line 141: how are you testing this? shouldn't beneficiary also be a participant/recipient in this case?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -0,0 +1,449 @@
...
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(True, self._fetch_account_current_state()[2])
+
+	def test_refresh_dirty_accounts_for_batch_deduplicates_repeated_transaction_participant_addresses(self):
+		# Arrange:
+		participant_address_text = self._address_text(BENEFICIARY_ADDRESS)
+		beneficiary_address = '980101010101010101010101010101010101010101010101'
```
Jaguar0625: test is correct, but any reason you didn't use BENEFICIARY_ADDRESS and RECIPIENT_TEST ?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -0,0 +1,449 @@
...
+	def test_get_native_mosaic_info_is_memoized(self):
+		# Arrange:
+		connector = FakeConnector(1, {0: [self._create_block(1)]})
+
+		# Act:
+		self._sync_with_connector(connector)
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(2, connector.paths.count('network/properties'))
```
Jaguar0625: somewhat unrelated, but would also cache network/properties, so this is only called once

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -392,6 +468,9 @@ def _fetch_receipts(database):
 
 		return cursor.fetchall()
 
+	def _beneficiary_address_text(self):
+		return str(self.puller.symbol_facade.network.address_class(bytes.fromhex(BENEFICIARY_ADDRESS)))
```
Jaguar0625: str(Address.from_decoded_address_hex_string(x))

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -333,24 +367,19 @@ def test_get_receipt_rows_by_height_groups_rows_after_reading_all_pages(self):
 		self.assertEqual([10], [row['amount'] for row in rows_by_height[1]])
 		self.assertEqual([20, 30], [row['amount'] for row in rows_by_height[2]])
 
-	def test_sync_block_batch_upserts_empty_heights_and_queries_batch_range(self):
+	def test_sync_block_batch_upserts_empty_heights_and_writes_previously_fetched_receipts(self):
```
Jaguar0625, line 370: why did you change this test?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -151,20 +148,29 @@ def test_sync_block_batch_calls_upsert_for_every_height_even_without_transaction
 			for height, transaction_rows in transaction_database.calls
 		])
 
-	def test_sync_block_batch_queries_exact_batch_range(self):
+	def test_sync_block_batch_writes_previously_fetched_transactions_for_exact_batch_rows(self):
```
Jaguar0625, line 151: why did you change this test?

Jaguar0625 (commented):
👍

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Thanks, added.

explorer/puller/tests/model/symbol/test_Account.py
daoka: Renamed.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Added.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Removed the subclass — paths now stays plain strings for both GET and POST, with a separate post_payloads_list tracked in parallel.

explorer/puller/tests/facade/symbol/test_Puller.py
daoka: Added.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Seeded True instead of False so it actually proves preservation rather than a coincidental False→False, and renamed to does_not_mark_transaction_participant_as_active_harvester.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Confirmed a real bug — the transaction's recipient defaulted to a different address, so the "also a participant" scenario was never exercised. Fixed the transaction to target BENEFICIARY_ADDRESS, and added assertions confirming the dedup (single accounts POST, address appearing once).

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: No good reason, cleaned it up — RECIPIENT_ADDRESS is now the beneficiary and BENEFICIARY_ADDRESS the repeated participant, matching the variable names, and dropped the hand-rolled literal.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Good idea, added — _get_network_properties memoizes it the same way, and updated the test to expect one fetch instead of two.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Switched to that everywhere this pattern appeared — simpler and drops the network dependency too.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Same reason as above.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: _sync_block_batch became write-only (takes pre-fetched rows) so all of a batch's network fetches finish before any writes begin — a mid-batch failure leaves zero partial writes. Added a docstring note explaining this so it's not a mystery from the diff alone.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Yes, planned but deferred — this needs `/statements/resolutions/address`  tracking each transaction's block-relative source id for the resolution-entry matching algorithm, which is a bigger change.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Confirmed against symbol-openapi: there's no batch equivalent for `/account/{id}/multisig`, only the per-address route. `POST /accounts` batches account state, but multisig has no counterpart.

explorer/puller/tests/model/symbol/test_Account.py
daoka: Added `test_create_account_row_returns_empty_mosaic_rows_when_account_has_no_mosaics.`

explorer/puller/puller/facade/SymbolPuller.py
daoka: Now that `/network/properties` is memoized (see the caching reply elsewhere in this PR), the passthrough doesn't avoid anything the cache wouldn't already avoid. Dropped the parameter; `_get_native_mosaic_info` now always calls the memoized `_get_network_properties()` internally.

explorer/puller/puller/facade/SymbolPuller.py
daoka: No functional need, you're right — Python dict iteration is already deterministic. Dropped the sorted(), now just list(dirty_addresses.keys()).

explorer/puller/puller/facade/SymbolPuller.py
daoka: Using the dedicated account refresh snapshot batch job ( #2329 /explorer/backend/symbol-accounts-refresh-snapshot) to reset is_harvesting_active to false after 7 days with no harvesting activity. Current mainnet estimate is ~10 minutes per full batch run. Dirty-key refresh intentionally only ever moves the flag toward true (partial per-batch visibility can't prove an account stopped harvesting); the full-population batch is the source of truth for the false transition.

explorer/puller/puller/facade/SymbolPuller.py
daoka: nowhere, actually. Looked closer: importance_percentage is a percentage of the sum across a full refresh run, which is never known at the point this method is called for one account. Removed the parameter entirely rather than just defaulting it to False, since True isn't just currently unused, it's structurally unusable.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Intentional — observed_height/first_seen_height/last_seen_height represent "synced height as of when this current-state snapshot was fetched," not "height the account was modified at." Since the whole batch's dirty accounts are fetched in one POST /accounts call after processing up to H+1, the fetched data genuinely reflects state as of H+1 for every account in that call.
This is also why these columns exist — they back rollback invalidation (last_seen_height/updated_at_height >= fork_height). While looking into this I noticed that invalidation wasn't actually wired up yet for symbol_accounts/symbol_multisig/symbol_account_mosaics, even though other current-state tables already handle this for their own rows. Added it here: rollback repair now deletes rows with last_seen_height/updated_at_height >= fork_height for these three tables too, in FK-safe order, alongside the existing block/transaction/receipt cleanup.

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: i think you can simplify by just mapping dirty_addresses to Address and then using str instead of adding extra map for text lookup 🤔

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: assume the two account are modified at H and H+10 and there is some rollback to H+5. in that case, you'll be rolling back / updating both accounts instead of only the latter. functionally this should work, but seems like potentially extra overhead?

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: correct, by "batch" i meant multiple async calls at once.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
Jaguar0625: i think you mean does_not_mark_transaction_participant_as_**inactive**_harvester since recipient is marked as active prior to sync?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
Jaguar0625: i think you further want to check harvested_block_timestamp since test name is mentioning timestamp? also, you might want to use two blocks since beneficiary and recipient are in same block (thus share timestamp)?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
Jaguar0625: i would swap addresses so beneficiaryAddress=BENEFICIARY_ADDRESS and recipientAddress=RECIPIENT_ADDRESS?

Jaguar0625 (commented):
👍  

(no new comments, replied to some comment threads)

explorer/puller/puller/facade/SymbolPuller.py
daoka: Good point — done. Mapped to a list of Address objects once (.bytes for lookups, str() for text), dropped the parallel text-lookup dict.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Got it — done. multisig lookups are now fetched concurrently per chunk via asyncio.gather, bounded to the same 100-address chunk size already used for the /accounts batch call.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Renamed to does_not_mark_transaction_participant_as_inactive_harvester, no other change.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Restructured to two blocks: the address is only a transaction participant in an old (out-of-window) block, and the block beneficiary in a recent (in-window) block. harvested_block_timestamp itself isn't persisted so can't be asserted directly, but this setup makes is_harvesting_active a real proof — a wrong implementation using the participant's (missing) timestamp would now fail the test.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Agreed — swapped so the constants match their roles, and dropped the now-redundant explicit overrides since they match the helpers' defaults.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Good point to widen the sample — a single live-node snapshot isn't representative. Reframing without the "shallow and rare" claim:
The false-positive window this causes is bounded by the sync batch size (MAX_PAGE_SIZE = 100 blocks, ~50 min at the 30s block target), regardless of how deep the overall unfinalized range gets — an account only gets swept up unnecessarily if it shares a batch with a later-modified account, and batches are capped at 100 blocks. A long finalization stall (I understand these have happened, e.g. from mass voting-key renewal misses) widens how much total repair work might be needed if a fork is eventually found deep in the backlog, but at that point most of the affected batches are genuinely invalid (built on the bad fork), not over-deleted — the "unnecessary" deletion stays confined to whichever single batch actually contains the fork point.
Given that, I'd rather keep last_seen_height as an accurate fetch-time snapshot (needed elsewhere) than add per-account touch-height tracking to shave a sub-batch edge case. Happy to revisit if you see a scenario where the batch-level bound doesn't hold.

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -373,6 +425,150 @@ def _parse_epoch_adjustment(network_properties):
...
+				raise ValueError('Malformed Symbol accounts batch response')
+			account_items_by_address = {bytes.fromhex(item['account']['address']): item for item in accounts_response}
+
+			multisig_responses = await asyncio.gather(*(
+				self.get_symbol_node(f'/account/{address}/multisig', not_found_as_error=False)
+				for address in chunk
+			))
+
+			for address, multisig_response in zip(chunk, multisig_responses):
+				address_bytes = address.bytes
```
Jaguar0625: i believe you should be able to use Address type as key and you can avoid lookup by address.bytes

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: Done. Both account_items_by_address and dirty_addresses now key by Address directly instead of converting to .bytes for the lookup; only one .bytes conversion remains, at the point data is packaged for DB storage (bytea columns / create_multisig_row's first argument).

Jaguar0625 (approved):
👍

PR #2391, opened 2026-07-10
[explorer/puller] feat: throttle Symbol node requests to a configurable rate

daoka:
## Problem

Live-node validation against a public testnet node repeatedly hit `HTTP 429 TooManyRequests` ("You have exceeded your request rate of 60 r/s.") during both the dirty-key multisig fetch loop and the account refresh snapshot's `/accounts` pagination. The account refresh run failed outright after exhausting retries at page 2956 of ~3,159 needed pages (~315K accounts on that testnet).

## Solution

`SymbolPuller.get_symbol_node`/`post_symbol_node` are the only two places any Symbol node request is ever dispatched — every other method (block/transaction/receipt page fetches, the accounts batch `POST`, multisig `GET`, account refresh pagination) funnels through them. Add a `RequestRateLimiter` and call it from those two choke points, so every call site gets throttled by one change instead of needing per-loop pacing.

`RequestRateLimiter` paces callers to a `max_requests_per_second` budget via an `asyncio.Lock`-serialized "next allowed time" watermark; concurrent callers (e.g. the block-page fetches already running via `asyncio.gather` up to `BLOCK_PAGE_FETCH_CONCURRENCY`) queue on the lock and get spaced out rather than all firing at once. `time_source`/`sleep` are constructor-injected so it's unit-testable without waiting in real time.

Defaults to 20 req/s — well under the observed 60 r/s ceiling, leaving headroom for other clients sharing a public node. Exposed as `--max-requests-per-second` on `sync_symbol_block.py` so operators can tune it (raise it for a private node with no such limit, lower it further if still hitting 429s).

`SymbolPuller.__init__` also accepts an optional `rate_limiter` for DI (used by tests); `max_requests_per_second` is ignored when `rate_limiter` is explicitly supplied.

### DIFF
--- a/explorer/puller/puller/facade/SymbolPuller.py
+++ b/explorer/puller/puller/facade/SymbolPuller.py
@@ -11,6 +11,7 @@ from symbollightapi.model.Exceptions import NodeException
 from zenlog import log
 
 from puller.db.SymbolDatabase import SymbolDatabase
+from puller.facade.RequestRateLimiter import RequestRateLimiter
 from puller.model.symbol.Block import create_block_row
 from puller.model.symbol.Receipt import INFLATION_RECEIPT_TYPE, create_receipt_rows
 from puller.model.symbol.Transaction import create_transaction_row
@@ -18,6 +19,7 @@ from puller.model.symbol.Transaction import create_transaction_row
 DatabaseConfiguration = namedtuple('DatabaseConfiguration', ['database', 'user', 'password', 'host', 'port'])
 MAX_PAGE_SIZE = 100
 BLOCK_PAGE_FETCH_CONCURRENCY = 10
+DEFAULT_MAX_REQUESTS_PER_SECOND = 20
 
 
 def _get_symbol_network(network_type):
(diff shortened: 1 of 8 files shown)


explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -108,6 +113,7 @@ async def get_symbol_node(self, url_path, property_name=None, not_found_as_error
 		"""Validates and dispatches a Symbol node GET request."""
 
 		normalized_path = self._validate_symbol_node_path(url_path)
+		await self._rate_limiter.wait_for_turn()
```
Jaguar0625: any reason not to put this in self._retry_operation since that seems to be point where all http requests are made (via lambda)?

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thank you for the feedback. I have moved it into _retry_operation's retry loop. It's also a behavior fix: retries after backoff previously bypassed the limiter and fired outside the rate budget; now every attempt waits for its turn. _retry_operation became an instance method to reach self._rate_limiter, dropping the delay parameter plumbing. Added a test covering the retry path.

Jaguar0625 (approved):
👍

PR #2392, opened 2026-07-10
[explorer/puller] feat: add Symbol account refresh snapshot workflow

daoka:
## Summary

* Add a dedicated account refresh snapshot workflow (`refresh_symbol_accounts.py` / `SymbolPuller.refresh_accounts`) that fully scans `GET /accounts?orderBy=id&order=desc` and writes both current-state (`symbol_accounts`) and immutable per-run snapshot rows (`symbol_account_refresh_accounts`, `symbol_account_refresh_mosaics`). This complements PR #2390 dirty-key refresh (block-sync PR): dirty-key only ever touches accounts that appear as a beneficiary or participant in recently synced blocks/transactions/receipts, so it can't guarantee full population coverage — an account with no recent on-chain activity would never get picked up. This workflow closes that gap by periodically walking every account, which account statics and full-depth list ranks need.
* `symbol_account_list_ranks` is rebuilt per successful run via `ROW_NUMBER() OVER (...)` for `ID`, `IMPORTANCE`, and native-mosaic `BALANCE` scopes — no SQL `OFFSET`, satisfying the phase's required "account list rank pagination without OFFSET" test.
* `symbol_account_refresh_state` (singleton, like `symbol_sync_state`) tracks `status` (`healthy`/`refreshing`/`stale`/`unhealthy`), `last_successful_run_id`, `last_scanned_page` (diagnostic restart cursor), and `last_error`. A failed run marks `status=unhealthy` and records the error but never touches `last_successful_run_id` — the previous successful run stays authoritative for statics/full-depth ranks until a new run succeeds, matching the required "failed refresh keeps previous successful run active" test.
* `is_harvesting_active` is recomputed for every visited account against `get_recently_harvesting_addresses` (distinct block beneficiaries within the same 7-day window used by dirty-key),
* `update_account_importance_rates` computes each account's importance percentage from the completed run's own snapshot total and propagates it to `symbol_accounts` for the same run only; concurrent/interrupted runs never mix percentages from different `refresh_run_id`s.


explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -233,6 +308,30 @@ def create_tables(self):
...
+			'ON symbol_account_refresh_mosaics(refresh_run_id, mosaic_id, amount DESC, address)')
+		cursor.execute(
+			'CREATE INDEX IF NOT EXISTS idx_symbol_account_refresh_mosaics_address '
+			'ON symbol_account_refresh_mosaics(refresh_run_id, address)')
+		cursor.execute(
+			'CREATE INDEX IF NOT EXISTS idx_symbol_account_list_ranks_page '
+			'ON symbol_account_list_ranks(refresh_run_id, rank_scope, rank)')
+		cursor.execute(
+			'CREATE INDEX IF NOT EXISTS idx_symbol_account_list_ranks_address '
+			'ON symbol_account_list_ranks(refresh_run_id, rank_scope, address)')
```
Jaguar0625: is there any reason you define some indexes in list constants at top of file and some inline?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -334,6 +433,87 @@ def _execute_upsert_sync_state(cursor, sync_state):
...
+		cursor.execute("SELECT pg_advisory_lock(hashtext('symbol_account_refresh'))")
+		self.connection.commit()
+
+	def release_account_refresh_lock(self):
+		"""Releases the database-backed account refresh lock."""
+
+		self.connection.rollback()
+		cursor = self.connection.cursor()
+		cursor.execute("SELECT pg_advisory_unlock(hashtext('symbol_account_refresh'))")
+		self.connection.commit()
```
Jaguar0625: based on usage, i'd prefer instead you return object that can be used with with syntax to manage this resource (lock)

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -334,6 +433,87 @@ def _execute_upsert_sync_state(cursor, sync_state):
...
+
+	def mark_account_refresh_failed(self, last_error):
+		"""Records a failed refresh after clearing any aborted transaction."""
+
+		try:
+			self.connection.rollback()
```
Jaguar0625: what do you intend for this rollback to do?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -342,13 +522,17 @@ def upsert_account_current_state(
...
+			cursor = self.connection.cursor()
+			self._execute_upsert_account_current_state(
+				cursor,
+				account_row,
+				mosaic_rows,
+				overwrite_is_harvesting_active)
+			self.connection.commit()
+		except Exception:
+			self.connection.rollback()
+			raise
```
Jaguar0625: consider refactoring this pattern try / commit / rollback; either a function or object supporting with statements

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -686,6 +1138,17 @@ def upsert_receipts_for_height(self, height, receipts, block_reward):
 			(block_reward, height))
 		self.connection.commit()
 
+	@staticmethod
+	def _stale_mark_account_refresh_state_if_needed(cursor, fork_height):
```
Jaguar0625, line 1058: i think this is called after a rollback? when do you reset the status to a different value than stale?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -276,14 +294,306 @@ def test_create_tables_creates_symbol_account_indexes(self):
...
+
+		# Act:
+		raised_error = None
+		try:
+			cursor.execute(
+				"INSERT INTO symbol_account_refresh_state (id, status) VALUES (2, 'healthy')")
+		except PsycopgError as error:
+			raised_error = error
+		finally:
+			database.connection.rollback()
```
Jaguar0625: why do you need explicit rollback call here and in similar tests ?

Jaguar0625: related: why are you not using assertRaises in these tests

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -276,14 +294,306 @@ def test_create_tables_creates_symbol_account_indexes(self):
...
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.mark_account_refresh_failed('refresh failed')
+
+		# The failed transaction must be cleared by the error handler.
+		cursor.execute('SELECT 1')
+		self.assertEqual((1,), cursor.fetchone())
+
+	def test_account_refresh_mosaics_reject_orphan_snapshot_row(self):
```
Jaguar0625: you seem to be testing the db (enforces foreign key). not sure if these test makes sense unless you are calling function on database?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1056,6 +1402,23 @@ def test_can_upsert_account_current_state_and_overwrite_harvesting_active_by_def
...
+			_create_block(2, beneficiary_address=b'at cutoff', timestamp=cutoff),
+			_create_block(3, beneficiary_address=b'at cutoff', timestamp=cutoff + datetime.timedelta(seconds=1)),
+			_create_block(4, beneficiary_address=b'new', timestamp=cutoff + datetime.timedelta(seconds=2))
+		])
+
+		# Act:
+		result = database.get_recently_harvesting_addresses(cutoff)
+
+		# Assert:
+		self.assertEqual({b'at cutoff', b'new'}, result)
```
Jaguar0625: since 'at cutoff' is both at and after, you are not strictly testing that an account that only shows up at cutoff will be returned; similarly you might want to test an account that appears both before and after cutoff

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1528,276 @@ def test_can_update_existing_multisig(self):
...
+		_insert_account_refresh_snapshot_rows(database, {
+			'refresh_run_id': 'run-1', 'account_search_id': 'id-1', 'account_search_order': 0,
+			'account_row': account_row1, 'mosaic_rows': mosaic_rows1, 'snapshot_height': 10, 'snapshot_at': snapshot_at
+		})
+		_insert_account_refresh_snapshot_rows(database, {
+			'refresh_run_id': 'run-1', 'account_search_id': 'id-2', 'account_search_order': 1,
+			'account_row': account_row2, 'mosaic_rows': mosaic_rows2, 'snapshot_height': 10, 'snapshot_at': snapshot_at
+		})
+		database.update_account_importance_rates('run-1')
+		database.activate_account_refresh('run-1', 10, snapshot_at)
```
Jaguar0625: why do you have these as separate functions? is there case when only one of them will be called ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1528,276 @@ def test_can_update_existing_multisig(self):
...
+		cursor = database.connection.cursor()
+		cursor.execute('DROP TABLE symbol_account_refresh_accounts CASCADE')
+		database.connection.commit()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.update_account_importance_rates('run-1')
+
+		cursor.execute('SELECT 1')
+		self.assertEqual((1,), cursor.fetchone())
```
Jaguar0625: do you intend to check symbol_account_refresh_accounts ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1528,276 @@ def test_can_update_existing_multisig(self):
...
+	def test_activate_account_refresh_rolls_back_when_state_table_is_missing(self):
+		# Arrange:
+		database = self._create_database()
+		cursor = database.connection.cursor()
+		cursor.execute('DROP TABLE symbol_account_refresh_state')
+		database.connection.commit()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.activate_account_refresh('run-1', 10, datetime.datetime(2026, 1, 1))
```
Jaguar0625: do you intend to check activate_account_refresh ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1528,276 @@ def test_can_update_existing_multisig(self):
...
+		cursor = database.connection.cursor()
+		cursor.execute('DROP TABLE symbol_account_list_ranks')
+		database.connection.commit()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.rebuild_account_list_ranks('run-1', NATIVE_MOSAIC_ID)
+
+		cursor.execute('SELECT 1')
+		self.assertEqual((1,), cursor.fetchone())
```
Jaguar0625: do you intend to check symbol_account_list_ranks ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1528,276 @@ def test_can_update_existing_multisig(self):
...
+		# repopulated by the next dirty-key touch or account refresh snapshot run, not by rollback repair itself.
+		self.assertEqual([], account_results)
+		self.assertEqual([], mosaic_results)
+		self.assertEqual([], multisig_results)
+		self.assertEqual(
+			[('run-1', 'id-1', expected_address)],
+			[(run_id, account_search_id, bytes(address)) for run_id, account_search_id, address in refresh_account_results])
+		self.assertEqual(
+			[('run-1', expected_address, expected_mosaic_row['mosaic_id'], expected_mosaic_row['amount'])],
+			[(run_id, bytes(address), mosaic_id, amount) for run_id, address, mosaic_id, amount in refresh_mosaic_results])
```
Jaguar0625, line 1786: what is reason for not also deleting these values? if you're dropping blocks >=10, aren't the refresh values also unlinked from the main chain?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1399,6 +2032,20 @@ def test_can_delete_blocks_from_height(self):
...
+		cursor = database.connection.cursor()
+		cursor.execute('DROP TABLE symbol_transaction_addresses')
+		database.connection.commit()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.delete_blocks_from_height(1)
+
+		cursor.execute('SELECT 1')
+		self.assertEqual((1,), cursor.fetchone())
```
Jaguar0625: do you intend to check symbol_account_refresh_state ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -574,6 +576,77 @@ def _is_harvested_block_within_active_window(harvested_block_timestamp):
...
+			except Exception as state_error:  # pylint: disable=broad-exception-caught
+				# Preserve the original refresh failure when failure-state persistence also fails.
+				log.error(f'Failed to record Symbol account refresh failure: {state_error}')
+			raise
+		finally:
+			try:
+				self.symbol_db.release_account_refresh_lock()
+			except Exception as lock_error:  # pylint: disable=broad-exception-caught
+				# Preserve the refresh outcome when lock cleanup fails.
+				log.error(f'Failed to release Symbol account refresh lock: {lock_error}')
```
Jaguar0625: is this recoverable? i mean in this state isn't entire puller basically stuck? if so, it feels like we shouldn't handle it and let puller die

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -574,6 +576,77 @@ def _is_harvested_block_within_active_window(harvested_block_timestamp):
 		cutoff_timestamp = datetime.now(timezone.utc) - timedelta(days=HARVESTING_ACTIVE_WINDOW_DAYS)
 		return harvested_block_timestamp >= cutoff_timestamp
 
+	async def refresh_accounts(self):  # pylint: disable=too-many-locals
+		"""Refreshes the full Symbol account population snapshot and current-state account rows."""
+
+		self.symbol_db.acquire_account_refresh_lock()
+		try:
+			refresh_run_id = str(uuid.uuid4())
```
Jaguar0625: do you need this to be globally unique vs just a counter ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -574,6 +576,77 @@ def _is_harvested_block_within_active_window(harvested_block_timestamp):
...
+					account_row, mosaic_rows = create_account_row(
+						item,
+						self.symbol_facade.network,
+						snapshot_height,
+						native_mosaic_id,
+						native_mosaic_divisibility)
+					account_row['is_harvesting_active'] = account_row['address'] in recently_harvesting_addresses
+					account_entries.append({
+						'refresh_run_id': refresh_run_id,
+						'account_search_id': item['id'],
```
Jaguar0625: isn't this decoded hex address ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -574,6 +576,77 @@ def _is_harvested_block_within_active_window(harvested_block_timestamp):
...
+						'account_search_order': account_search_order,
```
Jaguar0625: how are you using this value ?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -480,6 +664,258 @@ def upsert_multisig(self, address, multisig_row_or_none):
...
+				account_search_order = EXCLUDED.account_search_order,
+				public_key = EXCLUDED.public_key,
+				account_type = EXCLUDED.account_type,
+				importance = EXCLUDED.importance,
+				importance_percentage = EXCLUDED.importance_percentage,
+				snapshot_height = EXCLUDED.snapshot_height,
+				snapshot_at = EXCLUDED.snapshot_at
+			''',
+			{
+				**account_row,
```
Jaguar0625, line 687: why are you storing the account_row data in this table too vs having a foreign key link?
same for mosaic_row below.

explorer/puller/puller/workflows/refresh_symbol_accounts.py
```diff
@@ -0,0 +1,51 @@
+# pylint: disable=duplicate-code
```
Jaguar0625: can you refactor this with sync_symbol_block. aside from different call sync_block_headers vs refresh_accounts it is almost identical .

for motivation, you can look at [shoestring](https://github.com/symbol/product/blob/dev/tools/shoestring/shoestring/__main__.py) and its commands.

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -357,6 +359,15 @@ async def get(self, url_path, *_):  # pylint: disable=too-many-return-statements
 				'code': 'ResourceNotFound',
 				'message': f'no resource exists with id {address_text}'
 			})
+		if url_path.startswith('accounts?pageSize=100&pageNumber='):
+			page_number = int(url_path.split('pageNumber=')[1].split('&')[0])
```
Jaguar0625: can you use a regex instead ?

explorer/puller/tests/workflows/test_refresh_symbol_accounts.py
```diff
@@ -0,0 +1,155 @@
+# pylint: disable=duplicate-code
```
Jaguar0625: similarly, try to refactor the tests with the blocks ones since there is a lot of similarity/duplication

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -65,6 +159,69 @@ def _fetch_account_count(self, address_hex=BENEFICIARY_ADDRESS):
...
+			'chain_height': 1,
+			'finalized_height': 0,
+			'finalized_hash': b'finalized',
+			'finalized_epoch': 0,
+			'finalized_point': 0,
+			'last_synced_height': 0,
+			'last_synced_block_hash': b'last'
+		}
+
+	def _run_refresh_while_rollback_waits(self, connector):
```
Jaguar0625: are you assuming the pull blocks and pull accounts scripts will be executing in parallel? otherwise, i don't see how this is realistic scenario?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+				'''
+				SELECT rank, address
+				FROM symbol_account_list_ranks
+				WHERE refresh_run_id = %s AND rank_scope = %s
+				ORDER BY rank
+				''',
+				(refresh_run_id, rank_scope))
+			self.assertEqual(
+				list(enumerate(expected_addresses)),
+				[(rank, bytes(address)) for rank, address in cursor.fetchall()])
```
Jaguar0625: what am i missing? isn't list(enumerate(expected_addresses)) an address list being compared against a (rank, address) tuple list ?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+		connector = FakeConnector(101, {}, account_pages={1: page1, 2: page2})
+		set_symbol_connector(self.puller, connector)
+
+		# Act:
+		asyncio.run(self.puller.refresh_accounts())
+
+		# Assert:
+		state = self.puller.symbol_db.get_account_refresh_state()
+		self._assert_complete_refresh_state(state)
+		self._assert_complete_current_state()
```
Jaguar0625: probably rename to indicate accounts are being checked

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+
+		try:
+			# Act:
+			propagated_error = None
+			try:
+				asyncio.run(self.puller.refresh_accounts())
+			except RefreshFailureError as error:
+				propagated_error = error
+		finally:
+			self.puller.symbol_db = database
```
Jaguar0625, line 755: any reason not to use assertRaisesRegex here ?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+		# Arrange:
+		database = self.exit_stack.enter_context(SymbolDatabase(self.db_config))
+		sync_state = self._create_rollback_sync_state(status='invalid')
+
+		# Act:
+		raised_error = None
+		try:
+			database.repair_rollback_from_height(1, sync_state)
+		except PsycopgError as error:
+			raised_error = error
```
Jaguar0625: assertRaises ?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+		# Arrange:
+		page = [create_account_item(_address_hex(1), 'id-1', importance='10')]
+		connector = FakeConnector(1, {}, account_pages={1: page})
+		set_symbol_connector(self.puller, connector)
+
+		# Act:
+		asyncio.run(self.puller.refresh_accounts())
+		first_run_id = self.puller.symbol_db.get_account_refresh_state()['last_successful_run_id']
+		asyncio.run(self.puller.refresh_accounts())
+		second_run_id = self.puller.symbol_db.get_account_refresh_state()['last_successful_run_id']
```
Jaguar0625: why two calls to get_account_refresh_state ? 
aren't the below queries the same for the both first_run_id and second_run_id (aside from id)?

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +601,532 @@ def test_get_native_mosaic_info_is_memoized(self):
...
+	def test_refresh_accounts_recomputes_harvesting_active_for_every_visited_account(self):
+		# Arrange:
+		active_address = _address_hex(1)
+		inactive_address = _address_hex(2)
+		account_row, mosaic_rows = self._create_current_account_row(inactive_address, importance='100')
+		account_row['is_harvesting_active'] = True
+		self.puller.symbol_db.upsert_account_current_state(account_row, mosaic_rows)
+		self.puller.symbol_db.upsert_blocks([
+			self._create_block_row(1, active_address, datetime.now(timezone.utc) - timedelta(days=1)),
+			self._create_block_row(2, inactive_address, datetime.now(timezone.utc) - timedelta(days=8))
```
Jaguar0625: might want to swap so blocks have increasing timestamps ?

Jaguar0625 (commented):
👍

explorer/puller/puller/db/SymbolDatabase.py
daoka: There was no reason for the inconsistency. I moved the refresh-related index definitions into `ACCOUNT_REFRESH_INDEXES` and now create them through the same list-based pattern.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Updated. The separate acquire/release methods have been replaced with an `account_refresh_lock()` context manager, and all refresh and rollback callers now use it with `with`.

explorer/puller/puller/db/SymbolDatabase.py
daoka: The rollback clears a transaction that might have been left in an aborted state by the refresh failure. Without clearing it first, the subsequent update that records the `unhealthy` state could not be committed. I kept this behavior and documented the intent in the method docstring.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Updated. I added a shared `_transaction()` context manager that commits on success and rolls back on failure, and applied it to the affected refresh and rollback database operations.

explorer/puller/puller/db/SymbolDatabase.py
daoka: It is reset by the next account refresh. The refresh sets the status to `refreshing` when it starts and `finalize_account_refresh()` sets it to `healthy` after a successful publish. A failed refresh changes it to `unhealthy` instead.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: > why do you need explicit rollback call here and in similar tests ?

It is no longer needed. The database method now owns rollback through `_transaction()`, so I removed the explicit test cleanup and changed the test to use `assertRaises`.

> related: why are you not using assertRaises in these tests

Updated to use `with self.assertRaises(PsycopgError):`.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Agreed. These tests were only exercising PostgreSQL constraints through raw SQL, so I removed them. The schema introspection test remains and verifies that the composite foreign key is created as part of the required schema.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Updated. The fixture now covers four independent cases: before only, exactly at the cutoff only, both before and after, and after only. Block heights and timestamps are also ordered consistently.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: There was no caller that needed only one of these operations. I combined importance calculation, rank rebuilding, current-state publication, and activation into `finalize_account_refresh()`, which executes them in one transaction.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes. I replaced the `SELECT 1` probe with a finalization rollback test that verifies the previous `symbol_account_refresh_accounts.importance_percentage` value is restored when the final state update fails.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes. Activation is now part of `finalize_account_refresh()`. Its rollback test verifies that the previous healthy status, successful run ID, and completed height remain unchanged when finalization fails.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes. The consolidated finalization rollback test now seeds an existing rank row and verifies that it is restored when the final state update fails.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes. I replaced the probe with a rollback repair test that fails at the sync-state update and verifies both that the deleted block is restored and that the account refresh state remains `healthy` rather than being left `stale`.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Agreed. An unlock failure after a successful operation now propagates and terminates the puller. If the operation itself already failed, the unlock failure is treated as a secondary cleanup error and logged so that it does not replace the original exception.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Updated. `FakeConnector` now uses `re.fullmatch()` and a capture group to extract the account page number.

explorer/puller/tests/workflows/test_refresh_symbol_accounts.py
daoka: Updated. The shared fake puller, environment setup, and workflow assertions were moved into `workflow_test_utils.py`. I also added dedicated tests for `symbol_workflow_utils.py`, leaving each caller test focused on its workflow-specific behavior.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: They do not always need to run in parallel. They can run sequentially in the same scheduled job or as independent jobs. The lock protects the supported independent-execution case, where an account refresh could otherwise overlap rollback repair; the test exercises that mutual exclusion.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: `enumerate(expected_addresses)` does produce the expected `(zero-based rank, address)` pairs, but the intent was not clear. I renamed the intermediate value to `expected_rank_address_pairs` and added a comment that the rank is zero-based.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Updated. I renamed it to `_assert_complete_current_account_state()`.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: No reason. I changed this to `assertRaisesRegex` and retained the identity assertion to verify that the exact original refresh exception is propagated.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Updated to use `with self.assertRaises(PsycopgError):`.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Agreed. I changed the two refresh executions to a loop with one state read per run and moved the repeated snapshot/rank queries into a helper.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Updated. The seeded blocks now have timestamps that increase with height.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes, after the rollback the previous refresh snapshot may no longer represent the canonical chain. That is why the rollback marks the refresh state as stale when its completed height is at or above the fork height. We intentionally retain the run-scoped immutable snapshot and rank rows for diagnostics rather than deleting them. The mutable current-state rows are removed, and the next successful refresh creates a new run and moves last_successful_run_id to it. Consumers must treat a non-healthy refresh state as unavailable.

explorer/puller/puller/facade/SymbolPuller.py
daoka: It does not need to be globally unique outside this database, but it must not be reused while persisted snapshot runs remain. A process-local counter would reset across restarts, and deriving it only from the last successful run could reuse an ID left by a failed partial run. A database-backed sequence would work, but it would require additional schema and persistent allocation management. I kept the UUID because it provides run uniqueness across processes, restarts, and failed runs without introducing that additional state.

explorer/puller/puller/facade/SymbolPuller.py
daoka: `account_search_order` records the zero-based position of each account in the paginated `GET /accounts?orderBy=id&order=desc` response, and the counter continues across pages. During finalization, we materialize the `ID` rank scope with `ROW_NUMBER() OVER (ORDER BY account_search_order ASC, address) - 1`. The account list REST route then paginates this rank scope, so it does not depend on address order, insertion order, or PostgreSQL collation of the raw REST IDs.

explorer/puller/puller/db/SymbolDatabase.py
daoka: `symbol_accounts` and `symbol_account_mosaics` are mutable current-state tables. They can be updated independently by dirty-key refreshes, and rollback repair can delete their rows. The refresh tables represent a run-scoped snapshot used to build statistics and list ranks, so their values must remain stable after the run completes. Referencing the current-state rows would make an older run observe newer values, and a foreign key would either block rollback deletion or cascade-delete the snapshot.

The snapshot mosaics therefore reference the account snapshot from the same run through `(refresh_run_id, address)`, rather than referencing the mutable current-state tables.

explorer/puller/puller/facade/SymbolPuller.py
daoka: The decoded account address actually comes from item['account']['address']; the top-level item['id'] is the REST record ID. However, after rechecking the pagination flow, we found that this ID does not need to be persisted. We keep requesting orderBy=id&order=desc and materialize the returned order as account_search_order, which is used to build the ID rank. I removed account_search_id and the now-unused sort_value_text handling, and updated the tests to verify that account refresh no longer depends on the top-level id.

explorer/puller/puller/workflows/refresh_symbol_accounts.py
daoka: Following your suggestion, I consolidated the Symbol workflows behind a dedicated Shoestring-style dispatcher. The dispatcher owns subcommand registration and shared CLI behavior, while each command module exposes only `add_arguments()` and `run_main()`. Since this code is pre-release, the standalone compatibility entrypoints were removed. The canonical commands are `python -m puller.workflows.symbol sync-block` and `python -m puller.workflows.symbol refresh-accounts`. NEM workflows and the root-level CLI were not changed.

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -334,6 +428,97 @@ def _execute_upsert_sync_state(cursor, sync_state):
...
+		"""Gets the singleton Symbol account refresh state."""
+
+		cursor = self.connection.cursor()
+		cursor.execute(f'SELECT {", ".join(ACCOUNT_REFRESH_STATE_COLUMNS)} FROM symbol_account_refresh_state WHERE id = 1')
+		result = cursor.fetchone()
+
+		return dict(zip(ACCOUNT_REFRESH_STATE_COLUMNS, result)) if result else None
+
+	@contextmanager
+	def _transaction(self):
```
Jaguar0625: consider renaming to something like _database_transaction so it's clearer to the reader this doesn't have something to do with Symbol transaction

explorer/puller/puller/db/SymbolDatabase.py
Jaguar0625: > left in an aborted state by the refresh failure

is there a reason you don't want to clear the transaction upon failure (in an exception handler)?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1165,6 +1545,312 @@ def test_can_update_existing_multisig(self):
...
+				refresh_run_id, rank_scope, rank, address, sort_value_numeric, mosaic_id
+			) VALUES ('run-1', 'ID', 0, %s, NULL, NULL)
+			''',
+			(bytes.fromhex(ADDRESS1),))
+		database.connection.commit()
+
+		# Act + Assert:
+		with self.assertRaises(PsycopgError):
+			database.finalize_account_refresh(
+				'run-1', NATIVE_MOSAIC_ID, 'invalid-height', datetime.datetime(2026, 1, 1))
```
Jaguar0625, line 1665: if height was valid, importance height would be 0.123 based on _insert_account_refresh_snapshot_rows, correct?

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: is this just informational? since you're using order desc, when new accounts are added to blockchain, they should precede existing ones so all indexes will be different on subsequent fetches.

explorer/puller/tests/workflows/test_symbol.py
```diff
@@ -0,0 +1,174 @@
...
+	]
+
+
+def test_help_lists_symbol_commands(capsys):
+	# Act + Assert:
+	with pytest.raises(SystemExit) as exception:
+		asyncio.run(main(['--help']))
+
+	# Assert:
+	assert exception.value.code == 0
```
Jaguar0625: typically, we put expected values on lhs

explorer/puller/tests/workflows/test_symbol.py
```diff
@@ -0,0 +1,174 @@
...
+	# Act + Assert:
+	with pytest.raises(SystemExit) as exception:
+		asyncio.run(main([command, '--help']))
+
+	# Assert:
+	assert exception.value.code == 0
+	assert expected_text in capsys.readouterr().out
+
+
+def test_sync_dispatch_forwards_args():
```
Jaguar0625: prefer using full subcommand names - sync_block, refresh_accounts - in test names for increased clarity, especially if another similarly named command is added later

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
Jaguar0625: i understand, but this seems like something that should be handled at a higher layer - the layer running the commands - vs adding complexity to the puller 🤔

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +605,561 @@ def test_get_native_mosaic_info_is_memoized(self):
 		self.assertEqual(1, connector.paths.count('network/properties'))
 		self.assertEqual(1, connector.paths.count(f'mosaics/{NATIVE_MOSAIC_ID}'))
 
+	def test_refresh_accounts_pages_all_accounts_and_assigns_search_order_across_pages(self):
+		# Arrange:
+		page1 = []
+		for index in range(100):
+			item = create_account_item(_address_hex(index), f'id-{index}', importance=str(index + 1))
+			item.pop('id')
+			page1.append(item)
```
Jaguar0625: consider adding a helper like strip_id so you can use list comprehension

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
Jaguar0625: could you also change something - like account balance - between the two runs ?

Jaguar0625 (commented):
👍

explorer/puller/puller/db/SymbolDatabase.py
daoka: Updated. I renamed _transaction() to _database_transaction() so that it is clearly identified as a database transaction helper rather than a Symbol transaction operation.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Agreed. Each database operation now uses _database_transaction(), which rolls back in its exception path before re-raising. mark_account_refresh_failed() therefore no longer performs a defensive rollback before recording the failure state.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: No. 0.123 is the pre-finalization rollback sentinel. With a valid height, finalization would recompute the percentage as 100 / 100 = 1. I added a comment explaining why 0.123 and 0.456 are deliberately different from the successful result.

explorer/puller/tests/workflows/test_symbol.py
daoka: > is this just informational? since you're using order desc, when new accounts are added to blockchain, they should precede existing ones so all indexes will be different on subsequent fetches.

Yes, it is intentionally run-scoped rather than a stable account identifier. New accounts can shift every position in a later run, which is expected because each successful refresh rebuilds the ID rank from the observed orderBy=id&order=desc response order.

> typically, we put expected values on lhs

Updated. I changed the workflow equality assertions so that the expected value is consistently on the left-hand side.

explorer/puller/tests/workflows/test_symbol.py
daoka: Updated. The workflow test names now use the complete sync_block and refresh_accounts command names and describe the condition and expected behavior.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: > i understand, but this seems like something that should be handled at a higher layer - the layer running the commands - vs adding complexity to the puller 🤔

I removed the Puller advisory lock and the lock-specific tests. Exclusion between refresh-accounts and sync-block rollback repair is now owned by the external scheduler or operator, including manual execution, and the contract is documented in the Puller docstring and CLI help.

> consider adding a helper like strip_id so you can use list comprehension

Updated. I added a non-mutating _strip_id() helper and now use it from the list comprehension. It removes only the top-level REST record ID and leaves the input payload unchanged.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Updated. The restart test now changes the native mosaic amount between runs. It verifies the complete immutable snapshot for each run independently and confirms that the mutable current-state mosaic contains the second run’s amount.

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -444,10 +523,434 @@ def test_get_native_mosaic_info_is_memoized(self):
 		self.assertEqual(1, connector.paths.count('network/properties'))
 		self.assertEqual(1, connector.paths.count(f'mosaics/{NATIVE_MOSAIC_ID}'))
 
+	def test_refresh_accounts_pages_all_accounts_and_assigns_search_order_across_pages(self):
+		# Arrange:
+		page1 = []
+		for index in range(100):
+			page1.append(_strip_id(create_account_item(_address_hex(index), f'id-{index}', importance=str(index + 1))))
```
Jaguar0625, line 530: use list comprehension ?

Jaguar0625 (approved):
👍

Issue #2357, opened 2026-06-28
[explorer] Show harvested block statistics and recent blocks on account detail page


cryptoBeliever:
## Summary
Add a **Harvesting** section to the account detail page (`/accounts/[address]`) showing block-harvest activity for the account - similar to the legacy NEM explorer - including aggregate counts and a list of recently harvested blocks.

The section should cover harvesting **directly** and **via a linked remote account** (delegated harvesting).

## Motivation

The account detail page currently shows balance, importance, vested balance, mosaics, and transactions, but **no harvesting activity**.

Users running harvesting nodes (or main accounts with a remote harvester) cannot answer basic questions from the explorer:
- How many blocks has this account harvested (lifetime)?
- How active is harvesting recently (last day / month)?
- Which blocks were harvested most recently?

PR #2393, opened 2026-07-12
[explorer/frontend] feat: add harvest state section

AnthonyLaw:
Problem: Missing harvest state information from account pages. https://github.com/symbol/product/issues/2357
Solution: Show harvest state when harvested blocks > 0 

image


OlegMakarenko (commented):
The sections should not change each other width. Also I would better move harvesting into the history section with tabs: `Transactions` and `Harvested` for displaying the list of received rewards. If we really need to present the total harvested amount, I would move it into a new statistics section or tab. 

Let me design it first.

cryptoBeliever:
@OlegMakarenko, here are a few ideas for adding a harvested blocklist and statistics. Maybe one of them will be a good fit with a bit of fine-tuning.

1) 
W1-desktop
W1-mobile

2) 
W2-desktop
W2-mobile

3) 
W3-desktop
W3-mobile

4) 
W4-desktop
W4-mobile

5) 
W5-desktop
W5-mobile

PR #2401, opened 2026-07-14
[explorer/puller] feat: resolve namespace aliases in Symbol transaction fields

daoka:

Symbol transaction fields can contain unresolved namespace aliases instead of resolved addresses or mosaic ids.

The puller currently persists these unresolved values in normalized transaction columns and child rows. Address aliases are also excluded from dirty-account refresh, so accounts referenced only through aliases are not refreshed or searchable by their resolved address.

## What's the issue?

Persisting unresolved aliases in normalized transaction fields causes two related problems:

- `symbol_transaction_addresses` and transaction address columns do not reference the actual resolved account.
- Accounts referenced only through namespace aliases are omitted from dirty-account refresh.

Address persistence and dirty-account collection must use the same resolved value. Fixing only one side would leave inconsistent transaction and account state.

- Fetch address and mosaic resolution statements only for heights and resolution kinds where aliases are present.
- Resolve top-level and embedded transaction aliases using their transaction statement source.
- Apply the Symbol SDK-compatible resolution-entry selection algorithm.
- Store resolved values in normalized transaction address and mosaic columns.
- Run alias resolution before dirty-address collection so resolved accounts are included in account refresh.
- Deduplicate address rows when an alias resolves to an existing `(address, role)` pair.
- Preserve the original unresolved values in `body` and `raw_payload`.
- Reject missing statements, missing applicable entries, and missing aggregate parents before writing the batch.
- Add regression coverage for paging, per-kind requests, embedded transactions, restart convergence, no-partial-write behavior, and original payload preservation.


explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -411,6 +416,135 @@ async def _get_receipt_rows_by_height(self, start_height, end_height):
...
 
+	async def _get_resolution_statements(self, kind, height):
+		resolution_entries_by_unresolved = {}
+		page_number = 1
+		while True:
+			response = await self.get_symbol_node(
+				f'/statements/resolutions/{kind}?height={height}&pageSize={MAX_PAGE_SIZE}&pageNumber={page_number}'
+			)
+			if not isinstance(response, dict) or 'data' not in response:
+				raise ValueError(f'Malformed Symbol {kind} resolution page response')
```
Jaguar0625: since you have this same check in many places, consider refactoring into own function

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -411,6 +416,135 @@ async def _get_receipt_rows_by_height(self, start_height, end_height):
...
+
+		parent_row = top_level_rows_by_hash.get(row['aggregate_hash'])
+		if parent_row is None:
+			raise ValueError(f'Missing aggregate transaction for embedded transaction at height {row["height"]}')
+
+		return TransactionSource(parent_row['block_index'] + 1, row['embedded_index'] + 1)
+
+	@staticmethod
+	def _resolve_transaction_alias(statements, unresolved, source, kind, height):
+		unresolved_hex = unresolved.hex().upper() if isinstance(unresolved, bytes) else unresolved.upper()
```
Jaguar0625: are you calling this with both string and bytes input?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -411,6 +416,135 @@ async def _get_receipt_rows_by_height(self, start_height, end_height):
...
+		if unresolved_hex not in statements:
+			raise ValueError(f'Missing Symbol {kind} resolution at height {height} for unresolved {unresolved_hex}')
+
+		resolved = select_resolution_entry(statements[unresolved_hex], source.primary_id, source.secondary_id)
+		if resolved is None:
+			raise ValueError(f'Missing Symbol {kind} resolution entry at height {height} for unresolved {unresolved_hex}')
+
+		return bytes.fromhex(resolved) if 'address' == kind else resolved
```
Jaguar0625: shouldn't select_resolution_entry always return same type ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -411,6 +416,135 @@ async def _get_receipt_rows_by_height(self, start_height, end_height):
...
+
+		for height, transaction_rows in transaction_rows_by_height.items():
+			if height not in resolution_statements_by_height:
+				continue
+
+			resolution_statements = resolution_statements_by_height[height]
+			top_level_rows_by_hash = {
+				row['hash']: row
+				for row in transaction_rows
+				if not row['is_embedded']
```
Jaguar0625, line 556: where are you handling embedded rows?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -411,6 +416,135 @@ async def _get_receipt_rows_by_height(self, start_height, end_height):
...
+		}
+		return alias_addresses, alias_mosaic_ids
+
+	def _resolve_transaction_row(self, row, top_level_rows_by_hash, resolution_statements, height):
+		row_has_alias_address = any(Address(address_row['address']).is_alias() for address_row in row['address_rows'])
+		row_has_alias_address = row_has_alias_address or any(
+			address is not None and Address(address).is_alias()
+			for address in (row['recipient_address'], row['target_address'])
+		)
+		row_has_alias_mosaic = any(is_alias_mosaic_id(mosaic_row['mosaic_id']) for mosaic_row in row['mosaic_rows'])
```
Jaguar0625: refactor the checks for aliases with the checks in _alias_values_for_transaction_rows, so you're using same logic in both places

explorer/puller/puller/model/symbol/Resolution.py
```diff
@@ -0,0 +1,61 @@
...
+	"""Returns whether a mosaic id encodes a namespace alias."""
+
+	return is_mosaic_alias(int(mosaic_id_hex, 16))
+
+
+def select_resolution_entry(resolution_entries, primary_id, secondary_id):
+	"""Selects the resolution value applicable to a transaction source."""
+
+	if 1 == len(resolution_entries):
+		return resolution_entries[0]['resolved']
```
Jaguar0625: this seems wrong, statement will only apply if it's source is >= (primary_id, secondary_id)

explorer/puller/puller/model/symbol/Resolution.py
```diff
@@ -0,0 +1,61 @@
+from symbolchain.symbol.IdGenerator import is_mosaic_alias
+
+
+def is_alias_mosaic_id(mosaic_id_hex):
+	"""Returns whether a mosaic id encodes a namespace alias."""
+
+	return is_mosaic_alias(int(mosaic_id_hex, 16))
+
+
+def select_resolution_entry(resolution_entries, primary_id, secondary_id):
```
Jaguar0625, line 10: this seems a bit overcomplicated. is there a reason you're not just looping over resolution_entries and finding one with largest source <= (primary_id, secondary_id) ?

explorer/puller/tests/model/symbol/test_Resolution.py
```diff
@@ -0,0 +1,76 @@
...
+		entries = [_entry(1, 0), _entry(2, 0), _entry(5, 6)]
+		self._assert_selects_entry(entries, 5, 3, '0200')
+
+	def test_select_resolution_entry_falls_back_to_last_previous_primary_entry_when_current_primary_has_no_applicable_secondary_id(self):
+		entries = [_entry(1, 1, 'A'), _entry(1, 4, 'B'), _entry(1, 7, 'C'), _entry(2, 4, 'D')]
+		self._assert_selects_entry(entries, 2, 2, 'C')
+
+	def test_select_resolution_entry_falls_back_when_current_primary_has_zero_entry(self):
+		entries = [_entry(1, 0, 'A'), _entry(2, 0, 'B'), _entry(2, 4, 'C')]
+		self._assert_selects_entry(entries, 2, 2, 'A')
```
Jaguar0625: why is this not 'B' ?

explorer/puller/tests/model/symbol/test_Resolution.py
```diff
@@ -0,0 +1,76 @@
...
+
+	def _assert_selects_entry(self, entries, primary_id, secondary_id, expected):
+		# Act:
+		resolved = select_resolution_entry(entries, primary_id, secondary_id)
+
+		# Assert:
+		self.assertEqual(expected, resolved)
+
+	def test_select_resolution_entry_returns_single_entry_without_source_comparison(self):
+		self._assert_selects_entry([_entry(4, 2)], 1, 0, '0402')
```
Jaguar0625: entry at (4, 1) will not be in affect at (1, 0)

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute(
+			'''
+			SELECT encode(transaction.recipient_address, 'hex'), encode(address.address, 'hex'), address.role
+			FROM symbol_transactions transaction
+			JOIN symbol_transaction_addresses address ON address.transaction_id = transaction.id
+			WHERE address.role = 'recipient'
+			''')
+		self.assertEqual((RESOLVED_ADDRESS.lower(), RESOLVED_ADDRESS.lower(), 'recipient'), cursor.fetchone())
```
Jaguar0625, line 329: why is this not DECOY? isn't there only one transaction in block?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute(
+			'''
+			SELECT encode(transaction.target_address, 'hex'), encode(address.address, 'hex'), address.role
+			FROM symbol_transactions transaction
+			JOIN symbol_transaction_addresses address ON address.transaction_id = transaction.id
+			WHERE transaction.is_embedded AND address.role = 'target'
+			''')
+		self.assertEqual((RESOLVED_ADDRESS.lower(), RESOLVED_ADDRESS.lower(), 'target'), cursor.fetchone())
```
Jaguar0625, line 384: likewise, isn't embedded (1, 1) ?

Jaguar0625, line 384: check _resolution_paths in these two tests too ?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute("SELECT encode(recipient_address, 'hex') FROM symbol_transactions ORDER BY hash")
+		self.assertEqual([(RESOLVED_ADDRESS.lower(),), (RESOLVED_ADDRESS.lower(),)], cursor.fetchall())
+		self.assertEqual([
+			resolution_path('address', 1),
+			resolution_path('address', 1, 2)
+		], self._resolution_paths(connector))
+
```
Jaguar0625, line 577: would confirm both aliases appear in respective tx body

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+				transaction_path(1, 1): {'data': [create_node_transaction(1, recipientAddress=ALIAS_ADDRESS)]}
+			},
+			address_resolutions_by_height={
+				1: [create_resolution_statement(1, ALIAS_ADDRESS, [_resolution_entry(1, 0, RESOLVED_ADDRESS)])]
+			})
+		self._sync_with_connector(connector)
+		first_state = self._fetch_transaction_resolution_state()
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute('DELETE FROM symbol_sync_state')
+		self.puller.symbol_db.connection.commit()
```
Jaguar0625, line 644: what is purpose of this test ?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(first_state, self._fetch_transaction_resolution_state())
+		self.assertEqual([
+			resolution_path('address', 1),
+			resolution_path('address', 1)
+		], self._resolution_paths(connector))
+
+	def test_sync_block_headers_rejects_resolution_without_applicable_entry(self):
```
Jaguar0625: should you have similar check for mosaic resolution?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+				1: [create_resolution_statement(1, ALIAS_ADDRESS, [_resolution_entry(1, 1, RESOLVED_ADDRESS)])]
+			})
+
+		# Act + Assert:
+		self._assert_sync_rejects_node_response(
+			connector,
+			ValueError,
+			'Missing aggregate transaction.*height 1')
+
+	def test_sync_block_headers_rejects_malformed_resolution_page(self):
```
Jaguar0625: should you have similar check for mosaic resolution?

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +255,440 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+				transaction_path(1, 1): {'data': [create_node_transaction(1, recipientAddress=ALIAS_ADDRESS)]}
+			})
+
+		# Act + Assert:
+		self._assert_sync_rejects_node_response(
+			connector,
+			ValueError,
+			'Malformed Symbol address resolution page response')
+
+	def test_sync_block_headers_rejects_non_dict_resolution_page(self):
```
Jaguar0625: should you have similar check for mosaic resolution?

Jaguar0625 (changes requested):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thanks. I extracted `_get_node_page_data` and now use it for transaction, block, statement, resolution, and account page responses.

explorer/puller/puller/facade/SymbolPuller.py
daoka: It was previously called with both types. I changed the contract so `unresolved_hex` is always an uppercase hex string. Address bytes are converted by the caller, while mosaic IDs are uppercased directly.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Agreed. `select_resolution_entry` now consistently returns the resolved hex string from the node DTO, or `None` when no entry applies. Address conversion to bytes is handled only by the address caller.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Embedded rows are fetched with `embedded=true` and processed in the same row loop. `_transaction_resolution_source` looks up the parent aggregate by hash and calculates `(parent block_index + 1, embedded_index + 1)`. I also updated the integration test to verify source `(1, 1)`.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Done. I extracted `_alias_values_for_transaction_row` and now reuse the same address and mosaic alias detection logic for both batch request planning and per-row resolution.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: You're right; the previous fixture was inconsistent. I changed it to contain two top-level transactions with unique IDs at block indexes 0 and 1. The alias is on the second transaction, so its source is `(2, 0)` and the non-decoy resolution is expected.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Yes. I updated the aggregate block index and embedded index to 0, so the embedded source is `(1, 1)`. The resolution entries now use `(1, 0)` as the decoy and `(1, 1)` as the expected value.

daoka: Added. Both the top-level and embedded source tests now assert the exact address resolution request path for height 1.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Added. The paging test now selects both transactions by their `bytea` hashes and verifies each transaction hash, the original alias in its body, and its distinct resolved recipient address.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: This test verifies restart convergence. After a completed sync, it removes the sync state to force the same height to be synchronized again, then confirms that the resolved rows remain identical and the resolution endpoint is fetched again. I renamed the test and added a comment to make this explicit.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Added a mosaic counterpart. It verifies the error when no resolution entry applies, the exact mosaic resolution request path, and that no partial writes are made to blocks, sync state, transactions, transaction mosaics, receipts, or accounts.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Added a malformed mosaic resolution page test. It exercises the public sync path, verifies the expected `ValueError`, and checks the exact mosaic resolution request path.

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Added a non-dict mosaic resolution page test. It exercises the public sync path, verifies the expected `ValueError`, and checks the exact mosaic resolution request path.

explorer/puller/puller/model/symbol/Resolution.py
daoka: You're right. I rechecked this against the current `symbol/symbol` REST/Rosetta implementation instead of the deprecated TypeScript SDK. The selection is now simplified to scan the ordered entries in reverse and return the first entry whose source is `<= (primary_id, secondary_id)`, or `None` when no entry applies.

explorer/puller/puller/model/symbol/Resolution.py
daoka: You're right that the single-entry case still needs a source comparison. I initially followed the unconditional single-entry shortcut from the deprecated TypeScript SDK. I removed that shortcut, so an entry is now selected only when its source is applicable to the transaction source.

explorer/puller/tests/model/symbol/test_Resolution.py
daoka: You're right; the expected value should be `B`. For transaction source `(2, 2)`, entry `(2, 0)` is the largest source that is less than or equal to it. I updated the implementation and test accordingly.

explorer/puller/tests/model/symbol/test_Resolution.py
daoka: You're right. The unconditional single-entry shortcut was incorrect. I removed it and updated the test so that a sole entry at `(4, 2)` does not apply to transaction source `(1, 0)` and returns `None`.

explorer/puller/puller/model/symbol/Resolution.py
Jaguar0625: please don't use that sdk as reference :)

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
```diff
@@ -199,3 +263,546 @@ def test_sync_block_headers_keeps_existing_watermark_when_transaction_fetch_fail
...
+			resolution_kind='mosaic')
+
+		# Act:
+		self._assert_sync_rejects_node_response(
+			connector,
+			ValueError,
+			'Malformed Symbol mosaic resolution page response')
+
+		# Assert:
+		self.assertEqual([resolution_path('mosaic', 1)], self._resolution_paths(connector))
```
Jaguar0625, line 814: not sure if you want to check resolution_path in address variants too?

Jaguar0625 (approved):
👍

explorer/puller/tests/facade/symbol/test_PullerTransactionSync.py
daoka: Good point. I added assertions for the exact address resolution request path to both the malformed and non-dict address response tests, matching the existing mosaic variants. All 26 tests in test_PullerTransactionSync.py pass.

PR #2411, opened 2026-07-20
[explorer/puller] fix: keep linked remote address across account refreshes

cryptoBeliever:
## Summary

Closes #2408

After an account key link, the explorer showed the linked remote account correctly, but the `remote_address` was wiped as soon as the main account did anything else (e.g. sent a transfer), even though the link was still active on-chain. The link also sometimes failed to appear right after the key link transaction.

## Root cause

The puller does not reconstruct account state per block - for every address touched by a block it fetches the node's **current** account snapshot and upserts it. `remote_address`, however, is only known when the *remote* account is processed (via a forwarded lookup), so a normal refresh of the main account always carried `remote_address = NULL`. Because `upsert_account` unconditionally wrote `remote_address = EXCLUDED.remote_address` on conflict, any plain refresh of the main account cleared the link. Address extraction also relies on a `set`, so within a batch the main and remote accounts were processed in a non-deterministic order - which is why the link occasionally failed to
appear immediately after linking.

## Why clear on `DEACTIVATING`

`remoteStatus` is a NIS `AccountMetaData` field with values `REMOTE / ACTIVATING / ACTIVE / DEACTIVATING / INACTIVE`. During `DEACTIVATING` the link is still active and the remote still harvests, so it still appears in blocks and forwards to the main account - this is the last reliable window to observe the teardown. Once `INACTIVE` the remote stops harvesting and may never appear in a block again, so clearing only on `INACTIVE` could leave a stale link.

### Additional benefit of refreshing the block beneficiary

The forwarded lookup always resolves the remote to its **current** main account, but a remote can be re-linked to a different main account over time. When processing a historical block we would resolve the *current* main and miss refreshing the account that was actually harvesting back then. The block `beneficiary`, by contrast, is block-accurate - it is the main account for that specific block - so adding it to the refresh set keeps previously-linked main accounts up to date as well.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -453,7 +453,7 @@ class NemDatabase(DatabaseConnection):
 			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
 			ON CONFLICT (address)
 			DO UPDATE SET
-				remote_address = EXCLUDED.remote_address,
+				-- remote_address is owned exclusively by update_account_remote_address;
 				importance = EXCLUDED.importance,
 				balance = EXCLUDED.balance,
 				vested_balance = EXCLUDED.vested_balance,
@@ -478,6 +478,23 @@ class NemDatabase(DatabaseConnection):
 				account_info.min_cosignatories,
 				[address.bytes for address in account_info.cosignatory_of] if len(account_info.cosignatory_of) > 0 else None,
 				[address.bytes for address in account_info.cosignatories] if len(account_info.cosignatories) > 0 else None,
+			)
+		)
+
+	@staticmethod
+	def update_account_remote_address(cursor, address, remote_address):
+		"""Sets or clears the remote (linked) address on a main account."""
+
+		cursor.execute(
+			'''
+			UPDATE accounts
+			SET remote_address = %s,
+				updated_at = CURRENT_TIMESTAMP
+			WHERE address = %s
+			''',
+			(
+				remote_address.bytes if remote_address else None,
+				address.bytes
 			)
 		)
 
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -453,7 +453,6 @@ class NemDatabase(DatabaseConnection):
 			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
 			ON CONFLICT (address)
 			DO UPDATE SET
-				-- remote_address is owned exclusively by update_account_remote_address;
 				importance = EXCLUDED.importance,
 				balance = EXCLUDED.balance,
 				vested_balance = EXCLUDED.vested_balance,
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -15,6 +15,8 @@ from zenlog import log
 
 from puller.db.NemDatabase import NemDatabase
 
+ACCOUNT_KEY_LINK_MODE_ACTIVATE = 1
+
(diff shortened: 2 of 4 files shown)


explorer/puller/puller/facade/NemPuller.py
```diff
@@ -313,14 +318,13 @@ async def _process_account_batch(self, cursor, address_heights):
 			self.nem_db.upsert_account(cursor, account)
```
Wayonb, line 316: can you move the upsert_account() below?  i.e. after the account remote_status is known so only need to update the DB once.

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -313,14 +318,13 @@ async def _process_account_batch(self, cursor, address_heights):
...
 
 			if 'REMOTE' == account_info.remote_status:
-				# Try fetching forwarded account info if remote status is REMOTE
 				main_account_info = await self._retry_get_account_info(address, forwarded=True)
-				main_account_mosaics = await self._retry_get_account_mosaics(str(main_account_info.address))
-
-				main_mosaics_json = self._convert_mosaics_to_json(main_account_mosaics)
-				main_account = self._create_account_record(main_account_info, main_mosaics_json, height, remote_address=account.address)
+				if main_account_info.address != account.address:
+					remote_address = None if 'DEACTIVATING' == main_account_info.remote_status else account.address
```
Wayonb: What gets returned by the API after the remote status is inactive?

Wayonb (changes requested):

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -453,7 +453,7 @@ def upsert_account(cursor, account_info):
 			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
 			ON CONFLICT (address)
 			DO UPDATE SET
-				remote_address = EXCLUDED.remote_address,
+				-- remote_address is owned exclusively by update_account_remote_address;
```
AnthonyLaw: We can remove this

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw, line 314: In `_create_account_record` method, it passed `remote_address=None` as the default.
I think it's safe to remove it.

explorer/puller/puller/facade/NemPuller.py
cryptoBeliever: We can't do that, because we only want to update remote_account on the main account when we're processing a REMOTE account. A batch may well contain only the main account without its remote - and then we have no way of knowing whether to clear the link, set it, or leave it alone. When we do have the remote, we call the endpoint, and we know. Given only the main account, we can't tell whether it has a remote or not: there's no API that returns the remote account for a given main account.

However I've changed the approach here. The puller now processes ACCOUNT_KEY_LINK transactions directly. That means we always catch the moment the status becomes ACTIVATING - which is by definition the same moment an ACCOUNT_KEY_LINK with link mode is included in a block - and that's where we set remote_account. Symmetrically, we clear it when an ACCOUNT_KEY_LINK with unlink mode is processed, which is equivalent to the account entering DEACTIVATING. Also we don't need to call API so it should be a little performance improvement.

explorer/puller/puller/facade/NemPuller.py
cryptoBeliever: DEACTIVATING is the last point at which we can catch the change, because that status changes the moment the link is removed by an Account Key Link transaction. After roughly 6 hours the account turns INACTIVE and /account/get/forwarded on the remote account no longer returns the main account - but no transaction is executed at that point either, so there's nothing for the puller to catch.

However, I changed the approach as described above.

Wayonb (approved):
👍🏾

AnthonyLaw (approved):
👍🏼

PR #2414, opened 2026-07-21
[explorer/puller] feat: persist Symbol namespace current state

daoka:

Added Symbol namespace current-state synchronization to the puller.

### Persistence

- Added `symbol_namespaces` for namespace hierarchy, ownership, duration, registration type, alias state, raw payload, and observation height.
- Added `symbol_alias_names` for namespace, mosaic, and account artifact-to-name mappings.
- Added PostgreSQL enums, unique constraints, and indexes required by namespace filtering and registration-height ordering.
- Namespace upserts atomically replace their derived alias-name rows, preventing stale mosaic or account aliases.

### Normalization

- Derives the namespace ID from `level{depth - 1}` and preserves root and immediate-parent relationships.
- Supports one-, two-, and three-level namespace names.
- Maps the unlimited end-height sentinel to `NULL`.
- Converts registration and alias numeric values to documented enum labels.
- Maps address aliases to `account` alias-name artifacts.
- Preserves the complete node response in `raw_payload`.

### Synchronization

Dirty namespace IDs are collected from:

- namespace registration transactions;
- address and mosaic alias link/unlink transactions;
- embedded namespace registration transactions;
- namespace-expired and namespace-deleted receipts.

IDs are deduplicated in first-encounter order.

Namespace details are fetched before any batch write, using bounded concurrency in chunks of 100 while preserving input order. Namespace-name requests are also chunked at 100 unique level IDs and support duplicate and ancestor entries returned by the node.

Namespaces found on the node are upserted. A `ResourceNotFound` response deletes the corresponding namespace and its derived alias-name rows. Other fetch or normalization failures abort before blocks, accounts, or namespaces are written.

### Rollback handling

Rollback repair re-fetches namespaces observed at or above the fork height before starting the repair transaction.

The repair then atomically:

1. removes chain and account state affected by the rollback;
2. refreshes namespaces still present on the node;
3. deletes namespaces confirmed missing by the node;
4. updates synchronization state.

Namespaces are refreshed instead of being deleted unconditionally because they do not have a broad re-dirty signal. A valid namespace might never be touched again after a fork and could otherwise remain permanently absent from the Explorer.

If namespace fetching fails, the repair transaction is not started, allowing the next run to detect and retry the same rollback.

### Scope

- Limited to the Symbol puller and its tests.
- No REST implementation.
- No NEM-specific production changes.
- No schema migration or compatibility shim.
- Transaction address/mosaic namespace-alias resolution remains an independent change.


explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -137,7 +138,724 @@ def _insert_account_refresh_snapshot_rows(database, refresh_entry):
 	database.upsert_account_refresh_page([refresh_entry], last_scanned_page=1)
 
 
+def _create_namespace_row(namespace_id='A95F1F8A96159516', full_name='root', observed_height=10, **overrides):
```
Jaguar0625: consider searching and replacing with constants in SymbolNamespaceTestUtils

Jaguar0625: should you replace this with create_expected_root_namespace_row ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -137,7 +138,724 @@ def _insert_account_refresh_snapshot_rows(database, refresh_entry):
...
+					ADDRESS1.lower(), 1, None, 'mosaic', 'mosaic-original', None,
+					{'namespace': {'state': 'original'}}, 10)
+			], [
+				('mosaic', 'mosaic-original', 'root', 10),
+				('namespace', 'A95F1F8A96159516', 'root', 10),
+				('account', 'account-original', 'root', 10)
+			]),
+			fetch_namespace_state(database.connection))
+
+	def test_apply_namespace_entries_preserves_mixed_upsert_and_delete_input_order(self):
```
Jaguar0625, line 416: test is fine, but can you explain if and how this can happen in production?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -137,7 +138,724 @@ def _insert_account_refresh_snapshot_rows(database, refresh_entry):
...
+					'artifact_id': 'invalid',
+					'name': 'invalid',
+					'updated_at_height': 2
+				}]}
+			])
+
+		# Assert:
+		self.assertEqual(original_state, fetch_namespace_state(database.connection))
+
+	def test_upsert_namespace_updates_every_non_key_column_while_retaining_namespace_key(self):
```
Jaguar0625: by namespace key are you referring to namespace_id? i am not sure what this test is trying to test?

explorer/puller/tests/model/symbol/test_Namespace.py
```diff
@@ -0,0 +1,180 @@
...
+
+	def test_create_alias_name_rows_returns_only_namespace_name_when_alias_is_none(self):
+		# Arrange:
+		row = create_namespace_row(create_namespace_item(), {ROOT_ID: 'root'}, 123)
+
+		# Act:
+		alias_rows = create_alias_name_rows(row)
+
+		# Assert:
+		self.assertEqual([{
```
Jaguar0625, line 138: what is the point of storing artifact_namespace when there is no alias? how do you expect to use this?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -646,6 +666,80 @@ def _collect_dirty_addresses_for_batch(  # pylint: disable=too-many-branches
 
 		return dirty_addresses
 
+	@staticmethod
+	def _collect_dirty_namespace_ids_for_batch(transaction_rows_by_height, receipt_rows_by_height):
+		"""Collects deduplicated namespace ids whose current state may have changed in a synced batch, in first-encounter order."""
+
+		dirty_namespace_ids = {}
```
Jaguar0625, line 673: any reason not to use set() ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -646,6 +666,80 @@ def _collect_dirty_addresses_for_batch(  # pylint: disable=too-many-branches
...
+
+		for receipt_rows in receipt_rows_by_height.values():
+			for receipt_row in receipt_rows:
+				if receipt_row['receipt_type'] in (NAMESPACE_EXPIRED_RECEIPT_TYPE, NAMESPACE_DELETED_RECEIPT_TYPE):
+					dirty_namespace_ids[receipt_row['artifact_id']] = None
+
+		return list(dirty_namespace_ids)
+
+	def _expand_dirty_namespace_ids(self, direct_namespace_ids):
+		"""Adds known descendants of directly dirty roots in stable order."""
```
Jaguar0625: why do you need stable order?

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
```diff
@@ -0,0 +1,603 @@
...
+					1,
+					transaction_hash='B' * 64,
+					transaction_id='address-alias-link',
+					type=TransactionType.ADDRESS_ALIAS.value,
+					namespaceId=NAMESPACE_ROOT_ID,
+					address=BENEFICIARY_ADDRESS,
+					aliasAction=AliasAction.LINK.value)
+			],
+			{NAMESPACE_ROOT_ID: current_item},
+			{NAMESPACE_ROOT_ID: 'root'})
```
Jaguar0625, line 194: shouldn't this delete root?

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
```diff
@@ -0,0 +1,603 @@
...
+			('mosaic', self.COMPARISON_MOSAIC_ID, 'other', deletion_height - 2),
+			('namespace', self.COMPARISON_NAMESPACE_ID, 'other', deletion_height - 2)
+		], alias_rows)
+		self.assertEqual(list(range(1, deletion_height + 1)), block_heights)
+		self.assertEqual(deletion_height, sync_state['last_synced_height'])
+		self._assert_namespace_requests(
+			connector, [NAMESPACE_ROOT_ID, NAMESPACE_SUB_ID, NAMESPACE_SUB_SUB_ID], [])
+		self._assert_namespace_sync_requests(connector, deletion_height, deletion_height)
+
+	def test_sync_block_headers_extends_namespace_end_height_when_renewed_during_grace_period(self):
```
Jaguar0625, line 339: do you want to consider having one of the renewed namespaces have an alias too ?

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
```diff
@@ -0,0 +1,603 @@
...
+			('namespace', NAMESPACE_SUB_ID, 'root.child', renewal_height)
+		], alias_rows)
+		self.assertEqual(list(range(1, renewal_height + 1)), block_heights)
+		self.assertEqual(renewal_height, sync_state['last_synced_height'])
+		self._assert_namespace_requests(
+			connector, [NAMESPACE_ROOT_ID, NAMESPACE_SUB_ID, NAMESPACE_SUB_SUB_ID],
+			[{'namespaceIds': [NAMESPACE_ROOT_ID, NAMESPACE_SUB_ID, NAMESPACE_SUB_SUB_ID]}])
+		self._assert_namespace_sync_requests(connector, renewal_height, renewal_height)
+
+	def test_sync_block_headers_fetches_root_and_known_descendants_once_when_child_is_directly_dirty(self):
```
Jaguar0625: unclear from test name whether you want to test:
1. full namespace tree is pulled when single child is added/removed
2. only single namespace tree is created when multiple parts of tree are modified

Jaguar0625: i assume latter, so in that case would rename test

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
```diff
@@ -0,0 +1,603 @@
...
+		seed_namespace(
+			self.puller.symbol_db,
+			grandchild_item,
+			{NAMESPACE_ROOT_ID: 'root', NAMESPACE_SUB_ID: 'child', NAMESPACE_SUB_SUB_ID: 'grandchild'})
+		connector = self._create_namespace_sync_connector(
+			[
+				create_node_transaction(
+					1,
+					transaction_hash='A' * 64,
+					transaction_id='root-renewal',
```
Jaguar0625: i suppose this is technically a renewal, but i found naming misleading in tests where specific renewal behavior isn't being tested

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
```diff
@@ -0,0 +1,603 @@
...
+			[{'namespaceIds': [NAMESPACE_ROOT_ID, NAMESPACE_SUB_ID, NAMESPACE_SUB_SUB_ID]}])
+		namespace_rows, alias_rows = fetch_namespace_state(self.puller.symbol_db.connection)
+		self.assertEqual([
+			(NAMESPACE_ROOT_ID, None, NAMESPACE_ROOT_ID, 'root', 'root', 1, 'root',
+				BENEFICIARY_ADDRESS.lower(), 1, None, 'none', None, None, root_item, 1),
+			(NAMESPACE_SUB_SUB_ID, NAMESPACE_SUB_ID, NAMESPACE_ROOT_ID, 'grandchild', 'root.child.grandchild', 3, 'child',
+				BENEFICIARY_ADDRESS.lower(), 1, None, 'none', None, None, grandchild_item, 1),
+			(NAMESPACE_SUB_ID, NAMESPACE_ROOT_ID, NAMESPACE_ROOT_ID, 'child', 'root.child', 2, 'child',
+				BENEFICIARY_ADDRESS.lower(), 1, None, 'mosaic', NATIVE_MOSAIC_ID, None, child_item, 1)
+		], namespace_rows)
```
Jaguar0625: are you testing this order ?

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -20,6 +30,247 @@
...
+			2)
+		seed_namespace(
+			self.puller.symbol_db,
+			grandchild_before_item,
+			{NAMESPACE_ROOT_ID: 'root', NAMESPACE_SUB_ID: 'child', NAMESPACE_SUB_SUB_ID: 'grandchild'},
+			2)
+		seed_namespace(
+			self.puller.symbol_db,
+			comparison_item,
+			{'B95F1F8A96159516': 'before-fork'},
```
Jaguar0625: why is this named 'before-fork' isn't there only one?

Jaguar0625: isn't this test testing the 3-tier hierarchy reverts to *before_item (via end_height)?

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -20,6 +30,247 @@
...
+				NAMESPACE_ROOT_ID,
+				NAMESPACE_SUB_ID,
+				level_ids=[NAMESPACE_ROOT_ID, NAMESPACE_SUB_ID, NAMESPACE_SUB_SUB_ID]),
+			{NAMESPACE_ROOT_ID: 'root', NAMESPACE_SUB_ID: 'child', NAMESPACE_SUB_SUB_ID: 'grandchild'},
+			2)
+		connector = FakeConnector(
+			3,
+			{1: [create_node_block(2), create_node_block(3)]},
+			{2: create_node_block(2)},
+			namespace_by_id={NAMESPACE_ROOT_ID: create_namespace_item()},
```
Jaguar0625: children are deleted because this only returns info for root namespace?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -1,30 +1,398 @@
...
+			[NAMESPACE_ROOT_ID],
+			[{'namespaceIds': [NAMESPACE_ROOT_ID]}])
+
+	def test_sync_block_headers_deletes_dirty_namespace_state_when_namespace_is_not_found(self):
+		# Arrange:
+		seed_namespace(self.puller.symbol_db, create_namespace_item(), {NAMESPACE_ROOT_ID: 'root'})
+		connector = FakeConnector(
+			1,
+			{0: [create_node_block(1)]},
+			transactions_by_path={
```
Jaguar0625, line 110: not found because no namespace_by_id specified here ?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -1,30 +1,398 @@
...
+			]}},
+			namespace_by_id={
+				NAMESPACE_ROOT_ID: create_namespace_item(),
+				NAMESPACE_SUB_ID: create_namespace_item(namespace_id=NAMESPACE_SUB_ID, root_id=NAMESPACE_SUB_ID)
+			},
+			namespace_names={NAMESPACE_ROOT_ID: 'shared', NAMESPACE_SUB_ID: 'shared'})
+
+		# Act:
+		with self.assertRaises(PsycopgError):
+			self._sync_with_connector(connector)
```
Jaguar0625, line 192: why does write fail ?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -1,30 +1,398 @@
...
+		self.puller.symbol_db.connection.commit()
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(first_namespace_state, fetch_namespace_state(self.puller.symbol_db.connection))
+		self._assert_namespace_requests(
+			connector,
+			[NAMESPACE_ROOT_ID, NAMESPACE_ROOT_ID],
```
Jaguar0625: two because there are two syncs?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -1,30 +1,398 @@
...
+				namespace_id: create_namespace_item(namespace_id=namespace_id, root_id=namespace_id)
+				for namespace_id in namespace_ids
+			},
+			namespace_names={namespace_id: f'name-{namespace_id}' for namespace_id in namespace_ids})
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self._assert_namespace_requests(connector, namespace_ids, [{'namespaceIds': namespace_ids}])
```
Jaguar0625, line 342: i believe these are sorted? if you truly want to test order by appearance, you'll want the expected to not be sorted.

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -1,30 +1,398 @@
...
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute('SELECT name, full_name FROM symbol_namespaces')
+		self.assertEqual([('sub', 'root.sub')], cursor.fetchall())
+		self._assert_namespace_requests(
+			connector,
+			[NAMESPACE_SUB_ID],
```
Jaguar0625, line 373: why is NAMESPACE_ROOT_ID not looked up?

Jaguar0625 (commented):
👍

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: > consider searching and replacing with constants in SymbolNamespaceTestUtils

Thanks. I replaced the duplicated namespace IDs with the shared `NAMESPACE_ROOT_ID`, `NAMESPACE_SUB_ID`, and `NAMESPACE_SUB_SUB_ID` constants.

> should you replace this with create_expected_root_namespace_row ?

 `_create_namespace_row` now delegates to `create_expected_root_namespace_row` while preserving the DB-test-specific defaults and caller overrides.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Thanks. I renamed the test so it explicitly states that all non-key columns are updated while `namespace_id` is retained.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: The current production caller does not generate mixed entries for the same namespace ID. I kept this as a defensive test of apply_namespace_entries' input-order contract, where later entries intentionally determine the final state.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thanks. I updated the docstring to describe the exact contract: direct dirty IDs remain in first-encounter order, followed by known descendants of dirty roots.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: Thanks. The renewal fixture now assigns a mosaic alias to the renewed child namespace and fully verifies the namespace row and corresponding alias-name rows.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: Thanks. I changed the transaction ID to `root-registration` in tests that do not verify renewal-specific behavior. The dedicated renewal fixture remains unchanged.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: Thanks. I limited this test to namespace detail request order, deduplication, names POST payloads, and the normal block/transaction/statement requests. Namespace persistence remains covered by the lifecycle tests.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Thanks. The test now separates the namespace detail and names requests recorded after the initial sync and after the restarted sync, asserting one request per run.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Thanks. I changed the namespace IDs to a deliberately non-sorted valid hexadecimal sequence while retaining duplicate-source coverage and exact GET/POST order assertions.

explorer/puller/tests/model/symbol/test_Namespace.py
daoka: `artifact_type='namespace'` is not an alias-link row. It is the canonical namespace ID → full-name lookup entry. The `mosaic` and `account` rows are conditional alias-target mappings, so an unaliased namespace still keeps its `namespace` row. Unlinking an alias removes only the target mapping, while deleting the namespace removes all rows for its full name. This PR establishes the puller-side persistence contract; the REST consumer will follow in the namespace REST PR.

explorer/puller/puller/facade/SymbolPuller.py
daoka: A dict is used as an insertion-ordered set here. We need both deduplication and first-encounter order because the resulting order is preserved through root-descendant expansion and namespace detail requests. A plain set would lose that order, while sorted(set(...)) would introduce namespace-ID ordering instead of preserving the input order.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: No. `AliasAction.UNLINK` removes only the alias relationship; it does not delete the root namespace registration. The node still returns the root namespace with `alias.type = 0`, so the puller upserts it as `alias_type='none'` and removes only the derived mosaic alias row. The namespace itself is deleted only when `GET /namespaces/{id}` returns `ResourceNotFound`.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: You're right; this test covers the latter case. The fixture contains both a root registration and a child alias transaction, so both the root and child are directly dirty. Expanding the dirty root also discovers the child and grandchild, and first-seen deduplication ensures each namespace ID is fetched only once. I renamed the test to test_sync_block_headers_fetches_each_known_tree_namespace_once_when_root_and_child_are_directly_dirty; the fixture and assertions are unchanged.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: The original naming was ambiguous. The `*_before_item` rows were not the canonical pre-fork state; they represented orphaned-fork state already persisted in the database before rollback repair. The `*_updated_item` rows were the canonical state returned by the node and applied during repair. I renamed them to `orphaned_*_item` and `canonical_*_item`, and renamed the comparison namespace to `unaffected_item`. The test still verifies that all three hierarchy levels are refreshed to the canonical `end_height`, payload, and aliases with `updated_at_height = fork_height - 1`, while the unaffected namespace is preserved.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Yes. The root still exists on the canonical node, while the child and grandchild no longer exist. The puller requests all three namespace IDs individually: the root response becomes an upsert entry, and the explicit `ResourceNotFound` responses for the child and grandchild become delete entries. I updated the fixture to map both descendant responses explicitly instead of relying on FakeConnector's default not-found behavior.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Yes. Previously, omitting `namespace_by_id` caused FakeConnector to return its default `ResourceNotFound` response for the namespace. I made the response explicit in the fixture so the deletion condition is visible in the Arrange section. The existing assertions for deleting both the namespace and alias-name rows, while completing the block sync, are unchanged.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: The original intent was to verify that a later namespace write failure rolls back all namespace entries.
I confirmed that different namespace IDs cannot have the same name in valid Symbol data because the ID is derived from the parent ID and name. I updated the test to use SDK-generated IDs and valid Node data, then trigger the unique constraint failure with a stale local Explorer DB row.
The test still verifies namespace rollback, block preservation, and that the sync state is not advanced.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Thanks. I added a comment clarifying that only the child is dirty and that the root name needed for `root.child` comes from the ancestor entry returned by `POST /namespaces/names`.

explorer/puller/tests/db/test_SymbolDatabase.py
Jaguar0625: please add a comment in the test

explorer/puller/tests/db/test_SymbolDatabase.py
Jaguar0625: can you explain the purpose of the test?

explorer/puller/tests/model/symbol/test_Namespace.py
Jaguar0625: > It is the canonical namespace ID → full-name lookup entry

so, intent is for this table to be be queried in order to resolve a namespace id to a full qualified name ?

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: thank you. i understand *what* you are doing, but i'm not sure why you are ordering dirty namespaces in this way vs unordered or sorted, for example?

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
Jaguar0625: i was asking what exactly you are verifying in this test ?

explorer/puller/tests/facade/symbol/test_PullerRollback.py
Jaguar0625: thank you, much clearer now 👍

Jaguar0625 (commented):
👍

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Thanks. The production caller currently emits at most one entry per namespace ID. This test defensively verifies that `apply_namespace_entries` processes mixed duplicate entries in input order, with the later entry determining the final state. I added this explanation as a comment in the test and kept the fixture and assertions unchanged.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: The test guards the completeness of the `ON CONFLICT DO UPDATE SET` column list. It first inserts a namespace row, then upserts the same `namespace_id` with a different value for every non-key column. This makes any omitted update column observable while confirming that the primary key remains unchanged. I added a comment explaining this purpose and kept the existing full-row assertion.

explorer/puller/tests/model/symbol/test_Namespace.py
daoka: Yes. For `artifact_type='namespace'`, `artifact_id` is the namespace ID and `name` is its fully qualified name. This row supports namespace ID → full-name lookup for display and batch name resolution, independently of whether the namespace currently has an alias link.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Namespace state is order-independent, but preserving first-encounter order carries the transaction and receipt scan order through root expansion and namespace-detail request chunking. An unordered set would make request order and chunk composition implementation-dependent, while sorting would impose an unrelated namespace-ID priority. I added this rationale to the `_expand_dirty_namespace_ids` docstring and kept the algorithm unchanged.

explorer/puller/tests/facade/symbol/test_PullerNamespaces.py
daoka: This test verifies that direct dirty namespace IDs retain their first-encounter order during root expansion. When the child is encountered before the root, expanding the root rediscovers that child and discovers the grandchild. The existing child must remain in its original position without duplication, while only the newly discovered grandchild is appended, producing [child, root, grandchild].

Jaguar0625 (approved):
👍

PR #2417, opened 2026-07-21
[bridge/reporting] feat: Added bridge API and utils 

AnthonyLaw:
Problem: Missing bridge request to endpoint.
Solution: 
- Added `makeGetRequest` in utils to request data.
- Added  `parseSearchInput` in utils to handle search input.
- Added bridge endpoints, which support `wraps | unwrap` response.


bridge/reporting/api/bridge.js
```diff
@@ -0,0 +1,123 @@
+import config from '@/config';
+import { PAGE_SIZE } from '@/constants';
+import { makeGetRequest } from '@/utils/server';
+import { parseSearchInput } from '@/utils/validation';
+
+/**
+ * Criteria used to query a bridge report endpoint.
+ * @typedef {Object} ReportCriteria
+ * @property {string} baseUrl Bridge API base URL.
+ * @property {'wrap'|'unwrap'} operation Bridge operation.
```
Wayonb, line 10: How do these modes map to the backend?

bridge/reporting/api/bridge.js
```diff
@@ -0,0 +1,123 @@
...
+};
+
+/**
+ * Fetches configuration metadata for the wrapped and native bridges concurrently.
+ * A bridge configuration is null when its request fails.
+ * @returns {Promise<{wrapped: Object|null, native: Object|null}>} Configuration metadata for each bridge.
+ */
+export const fetchBridgeConfiguration = async () => {
+	const wrappedUrl = config.PUBLIC_BRIDGE_WRAPPED_URL;
+	const nativeUrl = config.PUBLIC_BRIDGE_NATIVE_URL;
```
Wayonb, line 72: Here, using wrap and native does not match the wrap and unwrap defined earlier. Should keep consistent for readability.

bridge/reporting/api/bridge.js
```diff
@@ -0,0 +1,123 @@
...
+ */
+export const fetchAllReportRows = async (criteria, onProgress) => {
+	const limit = criteria.limit || PAGE_SIZE;
+	let offset = 0;
+	let rows = [];
+	let hasMore = true;
+
+	while (hasMore) {
+		const { data, hasMore: pageHasMore, nextOffset } = await fetchReportPage({ ...criteria, limit, offset });
+		rows = [...rows, ...data];
```
Wayonb, line 116: Why would you need to fetch all instead of just paging?

bridge/reporting/setupTests.js
```diff
@@ -0,0 +1,8 @@
+const appConfig = {
+	PUBLIC_BRIDGE_WRAPPED_URL: 'https://bridge.example/wrapped',
+	PUBLIC_BRIDGE_NATIVE_URL: 'https://bridge.example/native',
+	PUBLIC_REQUEST_TIMEOUT: 5000
+};
+
+window.appConfig = appConfig;
+Object.assign(process.env, appConfig);
```
Wayonb: newline?

bridge/reporting/utils/validation.js
```diff
@@ -0,0 +1,34 @@
...
+ * @returns {ParsedSearchInput|null} Parsed input, or `null` when the input is invalid.
+ */
+export const parseSearchInput = input => {
+	const value = (input || '').trim();
+
+	if (!value)
+		return { type: null, value: '' };
+
+	if (TRANSACTION_HASH_PATTERN.test(value))
+		return { type: 'hash', value: value.replace(/^0x/i, '').toUpperCase() };
```
Wayonb, line 24: what happens if its starts wtih ``0X``?

Wayonb (changes requested):
👍🏾

bridge/reporting/api/bridge.js
AnthonyLaw: This method is actually used to export all records to CSV.

bridge/reporting/utils/validation.js
AnthonyLaw: It'll respond `null`.
It should cover `0X` as well. Let me update it.

bridge/reporting/api/bridge.js
AnthonyLaw: These are not bridge strategy modes (wrap, stake, or swap). 
The operation and resource fields map directly to backend endpoint paths:
```
wrap + requests → /wrap/requests
unwrap + requests → /unwrap/requests
wrap + errors → /wrap/errors
unwrap + errors → /unwrap/errors
```

baseUrl selects the bridge API instance, while `operation` and `resource` select its endpoint. 
The **wrapped** endpoint bridge supports both directions, whereas the **native** endpoint bridge supports only wrap

bridge/reporting/api/bridge.js
AnthonyLaw: The wrapped URL and native URL are different from `wrap` and `unwrap`

currently we have something like this in our bridge endpoint
```
ethereum-wrapped/wrap
ethereum-wrapped/unwrap

ethereum-native/wrap
```

Wayonb (approved):
👍🏾

PR #2418, opened 2026-07-22
[explorer] feat: base harvesting accounts stat and filter on recent block activity

cryptoBeliever:

The `/accounts` page reports harvesting in two places, each computed differently:

  - "Harvesting Accounts" stat - `COUNT(*) FILTER (WHERE harvested_blocks > 0)`, i.e. every account that has ever harvested a block.
  - "Active Harvesting" filter - `WHERE remote_status = 'ACTIVE'`, i.e. every account that has harvesting configured, whether or not it ever produced a block.

The donut chart and percentage next to the stat are derived from the same lifetime count.

## What's the issue?

Neither measure answers the question a reader actually has: who is harvesting right now?

- Lifetime history never decays. An account that harvested once and went dark keeps inflating the stat indefinitely.
- Configured status is not activity. A delegated key can be linked on an account that has never produced a block.
- The two definitions disagree, so the stat and the filter return unrelated populations on the same page.

Fixing one without the other would leave the page internally inconsistent, so both are changed together.

Separately, the per-row "Harvesting Activated" label was always `false`: the frontend read `data.harvestRemoteStatus` while REST returns `remoteStatus`. Left alone, filtering by "Active Harvesting" would return a list in which every row is labelled inactive.

Single definition. `rest/db/NemHarvestingActivity.py` defines the window: an account is an active harvester when it harvested at least one block within the last 7 days, expressed as a block height range (10080 blocks at the 60s target block time) so the result does not depend on when the query runs. The cutoff is clamped to height 1 for chains shorter than the window.

Applied consistently. The statistic, the filter and the new per-account field all use the same predicate agains `accounts.last_harvested_height`, which the puller already maintains per beneficiary. Accounts that never harvested keep `0` and never match.

Query shape. The chain height is read once and the cutoff derived in Python, then bound as a query parameter instead of being computed in SQL. psycopg2 interpolates client-side, so the planner sees a constant and can use the index rather than a sequential scan:

Schema. Adds `accounts_last_harvested_height_idx` to the puller's NEM schema.

API. The account DTO gains `isHarvestingActive`, computed with the same predicate. The frontend now reads it instead of the field name that never matched, which fixes the always-`false` label.

The new definition is not self-evident from a bare number, so every place the flag surfaces now explains it


explorer/frontend/__tests__/test-utils/accounts.js
```diff
@@ -5,7 +5,7 @@ export const accountPageResponse = [
 		balance: 2014883839.88001,
 		cosignatories: null,
 		cosignatoryOf: null,
-		harvestRemoteStatus: 'INACTIVE',
+		isHarvestingActive: true,
```
Wayonb, line 8: should this be false?

explorer/rest/rest/db/NemHarvestingActivity.py
```diff
@@ -0,0 +1,25 @@
+"""Definition of the 'recently harvesting' account window used by the NEM explorer."""
+
+HARVESTING_ACTIVE_WINDOW_DAYS = 7
```
Wayonb: Why 7 Days was picked?  
In the mainnet, do we know how long an account with 10K takes to harvest?

Wayonb: might want to make ``HARVESTING_ACTIVE_WINDOW_DAYS`` a config value
also maybe I missed it but is the number of days shown in the accounts title/description?  I am sure some node owners might ask why their node is not displayed 😅

Wayonb (changes requested):
Looks good👍🏾

explorer/rest/rest/db/NemDatabase.py
```diff
@@ -468,14 +472,23 @@ def _generate_transaction_sql_query(where_condition='', order_condition='', limi
 			{limit_condition}
 		'''
 
+	@staticmethod
+	def _read_harvesting_active_cutoff_height(cursor):
+		"""Reads the chain height and converts it into the lowest height that still counts as recent harvesting."""
+
+		cursor.execute('SELECT COALESCE(MAX(height), 0) FROM blocks')
```
AnthonyLaw: I see you’re calculating the number of blocks in 7 days.
I have another alternative solution, but I’m not sure if it’s the right approach.
Since the blocks table contains a timestamp, maybe we can do something like this?

```
SELECT COALESCE(MIN(height), 1)
FROM blocks
WHERE timestamp >= (
	SELECT MAX(timestamp) - INTERVAL '7 days'
	FROM blocks
)
```

This way, we can get the cutoff height directly from the block timestamps.

explorer/rest/rest/db/NemHarvestingActivity.py
cryptoBeliever: Yes, we do have tooltips on the frontend explaining, for example:
> Number of accounts that harvested at least one block within the last 7 days. Accounts that only have harvesting configured, or whose last harvested block was before this period, are not counted.

I just checked how it's implemented in Symbol now. I'm wondering which direction we should take now, because I noticed that Daoka implemented it a bit differently.

**NEM (this PR)** stores no flag. Instead, "recently harvesting" is computed at query time using a single predicate:

last_harvested_height >= cutoff

where:

cutoff = chain_height - 10080 blocks (7 days).

The same predicate drives the statistic, the filter, and the per-account isHarvestingActive value, so they cannot diverge. It's always up to date, requires no refresh job, and is index-backed.

**Symbol (dev)** materializes a boolean is_harvesting_active field on the account (with the window hardcoded to 7 days), which is updated in two places:

sync-block (incremental) - sets the flag to true for beneficiaries in the processed batch, but never clears it.
refresh-accounts (full workflow, triggered by an external scheduler) – rescans the entire account population and recomputes the flag. This is the only path that can expire it (true -> false).

**Question**

Is it worth aligning Symbol with NEM's query-time approach for consistency, or is it acceptable for the two chains to differ here?

Frontend configuration (if NEM keeps the query-time approach in REST)

The tooltips on the frontend display the window length literally, so the frontend also needs to know that value. Should we expose it as a parameter as well?

**The options I see are:**

a) Hardcode it in the frontend (simple, but it can drift from REST/Puller).
b) Add a separate frontend configuration parameter.
c) Expose it from REST to the frontend (e.g. via /account/statistics) so there's a single source of truth.

However, option (c) only works while REST owns the window logic. If we later move the logic into the puller or sync-account, REST would simply expose a stored flag, and (c) would no longer work.

So the choice between a/b/c is coupled to the broader architectural decision: keeping the logic query-time in REST keeps (c) viable, whereas moving it into the puller eliminates that option.

Which direction do you prefer?

Wayonb: > refresh-accounts (full workflow, triggered by an external scheduler)

How often does this workflow run?

> Is it worth aligning Symbol with NEM's query-time approach for consistency, or is it acceptable for the two chains to differ here?

Usually I would go with consistency, and this does not mean that you need to switch to using the Symbol implementation. 😄 
1. The Symbol implementation will have slightly better performance on queries, but the data returned might not always be updated.
2. The Nem implementation is always correct at the time of the query.

I would go with yours unless there is an issue with performance, which I think should not be a problem since the calculated field is just a comparison.

> The tooltips on the frontend display the window length literally, so the frontend also needs to know that value. Should we expose it as a parameter as well?

Yes, the whole point of making this value configurable is so that we can experiment in mainnet without the need to rebuild the code.

a) The drift can happen currently if someone updates the puller and not the frontend
b) This is better than current impl since all we need to do is restart the frontend to update
c) This is the best since the config lives in one place.

I think either b/c is good 👍🏾

cryptoBeliever: The account workflow runs once a day, so I decided to stick with my original solution in the query instead of rewriting it to a Symbol-style solution.

I also changed the parameter to 60 days. Based on what I was able to determine, this is an approximate harvesting period for an account holding 10,000–20,000 XEM.

I added an option to parametrize. The parameter can be configured in the REST API and in the frontend (it should be consistent). So it's option **B** because Symbol stores it in the puller anyway, so it would still need to be maintained separately from the REST API/frontend. As an additional benefit, we avoid an extra request on the Explorer views, since otherwise they would have to query the REST API to retrieve the parameter value.

explorer/frontend/__tests__/test-utils/accounts.js
cryptoBeliever: That's OK. I've changed the corresponding `isHarvestingActive` value in `accountPageResult` to true. Previously, all values were false, so we hadn't tested the true case.

explorer/rest/rest/db/NemDatabase.py
cryptoBeliever: Good call - adopted, with one change to the SQL.
```sql
  SELECT COALESCE((
      SELECT height
      FROM blocks
      WHERE timestamp >= (SELECT MAX(timestamp) FROM blocks) - make_interval(days => %s)
      ORDER BY timestamp
      LIMIT 1
  ), 1)
```
The old version had to convert days into blocks, which meant asserting a block rate in explorer code (`BLOCK_TARGET_SECONDS = 60`). The chain doesn't honour it exactly: the 86,400 blocks the code treated as "60 days" actually span 60 days 12:34:26. Expressing the window in days removes the conversion, so there's no calibration constant left to drift. `NemHarvestingActivity.py` and its test were removed for that reason.

AnthonyLaw (approved):
It look good for me 👍🏼

Wayonb (approved):
👍🏾

PR #2440, opened 2026-07-27
[explorer/puller] feat: persist Symbol mosaic current state

daoka:
## Summary

Added Symbol mosaic current-state synchronization to the puller.

- Added `symbol_mosaics` persistence with normalized node data, required
  indexes, and database-derived alias names.
- Collects dirty mosaics from definitions, resolved supply changes, and
  expiry receipts, then fetches them through `POST /mosaics` in 100-ID
  batches before starting any writes.
- Applies namespace changes before mosaic changes so alias names remain
  current, including alias retargeting, unlinking, and deletion.
- Refreshes surviving mosaics and deletes fork-only mosaics inside the
  existing atomic rollback transaction.

This is a puller-only change with no REST, frontend, or NEM production changes.


explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -589,14 +642,73 @@ def _execute_upsert_namespace(cursor, namespace_row, alias_name_rows):
...
 			{**namespace_row, 'raw_payload': Json(namespace_row['raw_payload'])})
-		cursor.execute('DELETE FROM symbol_alias_names WHERE name = %s', (namespace_row['full_name'],))
+		cursor.execute(
+			'DELETE FROM symbol_alias_names WHERE name = %s RETURNING artifact_type, artifact_id',
+			(namespace_row['full_name'],))
+		mosaic_ids = {
+			artifact_id: None
+			for artifact_type, artifact_id in cursor.fetchall()
+			if 'mosaic' == artifact_type
+		}
```
Jaguar0625, line 652: is there reason this can't be a set?

explorer/puller/tests/test/SymbolMosaicTestUtils.py
```diff
@@ -0,0 +1,142 @@
...
+
+def create_expected_mosaic_row(
+	mosaic_item,
+	observed_height,
+	*,
+	expiration_height=UNSET,
+	supply_mutable=UNSET,
+	transferable=UNSET,
+	restrictable=UNSET,
+	revokable=UNSET
```
Jaguar0625, line 67: why did you want separate UNSET vs None ?

explorer/puller/tests/model/symbol/test_Receipt.py
```diff
@@ -69,6 +70,10 @@ def test_receipt_type_labels_maps_supported_types_to_documented_labels(self):
 		self.assertEqual('namespaceExpired', NAMESPACE_EXPIRED_RECEIPT_TYPE)
 		self.assertEqual('namespaceDeleted', NAMESPACE_DELETED_RECEIPT_TYPE)
 
+	def test_mosaic_expired_receipt_type_is_documented_label(self):
+		# Arrange / Act / Assert:
+		self.assertEqual('mosaicExpired', MOSAIC_EXPIRED_RECEIPT_TYPE)
```
Jaguar0625: is there reason you wanted separate test for this vs including in previous one ?

explorer/puller/tests/model/symbol/test_Mosaic.py
```diff
@@ -0,0 +1,136 @@
...
+
+	def test_create_mosaic_row_retains_raw_payload(self):
+		# Arrange:
+		item = create_mosaic_item(item_id='A1B2C3D4E5F6071829304A5B')
+
+		# Act:
+		row = create_mosaic_row(item, 123)
+
+		# Assert:
+		self.assertEqual(create_expected_mosaic_row(item, 123), row)
```
Jaguar0625: aren't these tests the same ?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -73,6 +85,253 @@ def _create_namespace_batch_connector():
...
+			connector_arguments.update({
+				'namespace_by_id': {NAMESPACE_ROOT_ID: create_namespace_item(alias={'type': 1, 'mosaicId': MOSAIC_ID})},
+				'namespace_names': {NAMESPACE_ROOT_ID: 'root'}
+			})
+		if mosaics_response is not None:
+			connector_arguments['mosaics_response'] = mosaics_response
+
+		return FakeConnector(1, {0: [create_node_block(1)]}, **connector_arguments)
+
+	def test_sync_block_headers_persists_dirty_mosaic_definition(self):
```
Jaguar0625: test name strikes me as odd, maybe "persists new/discovered" ?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -73,6 +85,253 @@ def _create_namespace_batch_connector():
...
+					aggregate_hash,
+					0,
+					transaction_id='embedded-mosaic-supply-change',
+					type=TransactionType.MOSAIC_SUPPLY_CHANGE.value,
+					mosaicId=alias_mosaic_id,
+					delta=str(supply_delta),
+					action=MosaicSupplyChangeAction.INCREASE.value)
+			],
+			[
+				{'source': {'primaryId': 1, 'secondaryId': 0}, 'resolved': '0000000000000002'},
```
Jaguar0625: where does this one come from ?

Jaguar0625 (commented):
👍

explorer/puller/puller/db/SymbolDatabase.py
daoka: You're right; there is no ordering requirement here. I changed the database-side alias refresh tracking to use a set in both the namespace upsert and delete paths, and removed the unnecessary list conversions. The Puller dirty mosaic collector remains unchanged because its first-encounter order is part of its contract.

explorer/puller/tests/model/symbol/test_Receipt.py
daoka: There was no reason for this to be a separate test. I moved the `mosaicExpired` assertion into the existing documented-label test alongside the inflation and namespace receipt constants, and removed the standalone test.

explorer/puller/tests/model/symbol/test_Mosaic.py
daoka: You're right that the tests were too similar. The expiration test covers normalization of an unlimited duration to `None`, while the raw-payload test is intended to verify that the complete node response is retained without mutating the input. I renamed the latter and now build its full expected row from a `deepcopy` captured before Act, while retaining the non-default item ID and full-object comparison.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Agreed. I renamed it to `test_sync_block_headers_persists_mosaic_discovered_from_definition_transaction` to make it clear that the mosaic is newly discovered from a mosaic definition transaction. Existing `dirty` names remain where the tests operate on previously persisted mosaics.

explorer/puller/tests/test/SymbolMosaicTestUtils.py
daoka: `None` is a valid expected `expiration_height` for an unlimited mosaic, so it cannot also represent an unspecified expectation. `UNSET` distinguishes an omitted expectation from an explicitly expected `None`. The boolean flag expectations use the same sentinel for consistency. I documented this, along with the requirement to provide explicit expectations for non-default fixtures so the helper does not repeat production calculations and create tautological tests.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: This value is not derived from the transaction; it is an intentional decoy for the parent aggregate source `(1, 0)`. The embedded transaction at index 0 has source `(1, 1)`. Since `select_resolution_entry` selects the latest entry whose source is less than or equal to the transaction source, incorrectly using the parent source would select this decoy instead of `MOSAIC_ID`, causing the mosaic request and supply assertions to fail. I named it `mosaic_id_at_aggregate_source` and added a comment to make that purpose explicit.

Jaguar0625 (approved):
👍

PR #2448, opened 2026-07-27
[wallet/symbol/mobile] fix: add copy button to private key dialog on AccountDetails screen

OlegMakarenko:
## Problem:
- Not possible to copy private key from the Account Details screen.

## Solution:
- Added copy button to the private key dialog box.

Resolved: #2449


Wayonb (approved):
👍🏾

PR #2450, opened 2026-07-27
[wallet/symbol/mobile] feat: add cache system

OlegMakarenko:
## Problem:
- Many duplicated requests are fired to the network — no caching or deduplication exists anywhere in the request pipeline today.
- Events such as new-transaction confirmations fan out to multiple screens and widgets at once, so the same request can fire several times within about a second.

## Solution:
- Added a `RequestCache` class to the wallet common core project. It wraps the existing network request function, so no API service or screen code needs to change.
- Defined the caching policy — which URLs are cached, for how long, and under which invalidation scope — in a config file in the Symbol mobile wallet project.
- GET requests are deduplicated by default, so near-simultaneous duplicate fetches collapse into a single network call.
- Added a block listener that clears the relevant cache scope on every new block, with each entry's TTL acting as a fallback if that signal is ever missed.


Wayonb (approved):
👍🏾

PR #2451, opened 2026-07-27
[wallet/symbol/mobile] fix: improve account naming text

OlegMakarenko:
## Problem:
- "External Account" is used for two unrelated things: accounts added via private-key import, and the Ethereum-side accounts used for cross-chain bridge swaps. This is confusing to users.
- On the Actions screen, the "External Account" card is actually a shortcut into bridge account management — even though the screens it opens ("Your Bridge Accounts", "Bridge Account Details") already use "Bridge" terminology, not "External Account".

## Solution:
- Renamed the private-key import flow away from "External Account" terminology:
  - Button: "Add external account" → "Add existing private key"
  - Dialog/screen title: "External Account" → "Import Account"
  - Remove confirmation title: "Remove external account" → "Remove imported account"
- Renamed the Actions screen button from "External Account" to "Bridge Accounts", matching the "Your Bridge Accounts" / "Bridge Account Details" screens it already links to — no changes needed on that side.
- Scope is limited to user-facing copy; `WalletAccountType.EXTERNAL` / `addExternalAccount` internals are unchanged.

Resolved: #2399, #2420


Jaguar0625 (approved):
👍

PR #2452, opened 2026-07-27
[wallet/symbol/mobile] fix: Multisig Account Modification confirmation dialog text

OlegMakarenko:
## Problem:
- The multisig confirmation dialog always shows the title "Confirm Multisig Creation", even when the user is modifying an existing multisig account — this doesn't match the "Modify Multisig Account" screen and is misleading.
- The dialog body also reuses the create-specific text ("You are about to create a multisig account at %{address}...") during the modify flow, which is inaccurate since no new account is being created.
- Root cause: both the create and modify flows reuse the same localization key (`s_multisig_create_dialog_confirm_title`), which is only correct for the create flow.

## Solution:
- Added a separate modify-flow string for the confirmation dialog title and body, distinct from the create-flow strings.
- Updated the confirmation dialog to select the create vs. modify copy based on which flow triggered it.

Resolved: #2397


Wayonb (approved):
👍🏾

PR #2453, opened 2026-07-27
[wallet/symbol/mobile] fix: bridge related issues

OlegMakarenko:
## 1. Refund of unspent ETH (`refundETH` in the swap multicall)

Issue: #2446 

**The issue**
- When a swap sends native ETH and the pool cannot absorb the whole amount, the router spends only part of the ETH.
- The rest stays inside the router contract, where **anyone** can take it. It does not return to the user.

**What we did**
- Every native-ETH swap transaction now appends `refundETH()` as the **last** call of the router multicall.
- Leftover ETH returns to the sender in the same transaction. When there is no leftover, the call is a no-op (~1,700 gas).

**How others do it**
- Uniswap's own SDK adds the refund only for exact-output swaps or when price impact > 50% ([router-sdk swapRouter.ts](https://github.com/Uniswap/sdks/blob/main/sdks/router-sdk/src/swapRouter.ts)). That gate is provably too narrow: a plain exact-input swap on Arbitrum stranded **~11.3 ETH**, taken by the next user ([post-mortem](https://hackmd.io/@0x534154/Hy6_66gT5)); ~14+ ETH more was swept by bots over time ([analysis](https://techgeorgii.com/uniswaps-swaprouter-refund-bug-how-much-mev-bots-can-earn/)).
- Uniswap's web app has this exact bug **open since 2022** ([interface#3346](https://github.com/Uniswap/interface/issues/3346)); an open SDK issue proposes the unconditional refund we ship ([sdks#216](https://github.com/Uniswap/sdks/issues/216)).
- Uniswap's Universal Router SDK offers an opt-in `safeMode` that always sweeps — same behavior as ours, but off by default.
- MetaMask / 1inch / Rabby route through their own or aggregator contracts that refund internally, so the problem does not arise there (different architecture).

**Why ours, and why it is safer**
- Unconditional refund removes the *prediction problem*: liquidity can change between quote and execution, so any conditional gate can be wrong at the moment it matters.
- Cost is fixed and tiny (~1.2% of swap gas when unused); the downside of the alternative is unbounded fund loss.
- Safety invariants: the refund is always the **last** multicall entry (earlier placement could drain ETH needed by later calls); ERC-20 swaps are unchanged (the router pulls only what it spends, nothing can strand).
- Audit guidance agrees the router must never hold balances between transactions ([CVE-2022-48216 writeup](https://dedaub.com/blog/uniswap-reentrancy/)).

## 2. Insufficient-liquidity detection with "max you can swap"

Issue: #2446 

**The issue**
- Above ~0.23 ETH the pool runs dry. Uniswap V3 does **not** revert — it silently fills partially.
- The wallet showed the capped quote as a normal price: 1 ETH and 10 ETH produced the same "You receive", with no warning. Combined with issue 1, the extra ETH was lost.

**What we did**
- The wallet already calls QuoterV2 but used to throw away most of its answer. We now decode the full result: when the post-swap price equals the protocol boundary constant (`MIN/MAX_SQRT_RATIO ± 1`), the pool ran dry — an exact signal with no false positives ([QuoterV2.sol](https://github.com/Uniswap/v3-periphery/blob/main/contracts/lens/QuoterV2.sol)).
- On saturation, one extra counter-quote (`quoteExactOutputSingle` for the capped output, with a 0.01% haircut against rounding) computes the **maximum swappable amount**.
- The UI shows *"Insufficient pool liquidity. You can swap at most 0.2361 ETH right now"* and disables Send, through the existing estimation-error pipeline.
- A quoter revert (swap starts in a zero-liquidity region, or extreme tick walks hitting RPC gas caps) maps to the same error instead of "network error".

**How others do it**
- **Nobody detects this client-side.** Uniswap's router, SDK and app all treat a saturated quote as valid; their "Insufficient liquidity" text only appears when no route exists at all. Their V3 SDK silently caps the output ([v3swap.ts](https://github.com/Uniswap/sdks/blob/main/sdks/v3-sdk/src/utils/v3swap.ts)).
- Uniswap fixed this at the protocol level only in **V4**: the new quoter reverts with `NotEnoughLiquidity` ([BaseV4Quoter.sol](https://github.com/Uniswap/v4-periphery/blob/main/src/base/BaseV4Quoter.sol#L53)) — the same check we implement for V3.
- MetaMask/1inch/Rabby hide behind server-side aggregators that drop thin routes; none states a maximum swappable amount.

**Why ours, and why it is safer**
- We quote one known pool directly, so we can be exact where aggregator wallets can only be approximate.
- Discarded alternatives: a second probe quote (extra latency every keystroke, no max amount), local tick-data simulation (bot-grade infrastructure for one pool), binary search over the quoter (no production precedent; ~4.4 s for a sweep on a public RPC).
- Safety: detection uses strict equality with protocol constants — it cannot fire on a merely-expensive swap; the counter-quote failure only degrades the message (never breaks estimation); the displayed max is floor-truncated so the user can never be told more than is actually swappable.

## 3. Price impact ("Price difference") display with warning tiers

Issue: #2447 

**The issue**
- Swapping into a thin pool moves the price against the user: 0.1 ETH already lost ~19.6% versus the marginal rate. The wallet showed nothing — the bad rate just looked like *the* price.

**What we did**
- Compute price impact client-side: pool mid price from one cheap `slot0()` read, compared with the quote's execution price, using the same definition as Uniswap's SDK ([computePriceImpact.ts](https://github.com/Uniswap/sdks/blob/main/sdks/sdk-core/src/utils/computePriceImpact.ts)). Displayed **excluding** the 0.3% pool fee, because the fee already has its own summary row (no double counting).
- Summary always shows a "Price Difference" row for swap routes: warning style at **≥ 5%**, danger style at **≥ 15%**, plus an extra confirmation dialog before sending at the danger tier. Thresholds live in the app config, adjustable without a code change.
- Severity is always color **+ icon + wording** (never color alone — [WCAG 1.4.1](https://www.w3.org/WAI/WCAG22/Understanding/use-of-color.html)); when the reference price cannot be fetched, the row shows "Unknown" at warning style instead of hiding (treating unknown as risky — MetaMask/CoW pattern).

**How others do it**
- Uniswap warns > 5%, escalates ≥ 10%, never hard-blocks, and hides the row below 5% ([constants](https://github.com/Uniswap/interface/blob/main/packages/uniswap/src/constants/transactions.ts)); MetaMask warns at 5%, errors at 25%, always shows the row; 1inch shows a blocking modal at 5%; PancakeSwap still ships typed-"confirm" at 10% and a hard block at 15%; CoW uses relaxed tiers for bridge routes (5/10/15) ([priceImpact.ts](https://github.com/cowprotocol/cowswap/blob/main/apps/cowswap-frontend/src/common/constants/priceImpact.ts)).
- The industry's converged bands: ~5% acknowledge, ~10% escalate, 15-25% block.

**Why ours, and why it is safer**
- Our warn line (5%) matches the industry consensus; the critical tier (15%) is one notch relaxed because this bootstrap pool is deliberately thin and the true disaster case (saturation) is already **blocked** by fix 2 — CoW's bridge tiers are the precedent. Revisit to 10% once the pool deepens.
- Warn-and-confirm instead of hard block follows Uniswap/MetaMask/1inch: a hard block on a thin pool could dead-lock legitimate small swaps; typed-confirm ceremonies are legacy and hostile to accessibility ([NN/g on confirmation dialogs](https://www.nngroup.com/articles/confirmation-dialog/)).
- We compute client-side (fiat-based impact like Rabby/CoW is impossible here — the bridged token has no independent price feed; server-side like Uniswap/MetaMask needs a backend we don't have).
- Known limitation (accepted): `slot0` can be manipulated within a block — fine for a display warning, and every surveyed wallet has the same property; nothing on-chain depends on our number.

## 4. Bridge limit errors shown as validation messages

Issue: #2398 

- **The issue:** the bridge API rejects over-limit amounts with HTTP 400 and an `errorCode` (`REQUEST_LIMIT_EXCEEDED`, `DAILY_LIMIT_EXCEEDED`), but the wallet only ever looked at the error message, so estimation failed silently — no message on the amount field, no explanation.
- **What we did:** `NetworkError` now carries the parsed response body, so the bridge manager can map server error codes to `BridgeEstimationErrorCode` values and return them as estimation state instead of throwing. The amount field shows "Exceeds transfer limit" / "Daily limit exceeded", and any other estimation failure shows "Can't estimate. Try again" rather than nothing.
- Multi-step routes stop at the first failing step (a later step estimated from a failed one would be meaningless), and when several steps report errors the message is chosen by a fixed priority.

## 5. `UNKNOWN_ERROR` popup every 10 s on the swap screen

Issue: #2404 

- **The issue:** for every history transaction the wallet fetched its block just to read the timestamp. On the pruned Erigon node those blocks are gone, ethers surfaces the JSON-RPC error as `UNKNOWN_ERROR`, and the 10 s history poll turned it into a popup on every tick — permanently, for any account with transactions older than the pruning window.
- **What we did:** the Otterscan response already carries `blockTimestamp`, so it is used directly and blocks are requested only for transactions that lack one — the failing call disappears from this path. `BlockService` additionally treats a JSON-RPC error for a single block as "unavailable" and skips it, while genuine connection failures still propagate.
- Background history polling no longer raises error popups at all: a silent background refresh must not interrupt the user.

## 6. Disabled bridge pairs excluded from the UI

Issue: #2456 

- **The issue:** the `enabled` flag of the bridge config had no effect — a pair turned off by the operator was still offered for swapping.
- **What we did:** pair managers expose `isEnabled` (a Uniswap pool has no operator switch, so it is always enabled) and a multi-step route counts as enabled only when every step is. Disabled pairs are filtered out of the selector, and when nothing is left the screen shows a "Bridge Swap Unavailable" dialog instead of an empty selector.
- The pairs-status resolution was fixed along the way, so the "no pairs" state can no longer flash while bridges are still loading.

## 7. Received ETH amount overlapping the chain label in history

Issue: #2443 

- **The issue:** on narrow screens the received ETH amount did not fit its column and was drawn on top of the "ETHEREUM" label, making both unreadable.
- **What we did:** the amount is right-aligned in an over-wide cell, so a long value extends into the free space to its right instead of over the neighbouring text, and the history date drops the time part to leave that space.

## 8. Wrapped XYM renamed to Bridged XYM

Issue: #2421 

- Renamed the token to `Bridged XYM` / `bXYM` across config, known tokens and accounts, token images and the bridge controller, per the naming used in the docs.

## also changed

- **Summary polish:** every row now keeps a value (`-` when it is not known yet) instead of leaving a gap next to the title; the summary is cleared when the token pair changes, so values of the previous pair are never shown against the new one; price impact of a failed estimation step is ignored.
- **Wording:** "Bridge Fee" → "Operation Fee", "Price Difference" → "Price impact", shorter validation messages that fit the field, and a bridge description that warns about price impact on large swaps.
- **Error codes:** dropped `AMOUNT_HIGH` in favour of `INSUFFICIENT_LIQUIDITY` — the ERC-20 revert it was mapped from is a liquidity shortage, not an over-limit amount.
- **Refactor:** bridge pair creation and pairs-status resolution moved out of `useBridge` into `screens/bridge/utils/bridge-pairs.js`, covered by their own unit tests.


wallet/common/core/src/lib/bridge/SwapWorkflowManager.js
```diff
@@ -138,6 +149,10 @@ export class SwapWorkflowManager {
 		for (const manager of this.#pairManagers) {
 			const estimation = await manager.estimateRequest(currentAmount);
 			estimations.push(estimation);
+
+			if (estimation.error)
+				break;
```
Jaguar0625, line 154: when there is error, why do you still want to return earlier estimations instead of just the error?

wallet/common/ethereum/src/utils/uniswap.js
```diff
@@ -0,0 +1,84 @@
...
+
+	const sqrtPrice = BigInt(sqrtPriceX96);
+	const amountInBigInt = BigInt(amountIn);
+
+	const expectedOut = zeroForOne
+		? amountInBigInt * sqrtPrice * sqrtPrice / Q192
+		: amountInBigInt * Q192 / (sqrtPrice * sqrtPrice);
+	const expectedOutAfterFee = expectedOut * (FEE_DENOMINATOR - BigInt(poolFee)) / FEE_DENOMINATOR;
+
+	if (expectedOutAfterFee === 0n)
```
Jaguar0625, line 77: do you need to check negative? can poolFee be > FEE_DENOMINATOR?

wallet/symbol/mobile/__tests__/screens/bridge/utils/price-impact.test.js
```diff
@@ -0,0 +1,178 @@
...
+			{
+				description: 'returns undefined when the swap step failed',
+				config: { estimations: [failedSwapStepEstimation] },
+				expected: { priceImpact: undefined }
+			},
+			{
+				description: 'skips a failed step and returns the impact of a successful one',
+				config: { estimations: [failedSwapStepEstimation, createSwapStepEstimation(0.2)] },
+				expected: { priceImpact: 0.2 }
+			}
```
Jaguar0625, line 177: is this more meaningful than just showing user an error if any step is failing?

wallet/symbol/mobile/src/screens/bridge/components/EstimationSummary.jsx
```diff
@@ -9,9 +12,67 @@ import Animated, { FadeIn, FadeOut } from 'react-native-reanimated';
...
+ * every row keeps a value instead of an empty gap next to its title.
+ * @param {string|null|undefined} value - Row value.
+ * @param {string} [units] - Units shown after the value.
+ * @returns {string} Row value text.
+ */
+const formatRowValue = (value, units) => {
+	if (!value)
+		return MISSING_VALUE_TEXT;
+
+	return [value, units].filter(Boolean).join(' ');
```
Jaguar0625, line 43: why do you need filter(Boolean) here ?

wallet/symbol/mobile/src/screens/bridge/BridgeSwap.jsx
```diff
@@ -98,7 +107,8 @@ export const BridgeSwap = props => {
 		estimate,
 		clearEstimation,
 		isLoading:
-        isEstimationLoading
+        isEstimationLoading,
```
Jaguar0625: looks like this is space indented instead of tab indented

Jaguar0625 (commented):
👍

wallet/common/core/src/lib/bridge/SwapWorkflowManager.js
OlegMakarenko: Thanks for the review!

Even when one step fails, the other step's estimation still holds useful info that is shown in the summary. The error appears as a validation message under the amount field and Send stays disabled.

For example on ETH => XYM: when Uniswap estimation succeeds but the bridge hits the limit, the user sees "Daily limit exceeded" under the amount, and still sees Uniswap's "Price impact: 12.5%  High" in the summary.

wallet/common/ethereum/src/utils/uniswap.js
OlegMakarenko: Negative can't happen. It would need poolFee above 1_000_000 (fee over 100%). Uniswap's factory rejects it: https://github.com/Uniswap/v3-core/blob/main/contracts/UniswapV3Factory.sol#L63 

The value is set in our own config to 3000, which is 0.3%.

wallet/symbol/mobile/__tests__/screens/bridge/utils/price-impact.test.js
OlegMakarenko: Good point to raise. My thinking is that the error and the price impact are inputs to the same decision - which amount to enter. If we show only the error, the user drops the amount below the daily/request limit, resubmits, and discovers price impact is too high. Now they drop it again. This is two rounds for one decision.
Both together - and they can pick an amount that satisfies both constraints in one step.

Open to other approaches if you have one in mind.

Good catch on the test though: we stop at the first failed step, so a failed step is always last, and [failed, successful] can't occur.

wallet/symbol/mobile/src/screens/bridge/components/EstimationSummary.jsx
OlegMakarenko: Units is optional, and join still adds the separator for an undefined entry. So without the filter it will render a trailing space.

Since it's only two values, a ternary might read better. Happy to switch if you prefer that.

wallet/symbol/mobile/src/screens/bridge/BridgeSwap.jsx
OlegMakarenko: Thanks, fixed.

wallet/symbol/mobile/src/screens/bridge/components/EstimationSummary.jsx
Jaguar0625: it's ok. it wasn't clear that units was optional.

Jaguar0625 (approved):
👍

cryptoBeliever (approved):

PR #2457, opened 2026-07-27
[explorer/frontend] fix: header too wide css issue

AnthonyLaw:
Problem: The search modal is too wide because the reusable Card width overrides the header modal width. https://github.com/symbol/product/issues/2407
Solution: Increase the header modal selector specificity so its dimensions are independent of CSS loading order.

### DIFF
--- a/explorer/frontend/styles/components/Header.module.scss
+++ b/explorer/frontend/styles/components/Header.module.scss
@@ -72,7 +72,7 @@
 	user-select: none;
 }
 
-.modal {
+.header .modal {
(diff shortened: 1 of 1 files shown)


OlegMakarenko (approved):
Thanks! This works for now, but let's keep in mind it's a temporary fix rather than a solution for the underlying CSS import order. Other parts of the UI might have similar bugs, but since we've only spotted this one, I'm happy to merge this so we don't block the release.

Let's leave the issue open and revisit it later!

PR #2458, opened 2026-07-27
[bridge/reporting] feat: add header and tabs

AnthonyLaw:
Problem: missing header and tabs
Solution: Set up network status and tabs

image


Wayonb (approved):
👍🏾

PR #2460, opened 2026-07-28
[explorer/puller] feat: add Symbol metadata state synchronization

daoka:
## Summary

Added current-state synchronization for Symbol account, mosaic, and namespace metadata. The puller now collects dirty metadata keys from embedded transactions, resolves address and mosaic aliases, fetches canonical metadata through bounded exact-key node searches, and upserts or deletes rows in the new `symbol_metadata` table.

- Added the `symbol_metadata` table, enum, indexes, and database operations for:
- Added metadata normalization for account, mosaic, and namespace metadata.
- Added dirty-key collection and bounded metadata fetching.
- Extended transaction alias resolution for metadata.
- Extended rollback repair to refresh namespace, mosaic, and metadata state in one database transaction.
- Moved `RollbackRefreshEntries` to `SymbolDatabase.py`, which owns the rollback repair contract.

### DIFF
--- a/explorer/puller/puller/facade/SymbolPuller.py
+++ b/explorer/puller/puller/facade/SymbolPuller.py
@@ -1,3 +1,5 @@
+# pylint: disable=too-many-lines
+
 import asyncio
 import configparser
 import uuid
@@ -13,10 +15,11 @@ from symbollightapi.connector.SymbolConnector import SymbolConnector
 from symbollightapi.model.Exceptions import NodeException
 from zenlog import log
 
-from puller.db.SymbolDatabase import SymbolDatabase
+from puller.db.SymbolDatabase import RollbackRefreshEntries, SymbolDatabase
 from puller.facade.RequestRateLimiter import RequestRateLimiter
 from puller.model.symbol.Account import HARVESTING_ACTIVE_WINDOW_DAYS, create_account_row, create_multisig_row
 from puller.model.symbol.Block import create_block_row
+from puller.model.symbol.Metadata import METADATA_TRANSACTION_TYPE_LABELS, METADATA_TYPE_NUMBERS, create_metadata_row
 from puller.model.symbol.Mosaic import create_mosaic_row
 from puller.model.symbol.Namespace import create_alias_name_rows, create_namespace_row
 from puller.model.symbol.Receipt import (
@@ -37,6 +40,7 @@ MAX_PAGE_SIZE = 100
 ACCOUNT_BATCH_FETCH_SIZE = MAX_PAGE_SIZE
 BLOCK_PAGE_FETCH_CONCURRENCY = 10
 RESOLUTION_FETCH_CONCURRENCY = 10
+METADATA_FETCH_CONCURRENCY = 10
 DEFAULT_MAX_REQUESTS_PER_SECOND = 20
 ACCOUNT_PAGE_SIZE = 100
 
@@ -278,20 +282,22 @@ class SymbolPuller:
 
 	async def _repair_from_height(self, height, sync_state):
 		# Unlike account/multisig rows (deleted by the repair and repopulated by the next dirty-key touch or
-		# refresh snapshot run), namespaces and mosaics have no broad re-dirty signal: only registration,
-		# alias, supply, or expiry events touch their ids, and none may recur after a fork. Re-fetch node state
-		# before the repair write and apply it in the same transaction, deleting an artifact only when the node
-		# confirms it is gone.
+		# refresh snapshot run), namespaces, mosaics, and metadata have no broad re-dirty signal: only
+		# registration, alias, supply, expiry, or metadata events touch their keys, and none may recur after a
+		# fork. Re-fetch node state before the repair write and apply it in the same transaction, deleting an
+		# artifact only when the node confirms it is gone.
 		namespace_ids = self.symbol_db.get_namespace_ids_updated_from_height(height)
 		namespace_entries = await self._fetch_dirty_namespaces(namespace_ids, height - 1)
 		mosaic_ids = self.symbol_db.get_mosaic_ids_updated_from_height(height)
 		mosaic_entries = await self._fetch_dirty_mosaics(mosaic_ids, height - 1)
+		metadata_keys = self.symbol_db.get_metadata_keys_updated_from_height(height)
+		metadata_entries = await self._fetch_dirty_metadata(metadata_keys, height - 1)
 		self.symbol_db.repair_rollback_from_height(height, {
 			**sync_state,
 			'status': 'repairing',
 			'last_synced_height': height - 1,
 			'last_synced_block_hash': self.symbol_db.get_block_hash(height - 1)
-		}, namespace_entries, mosaic_entries)
+		}, RollbackRefreshEntries(namespace_entries, mosaic_entries, metadata_entries))
 		return height
 
 	async def _sync_block_pages(  # pylint: disable=too-many-locals
@@ -363,17 +369,18 @@ class SymbolPuller:
 			dirty_addresses,
 			observed_height,
 			native_mosaic_info)
-		direct_dirty_namespace_ids = self._collect_dirty_namespace_ids_for_batch(
-			transaction_rows_by_height,
-			receipt_rows_by_height)
-		dirty_namespace_ids = self._expand_dirty_namespace_ids(direct_dirty_namespace_ids)
+		dirty_namespace_ids = self._expand_dirty_namespace_ids(
+			self._collect_dirty_namespace_ids_for_batch(transaction_rows_by_height, receipt_rows_by_height))
 		dirty_namespace_entries = await self._fetch_dirty_namespaces(dirty_namespace_ids, observed_height)
 		dirty_mosaic_ids = self._collect_dirty_mosaic_ids_for_batch(transaction_rows_by_height, receipt_rows_by_height)
 		dirty_mosaic_entries = await self._fetch_dirty_mosaics(dirty_mosaic_ids, observed_height)
+		dirty_metadata_keys = self._collect_dirty_metadata_keys_for_batch(transaction_rows_by_height)
+		dirty_metadata_entries = await self._fetch_dirty_metadata(dirty_metadata_keys, observed_height)
 		self._sync_block_batch(batch_rows, transaction_rows_by_height, receipt_rows_by_height)
 		self._write_dirty_accounts_for_batch(dirty_account_rows)
 		self.symbol_db.apply_namespace_entries(dirty_namespace_entries)
 		self._write_dirty_mosaics(dirty_mosaic_entries)
+		self._write_dirty_metadata(dirty_metadata_entries)
 
 	def _sync_block_batch(self, batch_rows, transaction_rows_by_height, receipt_rows_by_height):
 		"""Writes previously-fetched block, transaction, and receipt rows for one batch.
@@ -497,6 +504,9 @@ class SymbolPuller:
 			for mosaic_row in row['mosaic_rows']
 			if is_alias_mosaic_id(mosaic_row['mosaic_id'])
 		}
+		metadata_target_id = row['metadata_target_id']
+		if metadata_target_id is not None and is_alias_mosaic_id(metadata_target_id):
+			alias_mosaic_ids.add(metadata_target_id)
 		return alias_addresses, alias_mosaic_ids
 
 	@classmethod
@@ -547,6 +557,13 @@ class SymbolPuller:
 						source,
 						'mosaic',
 						height)
+			if row['metadata_target_id'] in alias_mosaic_ids:
+				row['metadata_target_id'] = self._resolve_transaction_alias(
+					resolution_statements.mosaic,
+					row['metadata_target_id'].upper(),
+					source,
+					'mosaic',
+					height)
 
 	async def _resolve_transaction_rows_for_batch(self, transaction_rows_by_height):  # pylint: disable=too-many-locals
 		resolution_requests = []
@@ -803,6 +820,88 @@ class SymbolPuller:
 				self.symbol_db.delete_mosaic(entry['mosaic_id'])
 			else:
 				self.symbol_db.upsert_mosaic(entry['row'])
+
+	@staticmethod
+	def _collect_dirty_metadata_keys_for_batch(transaction_rows_by_height):
+		"""Collects unique metadata natural keys touched by metadata transactions."""
+
+		dirty_metadata_keys = {}
+		for transaction_rows in transaction_rows_by_height.values():
+			for transaction_row in transaction_rows:
+				metadata_type = METADATA_TRANSACTION_TYPE_LABELS.get(transaction_row['type'])
+				if metadata_type is None:
+					continue
+
+				body = transaction_row['body']
+				if 'mosaic' == metadata_type:
+					target_id = transaction_row['metadata_target_id']
+				elif 'namespace' == metadata_type:
+					target_id = body['targetNamespaceId']
+				else:
+					target_id = None
+
+				key = {
+					'metadata_type': metadata_type,
+					'source_address': transaction_row['signer_address'],
+					'target_address': transaction_row['target_address'],
+					'scoped_metadata_key': body['scopedMetadataKey'],
+					'target_id': target_id
+				}
+				key_identity = (
+					key['metadata_type'],
+					key['source_address'],
+					key['target_address'],
+					key['scoped_metadata_key'],
+					key['target_id'])
+				dirty_metadata_keys[key_identity] = key
+
+		return list(dirty_metadata_keys.values())
+
+	async def _fetch_dirty_metadata(self, metadata_keys, observed_height):
+		"""Fetches metadata by exact natural key in bounded concurrent batches."""
+
+		if not metadata_keys:
+			return []
+
+		async def fetch_entry(metadata_key):
+			address = Address(metadata_key['source_address'])
+			target_address = Address(metadata_key['target_address'])
+			metadata_number = METADATA_TYPE_NUMBERS[metadata_key['metadata_type']]
+			path = (
+				f'/metadata?sourceAddress={address}&targetAddress={target_address}'
+				f'&scopedMetadataKey={metadata_key["scoped_metadata_key"]}&metadataType={metadata_number}'
+			)
+			if metadata_key['target_id'] is not None:
+				path += f'&targetId={metadata_key["target_id"]}'
+
+			response = await self.get_symbol_node(path)
+			if not isinstance(response, dict) or 'data' not in response:
+				raise ValueError('Malformed Symbol metadata search response')
+			items = response['data']
+			if not isinstance(items, list):
+				raise ValueError('Malformed Symbol metadata search data')
+			if len(items) > 1:
+				raise ValueError('Symbol metadata exact-key search returned multiple entries')
+			if not items:
+				return {'key': metadata_key}
+
+			return {'row': create_metadata_row(items[0], observed_height)}
+
+		entries = []
+		for chunk_start in range(0, len(metadata_keys), METADATA_FETCH_CONCURRENCY):
+			chunk = metadata_keys[chunk_start:chunk_start + METADATA_FETCH_CONCURRENCY]
+			entries.extend(await asyncio.gather(*(fetch_entry(metadata_key) for metadata_key in chunk)))
+
+		return entries
+
+	def _write_dirty_metadata(self, entries):
+		"""Writes fetched metadata current-state changes after all batch fetches complete."""
+
+		for entry in entries:
+			if 'key' in entry:
+				self.symbol_db.delete_metadata_by_key(entry['key'])
+			else:
+				self.symbol_db.upsert_metadata(entry['row'])
 
 	async def _fetch_dirty_accounts_for_batch(  # pylint: disable=too-many-locals
 		self,
--- a/explorer/puller/puller/facade/SymbolPuller.py
+++ b/explorer/puller/puller/facade/SymbolPuller.py
@@ -442,14 +442,14 @@ class SymbolPuller:
 
 		return dict(rows_by_height)
 
-	async def _get_resolution_statements(self, kind, height):
+	async def _get_resolution_statements(self, request):
 		resolution_entries_by_unresolved = {}
 		page_number = 1
 		while True:
 			response = await self.get_symbol_node(
-				f'/statements/resolutions/{kind}?height={height}&pageSize={MAX_PAGE_SIZE}&pageNumber={page_number}'
+				f'/statements/resolutions/{request.kind}?height={request.height}&pageSize={MAX_PAGE_SIZE}&pageNumber={page_number}'
 			)
-			items = self._get_node_page_data(response, f'Malformed Symbol {kind} resolution page response')
+			items = self._get_node_page_data(response, f'Malformed Symbol {request.kind} resolution page response')
 			for item in items:
 				statement = item['statement']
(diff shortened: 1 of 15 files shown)


explorer/puller/tests/test/SymbolMetadataTestUtils.py
```diff
@@ -0,0 +1,109 @@
...
+			'value_hex': row[6],
+			'value_utf8': row[7],
+			'raw_payload': row[8],
+			'updated_at_height': row[9]
+		}
+		for row in cursor.fetchall()
+	]
+
+
+def fetch_metadata_row(database, composite_hash):
```
Jaguar0625: fetch => find

explorer/puller/puller/model/symbol/Transaction.py
```diff
@@ -217,6 +217,11 @@ def create_transaction_row(item, network, epoch_adjustment_seconds):
 	transaction_type = int(transaction['type'])
 	signer_public_key = bytes.fromhex(transaction['signerPublicKey'])
 	signer_address = address_from_public_key(signer_public_key, network)
+	metadata_target_id = (
```
Jaguar0625: what is purpose of this column? why aren't you also exposing target id for namespace metadata?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1182,7 +1463,7 @@ def test_repair_rollback_from_height_leaves_namespace_current_state_rows_in_plac
 		database.repair_rollback_from_height(2, _create_sync_state(
 			status='repairing',
 			last_synced_height=1,
-			last_synced_block_hash=b'hash 1'), [], [])
+			last_synced_block_hash=b'hash 1'), RollbackRefreshEntries([], [], []))
```
Jaguar0625: can't you use defaults when defining, so you can do `RollbackRefreshEntries()` here

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1226,50 +1507,153 @@ def test_repair_rollback_from_height_applies_namespace_and_mosaic_changes_in_sam
...
-			last_synced_block_hash=b'hash 1'), [
-			{'row': refreshed_row, 'alias_rows': _create_alias_name_rows(refreshed_row)},
-			{'namespace_id': deleted_row['namespace_id']}
-		], [
-			{'row': create_expected_mosaic_row(create_mosaic_item(mosaic_id=mosaic_id, supply='99'), 1)},
-			{'mosaic_id': deleted_mosaic_id}
-		])
+			last_synced_block_hash=b'hash 1')
+		expected_mosaic_row = create_expected_mosaic_row(create_mosaic_item(mosaic_id=mosaic_id, supply='99'), 1)
+		expected_mosaic_state = [create_persisted_mosaic_state(expected_mosaic_row, [])]
```
Jaguar0625, line 1562: refresh_entries.mosaic_entries[0]['row']

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1226,50 +1507,153 @@ def test_repair_rollback_from_height_applies_namespace_and_mosaic_changes_in_sam
...
-			('namespace', NAMESPACE_ROOT_ID, 'root', 1)
-		], cursor.fetchall())
-		cursor.execute('SELECT mosaic_id, supply, updated_at_height FROM symbol_mosaics ORDER BY mosaic_id')
-		self.assertEqual([(mosaic_id, 99, 1)], cursor.fetchall())
 		cursor.execute('SELECT height FROM symbol_blocks ORDER BY height')
 		self.assertEqual([(1,)], cursor.fetchall())
-		self.assertEqual('repairing', database.get_sync_state()['status'])
-		self.assertEqual(1, database.get_sync_state()['last_synced_height'])
+
+	# pylint: disable=too-many-locals
```
Jaguar0625: move inside function scope

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1226,50 +1507,153 @@ def test_repair_rollback_from_height_applies_namespace_and_mosaic_changes_in_sam
...
+		with self.assertRaises(PsycopgError):
+			database.repair_rollback_from_height(2, _create_sync_state(
+				status='repairing', last_synced_height=1, last_synced_block_hash=b'hash 1'), refresh_entries)
+
+		# Assert:
+		self.assertEqual(original_state['blocks'], fetch_full_block_state(database))
+		self.assertEqual(original_state['sync_state'], fetch_normalized_sync_state(database))
+		self.assertEqual(original_state['namespace'], fetch_namespace_state(database.connection))
+		self.assertEqual(original_state['mosaic'], fetch_mosaic_state(database))
+		self.assertEqual(original_state['metadata'], fetch_metadata_rows(database))
```
Jaguar0625, line 1660: not required, but i'd suggest checking that the length of all these are 2 (original + deleted)

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -547,6 +557,13 @@ def _resolve_transaction_row(self, row, top_level_rows_by_hash, resolution_state
 						source,
 						'mosaic',
 						height)
+			if row['metadata_target_id'] in alias_mosaic_ids:
+				row['metadata_target_id'] = self._resolve_transaction_alias(
+					resolution_statements.mosaic,
+					row['metadata_target_id'].upper(),
+					source,
+					'mosaic',
+					height)
```
Jaguar0625, line 556: consider packing these in a named tuple since only difference from previous call is second parameter

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -804,6 +821,88 @@ def _write_dirty_mosaics(self, entries):
...
+				else:
+					target_id = None
+
+				key = {
+					'metadata_type': metadata_type,
+					'source_address': transaction_row['signer_address'],
+					'target_address': transaction_row['target_address'],
+					'scoped_metadata_key': body['scopedMetadataKey'],
+					'target_id': target_id
+				}
```
Jaguar0625, line 848: is there reason you don't want to use composite hash as key ?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -804,6 +821,88 @@ def _write_dirty_mosaics(self, entries):
...
+			metadata_number = METADATA_TYPE_NUMBERS[metadata_key['metadata_type']]
+			path = (
+				f'/metadata?sourceAddress={address}&targetAddress={target_address}'
+				f'&scopedMetadataKey={metadata_key["scoped_metadata_key"]}&metadataType={metadata_number}'
+			)
+			if metadata_key['target_id'] is not None:
+				path += f'&targetId={metadata_key["target_id"]}'
+
+			response = await self.get_symbol_node(path)
+			if not isinstance(response, dict) or 'data' not in response:
```
Jaguar0625: use _get_node_page_data ?

explorer/puller/tests/facade/symbol/puller_test_utils.py
```diff
@@ -497,6 +504,54 @@ async def post(self, url_path, request_payload, *_):
...
+
+class BoundedMetadataConnector(BoundedDetailConnector):
+	DETAIL_PATH_PREFIX = 'metadata?'
+
+	def _get_detail_response(self, url_path):
+		response = self.metadata_by_query.get(url_path, {'data': []})
+		if isinstance(response, Exception):
+			raise response
+
+		return response
```
Jaguar0625: can move exception handling into base too and just return response here in subclasses

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -222,6 +224,151 @@ def test_sync_block_headers_leaves_rollback_state_unchanged_when_mosaic_fetch_fa
...
+		# Act / Assert:
+		with self.assertRaisesRegex(RuntimeError, 'metadata fetch failed'):
+			asyncio.run(self.puller.sync_block_headers())
+
+		# Assert:
+		self.assertEqual(original_block_state, fetch_full_block_state(self.puller.symbol_db))
+		self.assertEqual(original_sync_state, fetch_normalized_sync_state(self.puller.symbol_db))
+		self.assertEqual(original_namespace_state, fetch_namespace_state(self.puller.symbol_db.connection))
+		self.assertEqual(original_mosaic_state, fetch_mosaic_state(self.puller.symbol_db))
+		self.assertEqual(original_metadata_rows, fetch_metadata_rows(self.puller.symbol_db))
```
Jaguar0625, line 357: would also check lengths here

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -222,6 +224,151 @@ def test_sync_block_headers_leaves_rollback_state_unchanged_when_mosaic_fetch_fa
...
+		self.assertEqual([metadata_path(1, mosaic_id)], [
+			path for path in connector.paths if path.startswith('metadata?')])
+		self.assertEqual([{'namespaceIds': [NAMESPACE_ROOT_ID]}], [
+			payload for path, payload in connector.post_requests if 'namespaces/names' == path])
+		self.assertEqual([{'mosaicIds': [mosaic_id]}], [
+			payload for path, payload in connector.post_requests if 'mosaics' == path])
+		self.assertLess(
+			connector.paths.index(f'namespaces/{NAMESPACE_ROOT_ID}'),
+			connector.paths.index(metadata_path(1, mosaic_id)))
+		self.assertLess(connector.paths.index('mosaics'), connector.paths.index(metadata_path(1, mosaic_id)))
```
Jaguar0625: why are you using assertLess here?

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
```diff
@@ -0,0 +1,1150 @@
...
+		# Arrange:
+		rows = {}
+
+		# Act:
+		keys = self.puller._collect_dirty_metadata_keys_for_batch(rows)  # pylint: disable=protected-access
+
+		# Assert:
+		self.assertEqual([], keys)
+
+	def test_collect_dirty_metadata_keys_uses_resolved_target_address_for_account_metadata(self):
```
Jaguar0625: i don't understand what these tests are testing? based on name, it implies the address should be resolved, but _collect_dirty_metadata_keys_for_batch doesn't appear to do any resolution?

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
```diff
@@ -0,0 +1,1150 @@
...
+			target_id=MOSAIC_ID, value_utf8='hello'))
+		self.puller.symbol_db.upsert_metadata(sibling_row)
+
+		# Act:
+		self.puller._write_dirty_metadata([{'key': key}])  # pylint: disable=protected-access
+
+		# Assert:
+		self.assertEqual([sibling_row], fetch_metadata_rows(self.puller.symbol_db))
+
+	def test_fetch_dirty_metadata_rejects_malformed_response(self):
```
Jaguar0625: refactor this group of tests where only difference is metadata_by_query and regex value

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
```diff
@@ -0,0 +1,1150 @@
...
+				'version': 1,
+				'network': 152,
+				'targetAddress': ALIAS_ADDRESS,
+				'targetMosaicId': ALIAS_MOSAIC_ID,
+				'scopedMetadataKey': SCOPED_METADATA_KEY,
+				'valueSizeDelta': 5,
+				'value': '68656C6C6F'
+			})
+
+	def test_sync_block_headers_resolves_address_alias_and_persists_account_metadata(self):
```
Jaguar0625: can you explain what you want this test to verify on top of test_sync_block_headers_resolves_address_alias_for_embedded_metadata ?

Jaguar0625: is it just that it is account vs mosaic ? if so consider renaming previous one

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
```diff
@@ -0,0 +1,1150 @@
...
+		self.assertEqual([metadata_failure_path], [path for path in connector.paths if path.startswith('metadata?')])
+		self.assertEqual([{'addresses': account_addresses}], [
+			payload for path, payload in connector.post_requests if 'accounts' == path])
+		self.assertEqual([{'namespaceIds': [NAMESPACE_ID]}], [
+			payload for path, payload in connector.post_requests if 'namespaces/names' == path])
+		self.assertEqual([{'mosaicIds': [dirty_mosaic_id]}], [
+			payload for path, payload in connector.post_requests if 'mosaics' == path])
+		metadata_request_index = connector.paths.index(metadata_failure_path)
+		for request_path in ('accounts', f'namespaces/{NAMESPACE_ID}', 'mosaics'):
+			self.assertLess(connector.paths.index(request_path), metadata_request_index)
```
Jaguar0625: why using assertLess here?

Jaguar0625 (commented):
👍

explorer/puller/tests/test/SymbolMetadataTestUtils.py
daoka: Renamed `fetch_metadata_row` to `find_metadata_row` and updated all call sites.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Added tuple defaults for all three `RollbackRefreshEntries` fields. Empty refresh cases now use `RollbackRefreshEntries()`, while production callers that have refresh data continue to pass all three fields explicitly.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Updated `expected_mosaic_row` to use `refresh_entries.mosaic_entries[0]['row']`, avoiding a duplicate reconstruction of the same expected row.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Moved the `too-many-locals` pylint disable into the beginning of the affected test function body.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Added explicit assertions that the original namespace, mosaic, and metadata states each contain two entries before the rollback failure is triggered.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Added `ResolutionRequest(height, kind)` and now use it consistently when collecting resolution requests, issuing the concurrent fetches, and associating responses with their height and kind.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Reused `_get_node_page_data` for the common response-object and `data` extraction. The metadata-specific list validation and exact-key uniqueness check remain in `_fetch_dirty_metadata`.

explorer/puller/tests/facade/symbol/puller_test_utils.py
daoka: Moved configured-exception handling into `BoundedDetailConnector.get`. The metadata and namespace subclasses now only return their configured response from `_get_detail_response`.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Added explicit assertions that the original namespace, mosaic, and metadata states each contain one entry before triggering the metadata fetch failure.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Replaced the partial `assertLess` checks with one exact filtered request-sequence assertion: namespace detail, `namespaces/names`, `mosaics`, then metadata. This verifies both order and the absence of duplicate or unexpected requests.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: The resolution happens before the collector runs. These tests verify that the collector prefers the resolved `transaction_row['target_address']` over the unresolved `body['targetAddress']`; they do not claim that the collector performs resolution. I renamed the tests accordingly, made the alias and resolved values distinct, and refactored them through a branch-free helper that receives an explicit row and expected key.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: Extracted `_assert_fetch_dirty_metadata_rejects_response(response, expected_message)` and converted each case to pass only its response and expected error message. The multiple-entry response is created inline so no mutable response object is shared between tests.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: > can you explain what you want this test to verify on top of test_sync_block_headers_resolves_address_alias_for_embedded_metadata ?

This test covers the account-metadata variant of the same address-alias flow. Unlike mosaic metadata, account metadata sends no artifact target id, so the exact-key request omits `targetId` and the persisted `target_id` is `None`. I renamed the test to `test_sync_block_headers_resolves_address_alias_for_embedded_account_metadata` so the transaction type being covered is explicit.

> is it just that it is account vs mosaic ? if so consider renaming previous one

Yes, the metadata type is the main distinction. I renamed the related tests symmetrically to identify embedded account, mosaic, and namespace metadata explicitly. The mosaic-specific names also state whether the address alias, `targetMosaicId` alias, or both are resolved.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: Replaced the `assertLess` checks with one exact filtered sequence assertion covering `accounts`, namespace detail, `namespaces/names`, `mosaics`, and metadata. The filter uses exact matching for namespace and mosaic paths so startup requests such as `mosaics/{nativeMosaicId}` are not included accidentally.

explorer/puller/puller/model/symbol/Transaction.py
daoka: metadata_target_id is not a persisted column; it is an auxiliary normalized field used only for mosaic metadata. Symbol defines targetMosaicId as UnresolvedMosaicId, so it can contain a namespace alias and must pass through the existing mosaic resolver before the exact-key metadata lookup. The original value remains unchanged in the transaction body, while this field carries the resolved mosaic id.
Namespace metadata does have an artifact id, targetNamespaceId, but Symbol defines it as a concrete NamespaceId. Core stores it directly in the metadata key without passing it through the mosaic resolver. The collector therefore reads targetNamespaceId directly from the transaction body and includes it as the natural key's target_id; it is not omitted, only not duplicated into this auxiliary field. I kept the field and added a self-contained code comment explaining this distinction.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Persisted rows do use composite_hash as their primary key. The natural key is used only for dirty-key collection, exact-key lookup, and deletion when the lookup returns no entry. A metadata transaction does not contain the composite hash, and an empty node result cannot provide one. Computing it locally would duplicate core’s consensus-critical key-generation logic and introduce drift risk.

explorer/puller/puller/model/symbol/Transaction.py
Jaguar0625: i would name it something more specific then, like mosaic_target_id or mosaic_alias_id to avoid confusion. since you only seem to use this in puller if it is an alias, you can also move alias check into Transaction.py.

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: i think you can pass request directly to  _get_resolution_statements now?

Jaguar0625: i was also thinking about adding local function and refactoring like this (can do similar with account):

```py
		if alias_mosaic_ids:
			def _resolve_transaction_alias_mosaic(mosaic_id):
				return self._resolve_transaction_alias(
					resolution_statements.mosaic,
					mosaic_id.upper(),
					source,
					'mosaic',
					height)

			for mosaic_row in row['mosaic_rows']:
				if mosaic_row['mosaic_id'] in alias_mosaic_ids:
					mosaic_row['mosaic_id'] = _resolve_transaction_alias_mosaic(mosaic_row['mosaic_id'])

			if row['metadata_target_id'] in alias_mosaic_ids:
				row['metadata_target_id'] = _resolve_transaction_alias_mosaic(row['metadata_target_id'])
```

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
Jaguar0625: consider renaming to something like selects/extracts resolved target id

since verb "revolves" implies there is some resolution going on in the test (but there isn't)

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: Updated _get_resolution_statements to accept a ResolutionRequest directly and read request.kind and request.height. The bounded asyncio.gather now passes each request unchanged, and no old two-argument call sites remain.

daoka: Refactored _resolve_transaction_row with local address and mosaic resolver functions. Both mosaic_rows and mosaic_metadata_target_id use the mosaic helper, while address rows and normalized recipient/target address fields use the address helper. The existing source and entry-selection rules, resolved-hex-to-bytes conversion, address deduplication, and error behavior are unchanged.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: Renamed the helper and the three tests to use selects_row_target_address. The collector does not perform resolution; it selects the already-normalized transaction_row['target_address'] instead of raw body['targetAddress']. The raw alias and normalized row value remain distinct in each test, and the helper remains branch-free with only (transaction_row, expected_key).

explorer/puller/puller/model/symbol/Transaction.py
daoka: The alias check could technically be moved, but mosaic_metadata_target_id is not alias-only: it carries both direct mosaic IDs and unresolved aliases. Direct IDs remain unchanged, while aliases are replaced during the existing batch resolution phase.
Transaction.py only converts a transaction DTO into a row and does not have the resolution statements or transaction-source context needed to apply a resolution entry. SymbolPuller already aggregates aliases by height, decides which resolution statements to fetch, and applies them using the transaction source.
Storing only aliases in this field would require the collector to fall back to the raw body for direct IDs, giving it two input sources. I therefore kept one normalized field for both cases and left alias detection and resolution in SymbolPuller, alongside the existing address and mosaic-row resolution paths.

Jaguar0625: i missed this usage before  👍 
```py
target_id = transaction_row['mosaic_metadata_target_id']
```

Jaguar0625 (approved):
👍

PR #2461, opened 2026-07-28
[explorer/puller] fix: flush harvested fees for the trailing block batch

cryptoBeliever:
## Problem

`accounts.harvested_fees` silently under-counts. On a local testnet database the deficit was **-150,000** against the sum of `blocks.total_fee`, and `last_harvested_height` value.

## Root cause

`_db_writer` keeps two accumulators, `pending_addresses` and `accounts_harvested_fee`, and flushes both every `batch_size` blocks:

```python
if processed % batch_size == 0:
    asyncio.run(self._process_account_batch(cursor, pending_addresses))
    pending_addresses.clear()
    self._process_harvested_fees(cursor, accounts_harvested_fee)   # flushed
    accounts_harvested_fee.clear()
```

The shutdown path, entered after the stop signal, flushed only the first one:
```python
  if block is None:
      if pending_addresses:
          asyncio.run(self._process_account_batch(cursor, pending_addresses))
      block_queue.task_done()
      break
  ```     
      
`accounts_harvested_fee` was dropped, but `if processed % batch_size != 0: self._commit_blocks()` committed the blocks regardless. Since `update_account_harvested_fees` accumulates (`harvested_fees = harvested_fees + %s`) and the next run resumes from `SELECT MAX(height) FROM blocks`, the skipped fees were never recovered.

 ## Fix

Extract `_flush_pending_account_state`, which flushes and clears both accumulators, and call it from both the batch path and the shutdown path. The duplication that allowed the two paths to diverge is gone.

As a side effect `_process_account_batch` is no longer invoked with an empty dict, removing misleading `Processing batch of 0 addresses` log lines.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -334,6 +334,17 @@ class NemPuller:
 				last_height
 			)
 
+	def _flush_pending_account_state(self, cursor, pending_addresses, accounts_harvested_fee):
+		"""Flushes accumulated account and harvested fee state, clearing both accumulators."""
+
+		if pending_addresses:
+			asyncio.run(self._process_account_batch(cursor, pending_addresses))
+			pending_addresses.clear()
+
+		if accounts_harvested_fee:
+			self._process_harvested_fees(cursor, accounts_harvested_fee)
+			accounts_harvested_fee.clear()
+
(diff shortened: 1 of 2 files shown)


Wayonb (approved):
👍🏾

AnthonyLaw (approved):
Good catch 👍🏼

PR #2462, opened 2026-07-29
[bridge/reporting] feat: Added report panel in home page

AnthonyLaw:
Problem: missing search and payout filter.
Solution: Added a search input box and payout filter.

image


Wayonb (approved):
👍🏾

PR #2463, opened 2026-07-29
[explorer/puller] feat: add Symbol lock state synchronization

daoka:
This PR adds puller-only Symbol Hash Lock and Secret Lock synchronization.

- Added exact Hash Lock and Secret Lock current-state schemas, enums, and indexes.
- Collects affected Lock keys from transactions, expiry ranges, and rollback state.
- Fetches canonical Lock state from the node before writing, with bounded concurrency.
- Uses the node-provided Secret Lock `compositeHash` without calculating it locally.
- Validates and locally filters paginated Secret Lock responses, rejecting malformed or duplicate data.
- Applies Hash/Secret Lock updates after Metadata and restores affected state atomically during rollback.

### DIFF
--- a/explorer/puller/puller/db/SymbolDatabase.py
+++ b/explorer/puller/puller/db/SymbolDatabase.py
@@ -11,7 +11,7 @@ from puller.model.symbol.Lock import (
 	LOCK_HASH_ALGORITHM_LABELS,
 	LOCK_STATUS_LABELS,
 	RollbackLockKeys,
-	create_hash_lock_key_from_hex,
+	create_hash_lock_key,
 	create_secret_lock_search_key,
 	create_secret_lock_search_key_from_hex_secret,
 	lock_hash_algorithm_label
@@ -1005,19 +1005,20 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 
 		cursor = self.connection.cursor()
 		cursor.execute(
-			'SELECT hash FROM symbol_hash_locks WHERE updated_at_height >= %s ORDER BY hash',
+			'SELECT hash FROM symbol_hash_locks WHERE updated_at_height >= %s ORDER BY updated_at_height, hash',
 			(height,))
 
-		return [create_hash_lock_key_from_hex(bytes(row[0]).hex()) for row in cursor.fetchall()]
+		return [create_hash_lock_key(bytes(row[0])) for row in cursor.fetchall()]
 
 	def get_secret_lock_search_keys_updated_from_height(self, height):  # pylint: disable=invalid-name
 		"""Gets Secret Lock logical keys whose current state was observed at or after a height."""
 
 		return self._get_secret_lock_search_keys(
 			'updated_at_height >= %s',
-			(height,))
+			(height,),
+			'updated_at_height')
 
-	def get_hash_lock_hashes_expiring_between(self, start_height, end_height):  # pylint: disable=invalid-name
+	def get_hash_lock_hashes_reaching_end_height_between(self, start_height, end_height):  # pylint: disable=invalid-name
 		"""Gets Hash Lock keys whose end height is inside an inclusive range."""
 
 		cursor = self.connection.cursor()
@@ -1026,27 +1027,33 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 			SELECT hash
 			FROM symbol_hash_locks
 			WHERE end_height BETWEEN %s AND %s
-			ORDER BY hash
+			ORDER BY end_height, hash
 			''',
 			(start_height, end_height))
 
-		return [create_hash_lock_key_from_hex(bytes(row[0]).hex()) for row in cursor.fetchall()]
+		return [create_hash_lock_key(bytes(row[0])) for row in cursor.fetchall()]
 
-	def get_secret_lock_search_keys_expiring_between(self, start_height, end_height):  # pylint: disable=invalid-name
+	def get_secret_lock_search_keys_reaching_end_height_between(self, start_height, end_height):  # pylint: disable=invalid-name
 		"""Gets Secret Lock logical keys whose end height is inside an inclusive range."""
 
 		return self._get_secret_lock_search_keys(
 			'end_height BETWEEN %s AND %s',
-			(start_height, end_height))
+			(start_height, end_height),
+			'end_height')
 
-	def _get_secret_lock_search_keys(self, predicate, parameters):
+	def _get_secret_lock_search_keys(self, predicate, parameters, sort_column):
+		# Keep interpolated SQL identifiers restricted to the two fixed query contracts.
+		sort_column = {
+			'updated_at_height': 'updated_at_height',
+			'end_height': 'end_height'
+		}[sort_column]
 		cursor = self.connection.cursor()
 		cursor.execute(
 			f'''
 			SELECT owner_address, recipient_address, secret, hash_algorithm
 			FROM symbol_secret_locks
 			WHERE {predicate}
-			ORDER BY composite_hash
+			ORDER BY {sort_column}, composite_hash
 			''',
 			parameters)
 
@@ -1062,7 +1069,7 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 		cursor = self.connection.cursor()
 		cursor.execute(
 			'''
-			SELECT type, is_embedded, hash, signer_address, recipient_address, body
+			SELECT type, hash, signer_address, recipient_address, body
 			FROM symbol_transactions
 			WHERE height >= %s
 				AND type IN (%s, %s, %s, %s)
@@ -1076,27 +1083,25 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 				TransactionType.AGGREGATE_BONDED.value
 			))
 
-		hash_keys = {}
-		secret_keys = {}
-		for transaction_type, is_embedded, transaction_hash, signer_address, recipient_address, body in cursor.fetchall():
+		hash_keys = []
+		secret_keys = []
+		for transaction_type, transaction_hash, signer_address, recipient_address, body in cursor.fetchall():
 			if TransactionType.HASH_LOCK.value == transaction_type:
-				hash_key = create_hash_lock_key_from_hex(body['hash'])
-				hash_keys[hash_key] = None
-			elif TransactionType.AGGREGATE_BONDED.value == transaction_type and not is_embedded:
-				hash_key = create_hash_lock_key_from_hex(bytes(transaction_hash).hex())
-				hash_keys[hash_key] = None
+				hash_keys.append(create_hash_lock_key(body['hash']))
+			elif TransactionType.AGGREGATE_BONDED.value == transaction_type:
+				hash_keys.append(create_hash_lock_key(bytes(transaction_hash)))
 			elif TransactionType.SECRET_LOCK.value == transaction_type:
-				secret_key = create_secret_lock_search_key_from_hex_secret(
+				secret_keys.append(create_secret_lock_search_key_from_hex_secret(
 					bytes(signer_address), bytes(recipient_address), body['secret'],
-					lock_hash_algorithm_label(body['hashAlgorithm']))
-				secret_keys[secret_key] = None
+					lock_hash_algorithm_label(body['hashAlgorithm'])))
 			elif TransactionType.SECRET_PROOF.value == transaction_type:
-				secret_key = create_secret_lock_search_key_from_hex_secret(
+				secret_keys.append(create_secret_lock_search_key_from_hex_secret(
 					None, bytes(recipient_address), body['secret'],
-					lock_hash_algorithm_label(body['hashAlgorithm']))
-				secret_keys[secret_key] = None
+					lock_hash_algorithm_label(body['hashAlgorithm'])))
 
-		return RollbackLockKeys(list(hash_keys), list(secret_keys))
+		return RollbackLockKeys(
+			list(dict.fromkeys(hash_keys)),
+			list(dict.fromkeys(secret_keys)))
 
 	def delete_namespace(self, namespace_id):
 		"""Deletes one namespace and its derived alias-name rows when it exists."""
--- a/explorer/puller/puller/db/SymbolDatabase.py
+++ b/explorer/puller/puller/db/SymbolDatabase.py
@@ -3,9 +3,19 @@ from collections import namedtuple
 from contextlib import contextmanager
 
 from psycopg2.extras import Json
+from symbolchain.sc import TransactionType
 
 from puller.model.symbol.Account import ACCOUNT_TYPE_VALUES
 from puller.model.symbol.Block import BLOCK_TYPE_VALUES
+from puller.model.symbol.Lock import (
+	LOCK_HASH_ALGORITHM_LABELS,
+	LOCK_STATUS_LABELS,
+	RollbackLockKeys,
+	create_hash_lock_key_from_hex,
+	create_secret_lock_search_key,
+	create_secret_lock_search_key_from_hex_secret,
+	lock_hash_algorithm_label
+)
 from puller.model.symbol.Metadata import METADATA_TYPE_LABELS
 from puller.model.symbol.Namespace import NAMESPACE_ALIAS_TYPE_LABELS, NAMESPACE_REGISTRATION_TYPE_LABELS
 from puller.model.symbol.Receipt import RECEIPT_TYPE_LABELS
@@ -14,7 +24,9 @@ from puller.model.symbol.Transaction import MESSAGE_TYPE_LABELS, TRANSACTION_TYP
 from .DatabaseConnection import DatabaseConnection
 
 RollbackRefreshEntries = namedtuple(
-	'RollbackRefreshEntries', ['namespace_entries', 'mosaic_entries', 'metadata_entries'], defaults=((), (), ()))
+	'RollbackRefreshEntries',
+	['namespace_entries', 'mosaic_entries', 'metadata_entries', 'hash_lock_entries', 'secret_lock_entries'],
+	defaults=((), (), (), (), ()))
 
 SYNC_STATE_COLUMNS = [
 	'status',
@@ -38,6 +50,8 @@ SYMBOL_NAMESPACE_REGISTRATION_TYPE_VALUES = tuple(NAMESPACE_REGISTRATION_TYPE_LA
 SYMBOL_NAMESPACE_ALIAS_TYPE_VALUES = tuple(NAMESPACE_ALIAS_TYPE_LABELS.values())
 SYMBOL_ALIAS_ARTIFACT_TYPE_VALUES = ('mosaic', 'namespace', 'account')
 SYMBOL_METADATA_TYPE_VALUES = tuple(METADATA_TYPE_LABELS.values())
+SYMBOL_LOCK_STATUS_VALUES = tuple(LOCK_STATUS_LABELS.values())
+SYMBOL_LOCK_HASH_ALGORITHM_VALUES = tuple(LOCK_HASH_ALGORITHM_LABELS.values())
 ACCOUNT_REFRESH_STATE_STATUS_VALUES = ('healthy', 'refreshing', 'stale', 'unhealthy')
 ACCOUNT_REFRESH_STATE_COLUMNS = [
 	'last_successful_run_id',
@@ -302,6 +316,57 @@ SYMBOL_METADATA_INDEXES = [
 	'CREATE INDEX IF NOT EXISTS idx_symbol_metadata_target_id ON symbol_metadata(target_id)',
 	'CREATE INDEX IF NOT EXISTS idx_symbol_metadata_updated_height ON symbol_metadata(updated_at_height)'
 ]
+SYMBOL_HASH_LOCK_DEFINITIONS = [
+	'hash bytea PRIMARY KEY',
+	'owner_address bytea NOT NULL',
+	'mosaic_id varchar(16) NOT NULL',
+	'amount bigint NOT NULL',
+	'end_height bigint NOT NULL',
+	'status symbol_lock_status NOT NULL',
+	'raw_payload jsonb NOT NULL',
+	'updated_at_height bigint NOT NULL'
+]
+SYMBOL_SECRET_LOCK_DEFINITIONS = [
+	'composite_hash bytea PRIMARY KEY',
+	'owner_address bytea NOT NULL',
+	'recipient_address bytea NOT NULL',
+	'secret bytea NOT NULL',
+	'hash_algorithm symbol_lock_hash_algorithm NOT NULL',
+	'mosaic_id varchar(16) NOT NULL',
+	'amount bigint NOT NULL',
+	'end_height bigint NOT NULL',
+	'status symbol_lock_status NOT NULL',
+	'raw_payload jsonb NOT NULL',
+	'updated_at_height bigint NOT NULL'
+]
+SYMBOL_LOCK_INDEXES = [
+	'CREATE INDEX IF NOT EXISTS idx_symbol_hash_locks_owner_end_height '
+	'ON symbol_hash_locks(owner_address, end_height DESC, hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_hash_locks_mosaic_end_height '
+	'ON symbol_hash_locks(mosaic_id, end_height DESC, hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_hash_locks_status_end_height '
+	'ON symbol_hash_locks(status, end_height DESC, hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_hash_locks_end_height '
+	'ON symbol_hash_locks(end_height DESC, hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_hash_locks_updated_at_height '
+	'ON symbol_hash_locks(updated_at_height)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_owner_end_height '
+	'ON symbol_secret_locks(owner_address, end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_recipient_end_height '
+	'ON symbol_secret_locks(recipient_address, end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_mosaic_end_height '
+	'ON symbol_secret_locks(mosaic_id, end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_status_end_height '
+	'ON symbol_secret_locks(status, end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_hash_algorithm_end_height '
+	'ON symbol_secret_locks(hash_algorithm, end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_end_height '
+	'ON symbol_secret_locks(end_height DESC, composite_hash DESC)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_updated_at_height '
+	'ON symbol_secret_locks(updated_at_height)',
+	'CREATE INDEX IF NOT EXISTS idx_symbol_secret_locks_search '
+	'ON symbol_secret_locks(secret, recipient_address, hash_algorithm, owner_address)'
+]
 SYMBOL_TRANSACTION_DEFINITIONS = [
 	'id bigserial PRIMARY KEY',
 	'hash bytea UNIQUE',
@@ -414,6 +479,8 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 		_create_enum_type(cursor, 'symbol_namespace_alias_type', SYMBOL_NAMESPACE_ALIAS_TYPE_VALUES)
 		_create_enum_type(cursor, 'symbol_alias_artifact_type', SYMBOL_ALIAS_ARTIFACT_TYPE_VALUES)
 		_create_enum_type(cursor, 'symbol_metadata_type', SYMBOL_METADATA_TYPE_VALUES)
+		_create_enum_type(cursor, 'symbol_lock_status', SYMBOL_LOCK_STATUS_VALUES)
+		_create_enum_type(cursor, 'symbol_lock_hash_algorithm', SYMBOL_LOCK_HASH_ALGORITHM_VALUES)
 		_create_table(cursor, 'symbol_sync_state', SYMBOL_SYNC_STATE_DEFINITIONS)
 		_create_table(cursor, 'symbol_blocks', SYMBOL_BLOCK_DEFINITIONS)
 		_create_table(cursor, 'symbol_account_refresh_state', SYMBOL_ACCOUNT_REFRESH_STATE_DEFINITIONS)
@@ -436,6 +503,10 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 			cursor.execute(index_sql)
 		_create_table(cursor, 'symbol_metadata', SYMBOL_METADATA_DEFINITIONS)
 		for index_sql in SYMBOL_METADATA_INDEXES:
+			cursor.execute(index_sql)
+		_create_table(cursor, 'symbol_hash_locks', SYMBOL_HASH_LOCK_DEFINITIONS)
+		_create_table(cursor, 'symbol_secret_locks', SYMBOL_SECRET_LOCK_DEFINITIONS)
+		for index_sql in SYMBOL_LOCK_INDEXES:
 			cursor.execute(index_sql)
 		cursor.execute('CREATE SEQUENCE IF NOT EXISTS symbol_transaction_list_sequence_seq')
 		_create_table(cursor, 'symbol_transactions', SYMBOL_TRANSACTION_DEFINITIONS)
@@ -654,6 +725,19 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 				SymbolDatabase._execute_upsert_metadata(cursor, entry['row'])
 
 	@staticmethod
+	def _execute_hash_lock_entries(cursor, hash_lock_entries):
+		for entry in hash_lock_entries:
+			if 'hash' in entry:
+				SymbolDatabase._execute_delete_hash_lock(cursor, entry['hash'])
+			else:
+				SymbolDatabase._execute_upsert_hash_lock(cursor, entry['row'])
+
+	@staticmethod
+	def _execute_secret_lock_entries(cursor, secret_lock_entries):
+		for entry in secret_lock_entries:
+			SymbolDatabase._execute_replace_secret_locks(cursor, entry['key'], entry['rows'])
+
+	@staticmethod
 	def _execute_refresh_mosaic_alias_names(cursor, mosaic_ids):
 		for mosaic_id in mosaic_ids:
 			cursor.execute(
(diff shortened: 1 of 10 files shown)


explorer/puller/puller/model/symbol/Lock.py
```diff
@@ -0,0 +1,171 @@
...
+def create_hash_lock_key_from_hex(hash_value):
+	"""Creates a validated Hash Lock dirty key from a hexadecimal hash."""
+
+	if not isinstance(hash_value, str):
+		raise ValueError('Invalid Symbol Hash Lock key')
+
+	try:
+		return create_hash_lock_key(bytes.fromhex(hash_value))
+	except ValueError as exception:
+		raise ValueError('Invalid Symbol Hash Lock key') from exception
```
Jaguar0625, line 52: is there reason you wanted two functions instead of single one that would handle all valid types of input?

explorer/puller/puller/model/symbol/Lock.py
```diff
@@ -0,0 +1,171 @@
...
+	"""Creates a validated Secret Lock logical search key."""
+
+	if owner_address is not None and (not isinstance(owner_address, bytes) or 24 != len(owner_address)):
+		raise ValueError('Invalid Symbol Secret Lock owner address')
+	if not isinstance(recipient_address, bytes) or 24 != len(recipient_address):
+		raise ValueError('Invalid Symbol Secret Lock recipient address')
+	if not isinstance(secret, bytes) or 32 != len(secret):
+		raise ValueError('Invalid Symbol Secret Lock secret')
+	if hash_algorithm not in LOCK_HASH_ALGORITHM_LABELS.values():
+		raise ValueError(f'Unsupported Symbol Secret Lock hash algorithm {hash_algorithm}')
```
Jaguar0625: is there reason you don't want to use sdk types that should do these checks automatically during construction?

explorer/puller/tests/test/SymbolLockTestUtils.py
```diff
@@ -0,0 +1,60 @@
...
+	recipient_address='98' + '22' * 23,
+	secret='CC' * 32,
+	hash_algorithm=1,
+	mosaic_id='72C0212E67A08BCE',
+	amount='1234',
+	end_height='5678',
+	status=1,
+	item_id='secret-item',
+	lock_overrides=None,
+	**overrides
```
Jaguar0625: why do you need two overrides? don't these serve same purpose?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+			FROM information_schema.columns
+			WHERE table_name IN ('symbol_hash_locks', 'symbol_secret_locks')
+				AND column_name IN ('hash_algorithm', 'status')
+			ORDER BY table_name, column_name
+			''')
+		self.assertEqual([
+			('symbol_hash_locks', 'status', 'symbol_lock_status'),
+			('symbol_secret_locks', 'hash_algorithm', 'symbol_lock_hash_algorithm'),
+			('symbol_secret_locks', 'status', 'symbol_lock_status')
+		], cursor.fetchall())
```
Jaguar0625, line 4284: would prefer to move this (usage of enum) to separate test

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+
+	def test_create_tables_does_not_add_lock_column_defaults(self):
```
Jaguar0625, line 4286: does_not_allow_null_values ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+		initial_row = _create_hash_lock_row()
+		updated_row = _create_hash_lock_row(observed_height=2, end_height=200, status='used')
+
+		# Act:
+		database.upsert_hash_lock(initial_row)
+		database.upsert_hash_lock(updated_row)
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT hash, amount, end_height, status, updated_at_height FROM symbol_hash_locks')
+		updated_state = cursor.fetchall()
+		database.delete_hash_lock(create_hash_lock_key_from_hex(LOCK_HASH.hex()))
```
Jaguar0625: please move test of delete_hash_lock to own test

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+			create_secret_lock_search_key(LOCK_OWNER, LOCK_RECIPIENT, LOCK_SECRET, 'hash160'), [])
+
+		# Assert:
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT composite_hash, owner_address FROM symbol_secret_locks ORDER BY composite_hash')
+		self.assertEqual(
+			[(LOCK_COMPOSITE_HASH_2, LOCK_OWNER_2)],
+			[(bytes(composite_hash), bytes(owner_address)) for composite_hash, owner_address in cursor.fetchall()])
+
+	def test_secret_lock_unknown_owner_replace_removes_all_owner_siblings(self):
```
Jaguar0625, line 4394: i don't believe this is being tested. test should have some matching rows with non-null owner that will match with null (wildcard) owner.

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+				VALUES (%(composite_hash)s, %(owner_address)s, %(recipient_address)s, %(secret)s, %(hash_algorithm)s,
+				%(mosaic_id)s, %(amount)s, %(end_height)s, %(status)s, %(raw_payload)s, %(updated_at_height)s)''',
+			{**row, 'raw_payload': Json(row['raw_payload'])})
+		database.connection.commit()
+
+		# Act / Assert:
+		with self.assertRaisesRegex(ValueError, '^Invalid Symbol Secret Lock owner address$'):
+			database.get_secret_lock_search_keys_updated_from_height(1)
+
+	def test_get_hash_lock_hashes_updated_from_height_returns_ordered_boundary_rows(self):
```
Jaguar0625, line 4474: why do you want to order by hash instead of by height ?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+		secret_proof_transaction = create_transaction_entry(
+			2, 'secret-proof', type=TransactionType.SECRET_PROOF.value,
+			type_name=TRANSACTION_TYPE_LABELS[TransactionType.SECRET_PROOF.value],
+			signer_address=LOCK_OWNER_2, recipient_address=LOCK_RECIPIENT,
+			body={'secret': LOCK_SECRET.hex().upper(), 'hashAlgorithm': 0})
+		secret_proof_after = create_transaction_entry(
+			2, 'secret-proof-after', type=TransactionType.SECRET_PROOF.value,
+			type_name=TRANSACTION_TYPE_LABELS[TransactionType.SECRET_PROOF.value],
+			signer_address=LOCK_OWNER, recipient_address=LOCK_RECIPIENT,
+			body={'secret': LOCK_SECRET.hex().upper(), 'hashAlgorithm': 0})
```
Jaguar0625: are you intentionally testing that duplicates are collapsed in this test too?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -822,6 +906,198 @@ def get_metadata_keys_updated_from_height(self, height):  # pylint: disable=inva
...
+
+		hash_keys = {}
+		secret_keys = {}
+		for transaction_type, is_embedded, transaction_hash, signer_address, recipient_address, body in cursor.fetchall():
+			if TransactionType.HASH_LOCK.value == transaction_type:
+				hash_key = create_hash_lock_key_from_hex(body['hash'])
+				hash_keys[hash_key] = None
+			elif TransactionType.AGGREGATE_BONDED.value == transaction_type and not is_embedded:
+				hash_key = create_hash_lock_key_from_hex(bytes(transaction_hash).hex())
+				hash_keys[hash_key] = None
```
Jaguar0625: upstream will need to handle if someone sends hash_lock without aggregate bonded

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4007,3 +4055,730 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+		self.assertEqual([
+			(LOCK_COMPOSITE_HASH, 'unused', 1),
+			(LOCK_COMPOSITE_HASH_2, 'used', 1)
+		], [
+			(bytes(composite_hash), status, updated_at_height)
+			for composite_hash, status, updated_at_height in cursor.fetchall()])
+		self.assertEqual('repairing', database.get_sync_state()['status'])
+		self.assertEqual(1, database.get_sync_state()['last_synced_height'])
+
+	def test_repair_rollback_applies_hash_lock_delete_entries(self):
```
Jaguar0625, line 4651: are you missing equivalent test for secret lock?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -293,12 +302,26 @@ async def _repair_from_height(self, height, sync_state):
 		mosaic_entries = await self._fetch_dirty_mosaics(mosaic_ids, height - 1)
 		metadata_keys = self.symbol_db.get_metadata_keys_updated_from_height(height)
 		metadata_entries = await self._fetch_dirty_metadata(metadata_keys, height - 1)
+		rollback_transaction_keys = self.symbol_db.get_lock_keys_from_confirmed_transactions_from_height(height)
+		hash_lock_keys = self._union_hash_lock_keys(
+			self.symbol_db.get_hash_lock_hashes_updated_from_height(height),
+			rollback_transaction_keys.hash_keys)
+		secret_lock_keys = self._union_secret_lock_keys(
+			self.symbol_db.get_secret_lock_search_keys_updated_from_height(height),
+			rollback_transaction_keys.secret_keys)
```
Jaguar0625: can you explain your intent? why do you want to handle these lock transactions differently?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -900,6 +930,147 @@ def _write_dirty_metadata(self, entries):
...
 				self.symbol_db.upsert_metadata(entry['row'])
 
+	@staticmethod
+	def _assert_resolved_transaction_address(address, field_name):
+		if address is not None and Address(address).is_alias():
+			raise ValueError(f'Unresolved Symbol transaction {field_name} reached Lock dirty-key collection')
+
+	@staticmethod
+	def _union_hash_lock_keys(*key_lists):
+		keys = {}
```
Jaguar0625: can use set

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -900,6 +930,147 @@ def _write_dirty_metadata(self, entries):
...
+		return list(keys)
+
+	@staticmethod
+	def _union_secret_lock_keys(*key_lists):
+		keys = {}
+		for key_list in key_lists:
+			for key in key_list:
+				keys[key] = None
+
+		return list(keys)
```
Jaguar0625: these two are identical (aside from name)

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -900,6 +930,147 @@ def _write_dirty_metadata(self, entries):
...
+
+	def _collect_dirty_lock_keys_for_batch(self, block_rows, transaction_rows_by_height):
+		"""Collects Hash and Secret Lock dirty keys from transactions and expiring current rows."""
+
+		hash_keys = {}
+		secret_keys = {}
```
Jaguar0625: is there reason you don't want to use set here?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -900,6 +930,147 @@ def _write_dirty_metadata(self, entries):
...
+						lock_hash_algorithm_label(body['hashAlgorithm']))
+					secret_keys[secret_key] = None
+
+		start_height = block_rows[0]['height']
+		end_height = block_rows[-1]['height']
+		hash_keys.update({key: None for key in self.symbol_db.get_hash_lock_hashes_expiring_between(start_height, end_height)})
+		secret_keys.update({
+			key: None
+			for key in self.symbol_db.get_secret_lock_search_keys_expiring_between(start_height, end_height)
+		})
```
Jaguar0625: i am not sure i understand intent here? can you explain?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -900,6 +930,147 @@ def _write_dirty_metadata(self, entries):
...
+		return entries
+
+	async def _fetch_dirty_secret_locks(self, search_keys, observed_height):
+		"""Fetches and exactly filters paginated Secret Lock search results in bounded concurrent batches."""
+
+		async def fetch_entry(search_key):
+			page_number = 1
+			matching_rows = []
+			composite_hashes = set()
+			while True:
```
Jaguar0625, line 1001: this is correct, but will be quite inefficient. we should probably open issue against REST for nicer API.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			self.in_flight_lock_requests)
+		try:
+			if not self._release_scheduled:
+				self._release_scheduled = True
+				asyncio.get_running_loop().call_soon(self._requests_released.set)
+			await self._requests_released.wait()
+			response = self.secret_responses.get(url_path, {'data': []})
+			if isinstance(response, Exception):
+				raise response
+			return response
```
Jaguar0625: can be removed (handled by LockConnector)

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		finally:
+			self.in_flight_lock_requests -= 1
+
+
+class BoundedHashLockConnector(LockConnector):
```
Jaguar0625: refactor with previous; identical except for url path

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+	def test_sync_block_headers_rejects_secret_lock_search_response_without_data_without_writes(self):
+		self._assert_sync_rejects_malformed_secret_lock_response({}, '^Malformed Symbol Secret Lock search response$')
+
+	def test_sync_block_headers_rejects_secret_lock_search_response_with_non_list_data_without_writes(self):
+		self._assert_sync_rejects_malformed_secret_lock_response({'data': {}}, '^Malformed Symbol Secret Lock search response$')
+
+	def test_sync_block_headers_rejects_secret_lock_search_response_with_malformed_item_without_writes(self):
+		self._assert_sync_rejects_malformed_secret_lock_response({'data': [{}]}, '^Malformed Symbol Secret Lock response$')
+
+	def _assert_sync_rejects_second_secret_lock_page_without_writes(self, second_page_response, exception_type, error):
```
Jaguar0625: `without_writes` seems like it can be left off name?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			TransactionType.AGGREGATE_BONDED.value, {}, hash=bytes.fromhex(LOCK_HASH_2))
+
+		# Act:
+		keys = self._collect_lock_keys([top_level])
+
+		# Assert:
+		self.assertEqual([bytes.fromhex(LOCK_HASH_2)], [key.hash for key in keys.hash_keys])
+		self.assertEqual([], keys.secret_keys)
+
+	def test_collect_dirty_lock_keys_excludes_embedded_aggregate_bonded_hash(self):
```
Jaguar0625: unsure if needed since aggregate can never be embedded ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		# Act:
+		keys = self._collect_lock_keys([first, second, duplicate])
+
+		# Assert:
+		self.assertEqual([
+			(None, bytes.fromhex(RECIPIENT_ADDRESS), bytes.fromhex(SECRET), 'sha3_256'),
+			(bytes.fromhex(SIGNER_ADDRESS), bytes.fromhex(RECIPIENT_ADDRESS), bytes.fromhex(SECRET), 'hash160')
+		], keys.secret_keys)
+
+	def test_collect_dirty_lock_keys_includes_hash_lock_expiry_key(self):
```
Jaguar0625: what do you mean by "expiry key" ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			'amount': amount,
+			'end_height': end_height,
+			'status': status,
+			'raw_payload': raw_payload,
+			'updated_at_height': updated_at_height
+		} for composite_hash, owner_address, recipient_address, secret, hash_algorithm, mosaic_id, amount, end_height,
+			status, raw_payload, updated_at_height in cursor.fetchall()]
+		return hash_locks, secret_locks
+
+	def test_sync_block_headers_persists_hash_and_secret_locks(self):
```
Jaguar0625: i understand why you're combining hash and secret locks, but they are really separate transactions, so idk if it makes sense to test them together vs separately.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(([], [expected_row]), self._fetch_persisted_lock_state())
+		self.assertEqual([address_path, secret_only_path], [
+			path for path in connector.paths if path.startswith('lock/')
+		])
+
+	def test_sync_block_headers_converges_new_locks_created_and_expired_in_the_same_batch_to_empty(self):
```
Jaguar0625: intent is to only have explorer display active and completed locks?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			(1, TransactionType.HASH_LOCK.value, '1'),
+			(1, TransactionType.SECRET_LOCK.value, '1')
+		], cursor.fetchall())
+		self.assertEqual(2, self.puller.symbol_db.get_sync_state()['last_synced_height'])
+		self.assertEqual([
+			'lock/hash/' + LOCK_HASH,
+			secret_path
+		], [path for path in connector.paths if path.startswith('lock/')])
+
+	def test_sync_block_headers_updates_a_hash_lock_to_used_from_a_top_level_aggregate_bonded_transaction(self):
```
Jaguar0625: difference from previous test is that lock and unlock are not in same batch, correct ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			create_expected_secret_lock_row(
+				first_match, 1, bytes.fromhex('11' * 32), bytes.fromhex(SIGNER_ADDRESS),
+				bytes.fromhex(RECIPIENT_ADDRESS), bytes.fromhex(SECRET), 'hash160', MOSAIC_ID, 1234, 100, 'used'),
+			create_expected_secret_lock_row(
+				second_match, 1, bytes.fromhex('22' * 32), bytes.fromhex(other_owner),
+				bytes.fromhex(RECIPIENT_ADDRESS), bytes.fromhex(SECRET), 'hash160', MOSAIC_ID, 1234, 100, 'used'),
+			create_expected_secret_lock_row(
+				create_secret_lock_item(composite_hash='FF' * 32, secret=sibling_secret), 1, bytes.fromhex('FF' * 32),
+				bytes.fromhex(SIGNER_ADDRESS), bytes.fromhex(RECIPIENT_ADDRESS), bytes.fromhex(sibling_secret), 'hash160',
+				MOSAIC_ID, 1234, 100, 'unused')
```
Jaguar0625, line 1726: 44 and 55 are deleted existing siblings
11 and 22 come from secret request
FF is not deleted because it has different secret
correct?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(([], expected_rows), self._fetch_persisted_lock_state())
+		self.assertEqual([_secret_only_search_path(SECRET)], [
+			path for path in connector.paths if path.startswith('lock/')
+		])
+
+	def test_sync_block_headers_restores_a_deleted_hash_lock_from_rollback_transaction_key(self):
```
Jaguar0625: rollback - what is triggering rollback here?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			transactions_by_path={transaction_path(2, 2): {'data': []}},
+			statement_pages={statement_path(2, 2): {'data': []}})
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		cursor.execute('SELECT hash, status, updated_at_height FROM symbol_hash_locks')
+		lock_state = [
+			(bytes(lock_hash), status, updated_at_height)
```
Jaguar0625: shouldn't this be empty after call to delete_hash_lock ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		with self.assertRaisesRegex(RuntimeError, '^rollback lock fetch failed$'):
+			self._sync_with_connector(connector)
+
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute('SELECT height, type FROM symbol_transactions ORDER BY height, id')
+		self.assertEqual([(2, TransactionType.HASH_LOCK.value)], cursor.fetchall())
+		self.assertEqual([1, 2], self._fetch_block_heights(self.puller.symbol_db))
+		self.assertEqual(original_lock_state, self._fetch_lock_state())
+		self.assertEqual(original_sync_state, fetch_normalized_sync_state(self.puller.symbol_db))
```
Jaguar0625: can you use _fetch_complete_batch_state here ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+		rollback_connector = LockConnector(
+			2,
+			{1: [create_node_block(2, block_hash='FF' * 32)]},
+			{2: create_node_block(2, block_hash='FF' * 32)},
+			transactions_by_path={transaction_path(2, 2): {'data': []}},
+			statement_pages={statement_path(2, 2): {'data': []}})
+		self._sync_with_connector(rollback_connector)
+
+		# Assert:
+		self.assertEqual(([], []), self._fetch_lock_state())
```
Jaguar0625: if it rolled back to below end state, shouldn't lock entries be restored?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			{0: [create_node_block(1)]},
+			secret_responses={path: RuntimeError('secret lock fetch failed')},
+			transactions_by_path={transaction_path(1, 1): {'data': [transaction]}},
+			statement_pages={statement_path(1, 1): {'data': []}})
+
+		# Act / Assert:
+		self._assert_sync_failure_preserves_complete_batch_state(
+			connector, RuntimeError, '^secret lock fetch failed$')
+
+	def test_sync_block_headers_does_not_write_a_successful_hash_lock_when_a_later_secret_lock_fetch_fails(self):
```
Jaguar0625, line 1990: is there a special interaction you want to test here that is not covered in prior two tests?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1982 @@
...
+			secret_responses={secret_path: RuntimeError('secret lock fetch failed')},
+			transactions_by_path={transaction_path(1, 1): {'data': [hash_transaction, secret_transaction]}},
+			statement_pages={statement_path(1, 1): {'data': []}})
+
+		# Act / Assert:
+		self._assert_sync_failure_preserves_complete_batch_state(
+			connector, RuntimeError, '^secret lock fetch failed$')
+
+		# Assert:
+		self.assertEqual(['lock/hash/' + LOCK_HASH, secret_path], [path for path in connector.paths if path.startswith('lock/')])
```
Jaguar0625, line 2012: i assume you didn't modify `test_PullerRollback` and `test_PullerSync` because you added the tests you'd typically add there in this file? if so, that's fine, but just add comment at top of this file. 🙏

Jaguar0625 (commented):
👍

explorer/puller/puller/model/symbol/Lock.py
daoka: The two helpers originally represented the REST hex and database bytes boundaries. Since callers need both forms, I consolidated them into `create_hash_lock_key()`, which accepts either `bytes` or a hex string and normalizes through `Hash256(value).bytes` while preserving the existing domain error message.

explorer/puller/puller/model/symbol/Lock.py
daoka: The validation was originally manual to preserve domain-specific error messages. I changed it to use `Address` and `Hash256` as validators and immediately convert the validated values back to bytes, so SDK objects do not cross comparison or persistence boundaries. Decoded address hex values are handled with `Address.from_decoded_address_hex_string()`.

explorer/puller/tests/test/SymbolLockTestUtils.py
daoka: Agreed. The two override paths served the same purpose, so I removed `lock_overrides` and consolidated them into `**overrides`.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: I split the checks so the exact enum labels and the use of those enum types by the lock columns are verified in separate tests.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: I split the delete case from the upsert/update test and added an independent PostgreSQL-backed test for the public `delete_hash_lock()` API.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: I changed the setup to seed rows for two non-null owners using their concrete keys, then apply an empty replacement with `owner=None`. This now proves that the wildcard replacement removes both matching owner rows without using the wildcard path for seeding.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: I changed the queries to order primarily by height and use the key only as a deterministic tie-breaker: `updated_at_height, hash` or `end_height, hash`, with `composite_hash` for Secret Locks. The tests include same-height rows inserted in reverse key order to verify the tie-break explicitly.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes, the original test covered both source selection and deduplication. I split deduplication into its own test and removed the impossible embedded Aggregate Bonded fixture from the source-selection test.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: I added the equivalent rollback test for an empty Secret Lock replacement. It verifies that the target logical key is deleted while an unrelated sibling and the repairing sync state are preserved in the same repair transaction.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Agreed. I removed the duplicate Hash and Secret Lock union helpers.

The normal-batch collector now deduplicates both key types directly with sets. Rollback and finalization obtain their keys from their respective database queries, so a shared union helper is no longer necessary.

explorer/puller/puller/facade/SymbolPuller.py
daoka: These keys are used for finalization-driven Lock cleanup.

Symbol does not remove a Lock when it merely reaches `endHeight`. Expiration and completion update balances, receipts, or status, while the current Lock row is pruned only after finalization reaches its end height.

The puller therefore selects persisted rows with `end_height <= finalized_height`, re-fetches their current state from the node, and removes a row only when the node no longer returns it. This cleanup is independent of the current block batch and also runs when no new blocks are synchronized.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Agreed. The bounded connector now delegates to `super().get()` after waiting, so response handling and path recording are both owned by `LockConnector` and each path is recorded only once.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: I consolidated the two connector classes into a single `BoundedLockConnector` that accepts the target URL prefix.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Agreed. I removed the `without_writes` suffix from the public test names and the second-page helper. The tests still verify the complete batch snapshot remains unchanged through the assertion helper.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Correct. Aggregate Bonded cannot be embedded, so I removed the embedded guard from both collection paths, removed the unused `is_embedded` database column selection, and deleted the impossible fixtures.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: I removed the `expiry key` terminology and the batch-range expiry-key tests because they did not describe the actual Symbol Lock lifecycle.

The replacement tests verify finalization-driven cleanup: both used and unused Locks remain in current state until finalization reaches their end height, and are deleted only when a subsequent node fetch no longer returns them. The tests also cover cleanup without new blocks and cleanup deferred by `max_height`.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: I split normal persistence and restart convergence into separate Hash Lock and Secret Lock tests. Combined tests remain only where the behavior under test is cross-type atomicity.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: The Explorer stores the node's current Lock state rather than historical Lock entries.

This includes both unused (active) and used (completed) Locks. Expiration and completion do not immediately remove a Lock; completion changes its status to `Used`, and pruning occurs only when finalization reaches its end height.

The puller removes the local row only when finalization cleanup re-fetches the Lock and the node no longer returns it. The tests cover this behavior for both used and unused Hash and Secret Locks.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. These tests cover an existing lock whose completion arrives in a later batch. I renamed both the Hash Lock and Secret Lock tests so that distinction is explicit.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes, that interpretation is correct. I replaced the numeric composite hashes with named variables for the existing owner siblings, node replacement rows, and the different-secret survivor so the fixture roles are explicit.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: The rollback is triggered by the mismatch between the local block 2 hash and the node block 2 hash. I added the same explanation to all rollback tests using this fixture.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes, the table is empty immediately after `delete_hash_lock()`. That intermediate assertion was in the Arrange phase and made the final restoration assertion confusing. I removed it from all four sibling rollback tests; deletion is covered by the database test, while these tests now assert the restored state only after the sync action.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. I replaced the individual lock, sync-state, block, and transaction assertions with a before-and-after comparison using `_fetch_complete_batch_state()` so the complete batch state is covered consistently.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. The previous two tests cover failure atomicity for each lock type independently. This test covers the cross-type case where the Hash Lock fetch succeeds and the later Secret Lock fetch fails, proving that no successful Hash Lock result is written before all lock fetches complete.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes, that is the intent. I added a module docstring explaining that the Symbol Lock-specific integration cases normally associated with sync and rollback are intentionally consolidated in this file.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: The tests `test_create_tables_creates_exact_hash_lock_columns_types_and_nullability`
and `test_create_tables_creates_exact_secret_lock_columns_types_and_nullability`
already assert `is_nullable = 'NO'` for every lock column.

I kept `test_create_tables_does_not_add_lock_column_defaults` separate because
it verifies a different DDL contract: required lock values must be supplied
explicitly rather than being filled by implicit column defaults.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thank you for questioning this. My original intent was to recover Lock keys from rollback-range transactions when a Lock row had already disappeared from the current-state tables. That was based on an incorrect assumption that Symbol removes a Lock when its end height is reached.

I rechecked the node/core lifecycle. Expiry processing and completion do not remove the Lock. Both Used and Unused Lock histories are removed only when finalized-height pruning reaches `endHeight`. Since rollback repair only operates above the finalized height, every repairable Lock still has a current row when its key is collected. Therefore, Lock transactions do not need a separate transaction-derived rollback-key path.

I updated the implementation accordingly:

- `_repair_from_height()` now collects Lock keys only from current rows with `updated_at_height >= fork_height`.
- `get_lock_keys_from_confirmed_transactions_from_height()` and the key-union path were removed.
- Finalized Lock removal is reconciled separately on every sync, including runs with no new blocks, by rechecking rows with `end_height <= finalized_height` until REST no longer returns them.
- Hash and Secret Lock cleanup is applied atomically.

There is a separate pre-existing Metadata rollback case where an orphaned-branch transaction can delete the current row and leave no natural key to collect. I will track that separately rather than retaining an incorrect Lock-specific workaround in this PR.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Yes. I verified this against the Symbol core implementation. A Hash Lock can be confirmed even if the corresponding Aggregate Bonded transaction is never announced. The Hash Lock validators do not require the aggregate to exist, and `HashLockObserver` inserts the lock keyed by the Aggregate Bonded transaction hash.

If the Aggregate Bonded transaction arrives later, `CompletedAggregateObserver` marks the existing lock as used. If it never arrives, the unused lock is refunded at expiry and subsequently pruned after finalization. Conversely, an Aggregate Bonded transaction without an active matching Hash Lock is rejected by `AggregateHashPresentValidator`.

The puller therefore persists Hash Lock transactions independently and treats Aggregate Bonded transactions only as later state-change signals. This behavior is now made explicit by `test_sync_block_headers_persists_a_hash_lock_without_an_aggregate_bonded_transaction`.

The rollback transaction-key scan shown in this old hunk has since been removed; current rollback recovery uses persisted current Lock rows only.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thanks. The `_union_*_lock_keys` helpers were removed when the rollback and finalization cleanup paths were redesigned. For the remaining normal-batch dirty-key collection, I switched both Hash and Secret Lock keys to sets and convert them to lists only at the bounded-fetch boundary.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thanks. I rechecked the ordering requirement against the Symbol core and REST behavior. Transaction ordering has already been applied before the puller reads the current state, and each Lock key is fetched independently, so cross-key ordering is not part of the correctness or operational contract.

I changed both collectors to use sets and convert them to lists only before the slice-based bounded-fetch helpers. The collector tests now assert unique membership, while the bounded-fetch tests still verify that an explicitly supplied input sequence is preserved. I also made the same-batch Secret Lock request assertion order-independent and verified the tests with multiple `PYTHONHASHSEED` values

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: You are right to question this. I rechecked the Symbol core implementation and found that the original test was based on an incorrect lifecycle assumption.

Reaching `endHeight` does not remove a Lock. Expiration and completion only update balances, receipts, or status. Both used and unused Lock histories are removed only when finalized-height pruning reaches their end height.

Rollback verification starts at `finalized_height + 1`, so a Lock that is still within the repairable rollback range cannot already have been removed by finalization pruning. Its current row remains available as the rollback key.

I removed the residual-window tests and the batch-end pruning assumption. Cleanup now rechecks rows with `end_height <= finalized_height`, while rollback restoration uses current rows with `updated_at_height >= fork_height`. The updated tests cover current-row restoration, fork-only row deletion, survivor preservation, and finalization-driven removal for both used and unused Locks.

explorer/puller/puller/facade/SymbolPuller.py
daoka: I opened an issue https://github.com/symbol/symbol/issues/2127
This PR keeps the current paginated implementation until that REST API is available.

explorer/puller/tests/db/test_SymbolDatabase.py
Jaguar0625: what is the new deduplication test?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+		self.puller.symbol_db.connection.commit()
+		connector = self._finalization_connector(
+			hash_responses={'lock/hash/' + LOCK_HASH: {'code': 'ResourceNotFound', 'message': 'not found'}},
+			secret_responses={_secret_search_path(SIGNER_ADDRESS, SECRET): {'data': [create_secret_lock_item()]}})
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual([], self._fetch_persisted_lock_state()[0])
```
Jaguar0625: do you want to check connector.paths here and in other new tests like in previous test ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+			hash_responses={'lock/hash/' + LOCK_HASH: create_hash_lock_item()},
+			secret_responses={_secret_search_path(SIGNER_ADDRESS, SECRET): {'data': []}})
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual([], self._fetch_persisted_lock_state()[1])
+
+	def test_sync_block_headers_runs_finalization_cleanup_without_new_blocks(self):
```
Jaguar0625: test is ok, but don't all the previous tests also have no new blocks ?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		hash_rows, secret_rows = self._fetch_persisted_lock_state()
+		self.assertEqual(1, len(hash_rows))
+		self.assertEqual(1, len(secret_rows))
+
+	def test_sync_block_headers_reselects_a_still_present_finalized_lock_on_the_next_run(self):
```
Jaguar0625: i am not sure what you mean by "reselects". what behavior are you specifically testing with the two syncs? that lock is removed even though only second has pruned it?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+			finalized_height=3,
+			hash_responses={'lock/hash/' + LOCK_HASH_2: {'code': 'ResourceNotFound', 'message': 'not found'}},
+			transactions_by_path={transaction_path(3, 3): {'data': []}},
+			statement_pages={statement_path(3, 3): {'data': []}})
+		self._sync_with_connector(uncapped_connector)
+
+		# Assert:
+		self.assertEqual(([], []), self._fetch_persisted_lock_state())
+
+	def test_sync_block_headers_rolls_back_finalization_lock_writes_after_secret_constraint_failure(self):
```
Jaguar0625: this test seems a bit contrived, what is the purpose of it?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(([], expected_rows), self._fetch_persisted_lock_state())
+		self.assertEqual([_secret_only_search_path(SECRET)], [
+			path for path in connector.paths if path.startswith('lock/')
+		])
+
+	def test_sync_block_headers_restores_a_deleted_hash_lock_from_a_current_row_key(self):
```
Jaguar0625: main assertion is that observed_height is changed from 2 to 1, correct?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+		hash_rows, secret_rows = self._fetch_persisted_lock_state()
+		self.assertEqual([expected_survivor_hash], hash_rows)
+		self.assertEqual([expected_survivor_secret], secret_rows)
+		self.assertEqual([fork_hash_path, fork_secret_path], [
+			path for path in connector.paths if path.startswith('lock/')])
+		self.assertEqual([1, 2], self._fetch_block_heights(self.puller.symbol_db))
+		self.assertEqual(bytes.fromhex(f'{2:064X}'), self._fetch_block_hash(self.puller.symbol_db, 2))
+		actual_sync_state = self.puller.symbol_db.get_sync_state()
+		actual_sync_state['finalized_hash'] = bytes(actual_sync_state['finalized_hash'])
+		actual_sync_state['last_synced_block_hash'] = bytes(actual_sync_state['last_synced_block_hash'])
```
Jaguar0625: why the conversions instead of comparing hex?

explorer/puller/tests/facade/symbol/test_PullerLocks.py
```diff
@@ -0,0 +1,1992 @@
...
+			hash_responses={'lock/hash/' + LOCK_HASH: RuntimeError('rollback lock fetch failed')})
+
+		# Act / Assert:
+		with self.assertRaisesRegex(RuntimeError, '^rollback lock fetch failed$'):
+			self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual(original_state, self._fetch_complete_batch_state())
+
+	def test_sync_block_headers_keeps_all_rollback_state_unchanged_when_secret_lock_replacement_fetch_fails(self):
```
Jaguar0625, line 1927: this seems like stricter test than previous (for hash lock), should you make that one stricter too?

Jaguar0625 (commented):
👍

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Added `test_collect_dirty_lock_keys_deduplicates_hash_keys` and `test_collect_dirty_lock_keys_deduplicates_secret_keys`. They explicitly verify that duplicate Hash and Secret Lock keys are collapsed before fetching.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. I made the Hash and Secret failure tests symmetric. Both now use the shared `_seed_rollback_batch_state()` fixture, compare the complete batch state before and after the failed synchronization, and assert the exact Lock request paths. I also aligned their test names.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: The conversions were only needed to normalize binary hash values returned by the database before comparing the complete sync state. I replaced the inline conversions with the existing `fetch_normalized_sync_state()` helper, which consistently normalizes `finalized_hash` and `last_synced_block_hash` to bytes.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. More specifically, the important transition is the persisted `updated_at_height` changing from 2 to 1. The broader contract is that the entire rollback-affected row is replaced with canonical node state observed at the pre-fork height. I renamed and strengthened both the Hash and Secret tests symmetrically to make that intent explicit.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: The original purpose was to force a write failure and verify cleanup atomicity. However, mutating the production schema in a Facade test made the scenario contrived and duplicated coverage at the database boundary, so I removed it. Atomicity remains covered directly by `test_apply_finalization_lock_entries_rolls_back_hash_and_secret_writes_together` using the real PostgreSQL transaction boundary.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: The intended behavior is that finalization cleanup rechecks an eligible Lock on every synchronization while the node still returns it. The first synchronization refreshes and retains the canonical state; the second removes it after the node drops it. I renamed the test and added complete state and exact path assertions after both runs.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. The no-new-block condition was duplicated. I consolidated it into one test that removes both unused Hash and Secret Locks in the same run and explicitly verifies that no `blocks?` request is made. The used Hash and Secret cases retain their separate cross-type non-interference coverage.

explorer/puller/tests/facade/symbol/test_PullerLocks.py
daoka: Yes. I added exact Lock request-path assertions to the finalization behavior tests. This includes both runs of the `max_height` scenario: the capped run does not request the deferred Lock, and the later uncapped run requests only that Lock.

Jaguar0625 (approved):
👍

PR #2464, opened 2026-07-31
[explorer/puller] feat: Added detect fork heigh logic

AnthonyLaw:
Problem: It's missing the rollback feature.
Solution: Added a detection for fork height logic. It'll compare the block hash from the chain versus the database.

Note: PRs will be split into small parts for easy review.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -139,6 +139,16 @@ class NemPuller:
 			delay
 		)
 
+	async def _retry_get_block(self, height, retries=3, delay=2):
+		"""Retries fetching a block at a specific height with exponential backoff."""
+
+		return await self._retry_operation(
+			lambda: self.nem_connector.get_block(height),
+			f'fetching block at height {height}',
+			retries,
+			delay
+		)
+
 	async def _retry_get_account_info(self, address, retries=3, delay=2):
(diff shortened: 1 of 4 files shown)


explorer/puller/puller/facade/NemPuller.py
```diff
@@ -139,6 +141,29 @@ async def _retry_get_blocks_after(self, height, retries=3, delay=2):
...
+		comparison_height = min(db_height, chain_height)
+		minimum_height = max(1, db_height - NEM_MAX_ROLLBACK_DEPTH)
+
+		for height in range(comparison_height, minimum_height - 1, -1):
+			db_block_hash = self.nem_db.get_block_hash(height)
+			node_block = await self._retry_get_block(height)
+
+			if db_block_hash.lower() == node_block.block_hash.lower():
+				return None if height == db_height else height
+
```
Wayonb, line 170: What happens when hash never matches in the 360 blocks?

Wayonb (changes requested):

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Hmm, in this case, I’ll raise an error so we can investigate it manually. Technically, this is unlikely to happen because we synchronize blocks every 30 seconds.

Wayonb (approved):
👍🏾

PR #2465, opened 2026-08-02
[explorer/puller] fix: recover Symbol metadata state during rollback

daoka:

Symbol Metadata rollback recovery derives natural keys only from current
`symbol_metadata` rows updated at or after the fork height.

Mosaic Metadata transactions preserve the unresolved target mosaic ID in their
body, but do not persist the resolved target ID needed to reconstruct the
canonical Metadata natural key.

This is a follow-up to #2463.

## What's the issue?

An orphaned Metadata update can leave an empty value. Symbol core removes the current Metadata entry in that case.

If the Explorer current-state row has already been removed, rollback recovery has no natural key with which to query the canonical Node state. The missing Metadata entry therefore cannot be restored from current rows alone.

Mosaic aliases require additional handling because the raw transaction body contains the unresolved mosaic ID, while the Metadata natural key uses the resolved mosaic ID.

This change persists the resolved Mosaic Metadata target as a private
`symbol_transaction_mosaics` relation:

- `role = 'metadata_target'`
- `amount = 0`
- `position = 0`
- exactly one relation for each Mosaic Metadata transaction

Account and Namespace Metadata transactions do not create this relation. The unresolved values in `body` and `raw_payload` remain unchanged.

Rollback recovery now:

1. Collects Metadata keys from current-state rows.
2. Reconstructs Metadata keys from confirmed transactions at or after the fork height.
3. Uses the private resolved relation for Mosaic Metadata instead of the raw `body.targetMosaicId`.
4. Canonicalizes scoped Metadata keys and non-null target IDs to uppercase 16-character hexadecimal strings.
5. Deduplicates both key sources using the complete Metadata natural key.
6. Fetches all canonical state before invoking the existing atomic rollback repair.

Malformed transaction bodies, missing or duplicate target relations, invalid sentinels, and malformed hexadecimal identifiers fail deterministically before rollback writes begin.

The exactly-one relation contract is enforced by application validation rather than a partial unique index. The existing relation primary key permits different mosaic IDs for the same transaction and role, and retaining application validation allows malformed persisted state to be detected and tested explicitly.

`metadata_target` is private persistence state. It must not be exposed through public `transferMosaicId` searches or Mosaic-detail transaction results.

There is no public API breaking change. This updates the pre-release declarative schema without adding an `ALTER TYPE` compatibility shim. Existing development databases with the old enum or non-canonical Metadata keys must be recreated.

REST, frontend, NEM, and the merged Lock synchronization behavior are unchanged.


explorer/puller/tests/model/symbol/test_Metadata.py
```diff
@@ -57,6 +84,112 @@ def test_create_metadata_row_preserves_namespace_target_id(self):
...
+			scoped_metadata_key='1234567890abcdef')
+
+		# Act:
+		row = create_metadata_row(item, 123)
+
+		# Assert:
+		self.assertEqual('1234567890ABCDEF', row['scoped_metadata_key'])
+		self.assertEqual('ABCDEF0123456789', row['target_id'])
+
+	def test_canonical_metadata_hex_rejects_invalid_values(self):
```
Jaguar0625: should you have a separate success test for canonical_metadata_hex ?

explorer/puller/tests/model/symbol/test_Metadata.py
```diff
@@ -57,6 +84,112 @@ def test_create_metadata_row_preserves_namespace_target_id(self):
...
+			'metadata_type': 'account',
+			'source_address': b'source',
+			'target_address': b'target',
+			'scoped_metadata_key': '0000000000000001',
+			'target_id': '0000000000000002'
+		}
+
+		# Act + Assert:
+		with self.assertRaisesRegex(ValueError, '^Invalid Symbol account Metadata target id$'):
+			canonical_metadata_key(key)
```
Jaguar0625, line 143: should you have separate success test for canonical_metadata_key ?

explorer/puller/tests/model/symbol/test_Metadata.py
```diff
@@ -2,11 +2,38 @@
 
 from symbolchain.sc import TransactionType
 
-from puller.model.symbol.Metadata import METADATA_TRANSACTION_TYPE_LABELS, METADATA_TYPE_LABELS, METADATA_TYPE_NUMBERS, create_metadata_row
+from puller.model.symbol.Metadata import (
+	METADATA_TRANSACTION_TYPE_LABELS,
```
Jaguar0625, line 6: consider grouping related tests in regions, e.g.

```py
# region canonical_metadata_hex

...

# endregion
```

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -870,21 +876,23 @@ def _collect_dirty_metadata_keys_for_batch(transaction_rows_by_height):
...
+				if 'namespace' == metadata_type:
+					target_id = body.get('targetNamespaceId')
 
-				key = {
+				key = canonical_metadata_key({
 					'metadata_type': metadata_type,
 					'source_address': transaction_row['signer_address'],
 					'target_address': transaction_row['target_address'],
-					'scoped_metadata_key': body['scopedMetadataKey'],
+					'scoped_metadata_key': body.get('scopedMetadataKey'),
```
Jaguar0625: why did you change this to optional?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -1361,6 +1368,169 @@ def test_get_metadata_keys_updated_from_height_returns_exact_keys_in_composite_h
...
+		self.assertEqual([
+			{'metadata_type': 'account', 'source_address': b'signer-account', 'target_address': b'target-account',
+				'scoped_metadata_key': 'ABCDEF0123456789', 'target_id': None},
+			{'metadata_type': 'mosaic', 'source_address': b'signer-mosaic', 'target_address': b'target-mosaic',
+				'scoped_metadata_key': '1234567890ABCDEF', 'target_id': '72C0212E67A08BCE'},
+			{'metadata_type': 'namespace', 'source_address': b'signer-namespace', 'target_address': b'target-namespace',
+				'scoped_metadata_key': '0000000000000003', 'target_id': 'A95F1F8A96159516'}
+		], keys)
+
+	def test_get_confirmed_metadata_keys_since_rejects_invalid_relation_cardinality(self):
```
Jaguar0625, line 1434: why are you doing all the validation on get vs when the database is being updated?

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
```diff
@@ -591,8 +688,8 @@ def test_sync_block_headers_persists_direct_mosaic_metadata_from_embedded_transa
 				'version': 1,
 				'network': 152,
 				'targetAddress': RECIPIENT_ADDRESS,
-				'targetMosaicId': MOSAIC_ID,
-				'scopedMetadataKey': SCOPED_METADATA_KEY,
+				'targetMosaicId': MOSAIC_ID.lower(),
+				'scopedMetadataKey': lowercase_scoped_key,
```
Jaguar0625, line 692: is there reason you don't want to store canonical version in db?

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -369,6 +433,219 @@ def test_sync_block_headers_leaves_metadata_rollback_state_unchanged_when_metada
...
+			metadata_type='namespace',
+			target_id='A95F1F8A96159516',
+			scoped_metadata_key='0000000000000010',
+			value_utf8='hello')
+
+		# Act:
+		asyncio.run(self.puller.sync_block_headers())
+
+		# Assert:
+		self._assert_metadata_rollback_recovery(connector, metadata_key_path, expected_metadata_row)
```
Jaguar0625, line 496: check last_synced_height like in previous test?

explorer/puller/tests/facade/symbol/test_PullerRollback.py
```diff
@@ -369,6 +433,219 @@ def test_sync_block_headers_leaves_metadata_rollback_state_unchanged_when_metada
...
+			metadata_type='mosaic',
+			target_id='72C0212E67A08BCE',
+			scoped_metadata_key='0000000000000012',
+			value_utf8='hello')
+
+		# Act:
+		asyncio.run(self.puller.sync_block_headers())
+
+		# Assert:
+		self._assert_metadata_rollback_recovery(connector, metadata_key_path, expected_metadata_row)
```
Jaguar0625, line 568: how is this testing deduplication? because there is only single metadata_key_path ?

Jaguar0625 (commented):
👍

explorer/puller/tests/model/symbol/test_Metadata.py
daoka: Added explicit happy-path tests for `canonical_metadata_hex()`. They verify that lowercase hexadecimal input is normalized to uppercase and that an already canonical value remains unchanged.

explorer/puller/tests/model/symbol/test_Metadata.py
daoka: Added explicit happy-path tests for `canonical_metadata_key()`. The Mosaic case verifies the complete canonicalized dictionary, including preservation of an unrelated `composite_hash` field. The Account case verifies that `target_id=None` remains valid and unchanged.

explorer/puller/tests/model/symbol/test_Metadata.py
daoka: Grouped the Metadata model tests into regions following the definition order in `Metadata.py`. The canonicalization assertion helpers and the previously separated `create_metadata_row` tests are now located in their corresponding regions.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Good point. This path reads an in-memory transaction body created from the Node DTO during the current synchronization cycle, so these protocol-required fields should not be treated as optional. I restored bracket access for `targetNamespaceId` and `scopedMetadataKey`. The defensive `.get()` access remains only in the database recovery path that reads persisted JSONB.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Validation is not limited to reads. Row-level invariants are validated through `canonical_metadata_key()` when Metadata rows are updated or deleted. During normal synchronization, `_collect_dirty_metadata_keys_for_batch()` also verifies that Mosaic Metadata has exactly one `metadata_target` relation, that Account and Namespace Metadata have none, and that the relation sentinel values are valid before `_sync_block_batch()` writes to the database.  
  
Rollback recovery validates these persisted relations again because their cardinality and applicability are not enforced by the existing database schema. This prevents rollback repair from proceeding with malformed relation data.

explorer/puller/tests/facade/symbol/test_PullerMetadata.py
daoka: The canonical value is stored in the database, but in the normalized `symbol_transaction_mosaics` relation with `role='metadata_target'`, rather than in `body`.  
  
`body['targetMosaicId']` intentionally retains the value from the Node DTO, which can be an unresolved alias. Overwriting it would lose that original value, while adding another resolved field to `body` would duplicate the relation and introduce an Explorer-specific field that is not part of the Node DTO. Normal synchronization and rollback recovery therefore use the canonical relation value as the target ID.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: Added the `last_synced_height` assertion to the shared rollback recovery assertion helper, so all six successful Metadata recovery tests now verify that the synchronization watermark advances to height 3. The duplicate assertion in the Account Metadata test was removed.

explorer/puller/tests/facade/symbol/test_PullerRollback.py
daoka: The existing single-request assertion proves the final deduplicated result, but it did not independently prove that both key sources contributed. I updated both the Mosaic and Namespace tests to read the current-row keys and confirmed-transaction keys before rollback, verify that each source produces the same single natural key, and then verify that synchronization performs only one exact-key Metadata request.

Jaguar0625 (approved):
👍

PR #2473, opened 2026-08-04
[explorer/puller] feat: add Symbol mosaic restriction state synchronization

daoka:
Added Symbol Mosaic Restriction current-state synchronization to the puller.

The implementation:

- adds strict normalization for Address and Global Mosaic Restriction responses;
- validates entry types, required fields, uint64 boundaries, removal sentinels, and resolved current-state identifiers;
- adds the `symbol_mosaic_restrictions` PostgreSQL table with:
  - a finite entry-type enum;
  - logical-key uniqueness;
  - address/global target consistency;
  - structured restrictions and raw payload persistence;
- collects dirty keys from confirmed Mosaic Restriction transactions using resolved position-0 mosaic relations and resolved target addresses;
- performs exact filtered REST requests for each logical restriction key;
- atomically replaces, updates, or deletes current-state rows;
- restores rollback-affected restrictions using the union of:
  - current rows updated in the rollback range;
  - persisted transaction child relations from the orphaned branch;
- preserves restart convergence and `fork_height - 1` observation semantics;
- adds a shared bounded asynchronous collection helper used by Metadata, Locks, and Mosaic Restrictions;
- strengthens Lock alias invariants after sharing the resolved-mosaic validation path.


explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -976,6 +1016,123 @@ def _metadata_key_from_confirmed_transaction_rows(rows):
...
+			WHERE entry_type = %(entry_type)s
+				AND mosaic_id = %(mosaic_id)s
+				AND target_address IS NOT DISTINCT FROM %(target_address)s
+			''',
+			{
+				'entry_type': entry_type,
+				'mosaic_id': restriction_key.mosaic_id,
+				'target_address': restriction_key.target_address
+			})
+		for restriction_row in restriction_rows:
```
Jaguar0625, line 1042: any reason not to use execute_values ?

explorer/puller/puller/db/SymbolDatabase.py
```diff
@@ -976,6 +1016,123 @@ def _metadata_key_from_confirmed_transaction_rows(rows):
...
+	def _execute_replace_mosaic_restrictions(cursor, restriction_key, restriction_rows):
+		if not isinstance(restriction_key, MosaicRestrictionKey):
+			raise ValueError('Invalid Symbol mosaic restriction logical key')
+		entry_type = mosaic_restriction_entry_type_label(restriction_key.entry_type)
+		cursor.execute(
+			'''
+			DELETE FROM symbol_mosaic_restrictions
+			WHERE entry_type = %(entry_type)s
+				AND mosaic_id = %(mosaic_id)s
+				AND target_address IS NOT DISTINCT FROM %(target_address)s
```
Jaguar0625, line 1035: what is purpose of this condition? aren't you reinserting everything below?

explorer/puller/tests/facade/test_async_utils.py
```diff
@@ -0,0 +1,28 @@
...
+	def test_gather_in_chunks_rejects_non_positive_or_boolean_chunk_size(self):
+		# Act + Assert:
+		async def fetch_item(item):
+			return item
+
+		for chunk_size in (0, -1, True, '10'):
+			with self.subTest(chunk_size=chunk_size), self.assertRaises(ValueError):
+				asyncio.run(gather_in_chunks([], chunk_size, fetch_item))
+
+	def test_gather_in_chunks_preserves_the_explicit_non_sorted_sequence(self):
```
Jaguar0625: consider adding a test with 3+ chunks

explorer/puller/puller/model/symbol/format.py
```diff
@@ -13,6 +14,40 @@ def bytes_from_hex_or_none(value):
...
+	if not isinstance(value, str) or not is_hex_text(value):
+		raise ValueError(f'Invalid {context_name} {field_name}')
+
+	try:
+		return Address.from_decoded_address_hex_string(value).bytes
+	except ValueError as exception:
+		raise ValueError(f'Invalid {context_name} {field_name}') from exception
+
+
+def hash_bytes_from_hex(node, field_name, context_name):
```
Jaguar0625, line 38: consider refactoring as this is identical to previous except for sdk type

explorer/puller/puller/model/symbol/MosaicRestriction.py
```diff
@@ -0,0 +1,189 @@
...
+
+	ADDRESS = 'address'
+	GLOBAL = 'global'
+
+
+MosaicRestrictionKey = namedtuple(
+	'MosaicRestrictionKey',
+	['entry_type', 'mosaic_id', 'target_address'])
+
+_ENTRY_TYPE_BY_NODE_VALUE = {0: MosaicRestrictionEntryType.ADDRESS, 1: MosaicRestrictionEntryType.GLOBAL}
```
Jaguar0625: not sure why you are using the name "NODE" vs something like "ENUM" ?

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
```diff
@@ -0,0 +1,591 @@
...
+
+	def test_key_rejects_alias_address(self):
+		self._assert_key_rejects(
+			MosaicRestrictionEntryType.ADDRESS,
+			'72C0212E67A08BCE',
+			bytes([1]) + bytes(23))
+
+	def test_key_rejects_alias_mosaic_id(self):
+		# Act + Assert:
+		with self.assertRaises(ValueError):
```
Jaguar0625: can you use _assert_key_rejects ?

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
```diff
@@ -0,0 +1,591 @@
...
+		self._assert_key_rejects(MosaicRestrictionEntryType.ADDRESS, '72C0212E67A08BCE', 'not-bytes')
+
+	def test_key_rejects_empty_address(self):
+		self._assert_key_rejects(MosaicRestrictionEntryType.ADDRESS, '72C0212E67A08BCE', b'')
+
+	def test_key_rejects_alias_address(self):
+		self._assert_key_rejects(
+			MosaicRestrictionEntryType.ADDRESS,
+			'72C0212E67A08BCE',
+			bytes([1]) + bytes(23))
```
Jaguar0625, line 195: include "global" or "address" in all test names as appropriate

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
```diff
@@ -0,0 +1,591 @@
...
+
+		# Act + Assert:
+		with self.assertRaises(ValueError):
+			create_mosaic_restriction_row(item, 1)
+
+	def test_model_rejects_boolean_version(self):
+		self._assert_model_rejects(create_restriction_item(version=True))
+
+	def test_model_rejects_false_version(self):
+		self._assert_model_rejects(create_restriction_item(version=False))
```
Jaguar0625: can have single test with both True/False asserts

explorer/puller/puller/model/symbol/MosaicRestriction.py
```diff
@@ -0,0 +1,189 @@
...
+MOSAIC_RESTRICTION_ENTRY_TYPE_BY_TRANSACTION_TYPE = {
+	TransactionType.MOSAIC_ADDRESS_RESTRICTION.value: MosaicRestrictionEntryType.ADDRESS,
+	TransactionType.MOSAIC_GLOBAL_RESTRICTION.value: MosaicRestrictionEntryType.GLOBAL
+}
+
+
+def mosaic_restriction_entry_type_from_node(value):  # pylint: disable=invalid-name
+	"""Map the strict Symbol node entryType integer to the local enum."""
+
+	if not isinstance(value, int) or isinstance(value, bool) or value not in _ENTRY_TYPE_BY_NODE_VALUE:
```
Jaguar0625: consider refactoring ` not isinstance(value, int) or isinstance(value, bool) ` into own predicate since it appears in many places

explorer/puller/puller/model/symbol/MosaicRestriction.py
```diff
@@ -0,0 +1,189 @@
...
+	if MosaicRestrictionType.NONE == validated_restriction_type:
+		raise ValueError('Invalid Symbol global mosaic restriction restrictionType')
+
+
+def _is_decimal_u64(value):
+	if not isinstance(value, str) or not value.isascii() or not value.isdecimal():
+		return False
+	normalized = value.lstrip('0') or '0'
+	return len(normalized) < 20 or (
+		len(normalized) == 20 and normalized <= '18446744073709551615')
```
Jaguar0625: use MOSAIC_ADDRESS_RESTRICTION_REMOVAL_SENTINEL ?

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
```diff
@@ -0,0 +1,591 @@
...
+		item = create_restriction_item(1)
+		item['mosaicRestrictionEntry']['restrictions'][0]['restriction']['restrictionType'] = '1'
+		self._assert_model_rejects(item)
+
+	def test_model_rejects_unknown_global_restriction_type(self):
+		item = create_restriction_item(1)
+		item['mosaicRestrictionEntry']['restrictions'][0]['restriction']['restrictionType'] = 999
+		self._assert_model_rejects(item)
+
+	def test_key_shape_is_domain_values(self):
```
Jaguar0625: not sure purpose of this test?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4971,3 +4994,546 @@ def test_repair_rollback_restores_chain_and_lock_state_when_later_secret_lock_wr
...
+		# Arrange:
+		rows = [(
+			5, 'transaction-1', 999, None, RESTRICTION_MOSAIC_ID, 0)]
+
+		# Act / Assert:
+		with self.assertRaises(ValueError):
+			SymbolDatabase._mosaic_restriction_key_from_confirmed_transaction_rows(  # pylint: disable=protected-access
+				rows)
+
+	def test_confirmed_global_mosaic_restriction_query_rejects_missing_position_zero_mosaic(self):
```
Jaguar0625: i'm not sure what this test is verifying?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4971,3 +4994,546 @@ def test_repair_rollback_restores_chain_and_lock_state_when_later_secret_lock_wr
...
+			f'Invalid confirmed Symbol global Mosaic Restriction transaction at height 5, id {transaction_id}'):
+			database.get_confirmed_mosaic_restriction_keys_since(5)
+
+	def test_confirmed_address_mosaic_restriction_query_rejects_missing_target_address(self):
+		# Arrange:
+		database = self._create_database()
+		database.upsert_blocks([_create_block(5)])
+		database.upsert_transactions_for_height(5, [create_transaction_entry(
+			5,
+			key='missing-address-target',
```
Jaguar0625: shouldn't this be valid key?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4971,3 +4994,546 @@ def test_repair_rollback_restores_chain_and_lock_state_when_later_secret_lock_wr
...
+		cursor.execute('SELECT id FROM symbol_transactions WHERE height = %s', (5,))
+		transaction_id = cursor.fetchone()[0]
+
+		# Act / Assert:
+		with self.assertRaisesRegex(
+			ValueError,
+			f'Invalid confirmed Symbol address Mosaic Restriction target at height 5, id {transaction_id}'):
+			database.get_confirmed_mosaic_restriction_keys_since(5)
+
+	def test_confirmed_mosaic_restriction_query_groups_rows_for_one_transaction(self):
```
Jaguar0625: test name seems misleading since it seems to be verifying there can't be multiple rows associated with global restriction?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4971,3 +4994,546 @@ def test_repair_rollback_restores_chain_and_lock_state_when_later_secret_lock_wr
...
+
+	def test_rollback_refresh_entries_supports_zero_argument_immutable_defaults(self):
+		# Arrange / Act:
+		entries = RollbackRefreshEntries()
+
+		# Assert:
+		self.assertEqual(((), (), (), (), (), ()), entries)
+		self.assertEqual((), entries.mosaic_restriction_entries)
+		with self.assertRaises(AttributeError):
+			entries.mosaic_restriction_entries = ()
```
Jaguar0625: not sure if needed since namedtuple never has assignable fields

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -1055,12 +1126,46 @@ async def fetch_entry(search_key):
...
+
+			response = await self.get_symbol_node(path)
+			if not isinstance(response, dict) or not isinstance(response.get('pagination'), dict):
+				raise ValueError('Malformed Symbol Mosaic Restriction search response')
+			pagination = response['pagination']
+			# Empty data is a deletion only after the node proves this was the exact first page requested;
+			# accepting a wrong page would turn a malformed response into a destructive stale-row delete.
+			page_number = pagination.get('pageNumber')
+			page_size = pagination.get('pageSize')
+			if not _is_exact_json_integer(page_number) or page_number != 1 or not _is_exact_json_integer(page_size) or page_size != MAX_PAGE_SIZE:
```
Jaguar0625: why page_number != 1 ?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+			'restrictions': []
+		}
+	}
+
+
+class PullerMosaicRestrictionsTest(TestCase):  # pylint: disable=too-many-public-methods
+	@staticmethod
+	def _fetch_response(response, key=None):
+		# Arrange:
+		puller = object.__new__(SymbolPuller)
```
Jaguar0625: why object.__new__ ?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+	def test_collector_ignores_unrelated_transaction(self):
+		# Arrange:
+		transaction = _global_transaction()
+		transaction['type'] = TransactionType.TRANSFER.value
+
+		# Act:
+		keys = SymbolPuller._collect_dirty_mosaic_restriction_keys_for_batch({10: [transaction]})
+
+		# Assert:
+		self.assertEqual(set(), keys)
```
Jaguar0625: these tests seem to be the same (unrelated transaction)

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		transaction = _global_transaction(reference_mosaic_id='1111111111111111')
+
+		# Act:
+		keys = SymbolPuller._collect_dirty_mosaic_restriction_keys_for_batch({
+			11: [transaction]})
+
+		# Assert:
+		self.assertEqual({MosaicRestrictionKey(MosaicRestrictionEntryType.GLOBAL, MOSAIC_ID, None)}, keys)
+
+	def test_collector_uses_resolved_position_zero_mosaic_and_target_address(self):
```
Jaguar0625: what do you mean by "position zero"?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		async def get_symbol_node(path):
+			paths.append(path.lstrip('/'))
+			return {'pagination': {'pageNumber': 1, 'pageSize': 100}, 'data': []}
+
+		puller.get_symbol_node = get_symbol_node
+		target_address = bytes.fromhex(RECIPIENT_ADDRESS)
+		key = MosaicRestrictionKey(MosaicRestrictionEntryType.ADDRESS, MOSAIC_ID, target_address)
+
+		# Act:
+		asyncio.run(puller._fetch_dirty_mosaic_restrictions([key], 20))
```
Jaguar0625: do you want to check entries like previous test ?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		reference_mosaic_id='0000000000000000'
+	):
+		overrides = {'type': entry_type, 'mosaicId': mosaic_id}
+		if entry_type == TransactionType.MOSAIC_ADDRESS_RESTRICTION.value:
+			overrides['targetAddress'] = target_address
+		else:
+			overrides['referenceMosaicId'] = reference_mosaic_id
+		return create_node_transaction(height, transaction_hash='A' * 64, transaction_id='restriction-transaction', **overrides)
+
+	def _create_normal_connector(self, response=None, failure=None):
```
Jaguar0625: do you need both failure and response? are you ever passing two arguments?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+			"SELECT body->>'mosaicId', body->>'targetAddress', encode(target_address, 'hex') "
+			'FROM symbol_transactions WHERE height = 2')
+		self.assertEqual([(
+			ALIAS_MOSAIC_ID, ALIAS_ADDRESS, RECIPIENT_ADDRESS.lower())], cursor.fetchall())
+		cursor.execute(
+			'SELECT mosaic_id, position FROM symbol_transaction_mosaics '
+			'WHERE height = 2 AND role = %s',
+			('restriction',))
+		self.assertEqual([(MOSAIC_ID, 0)], cursor.fetchall())
+		self.assertEqual(1, normal_connector.paths.count(path))
```
Jaguar0625: you seem to be testing alias resolution on sync only (here), not rollback; unless i misunderstand the test?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+			'SELECT COUNT(*) FROM symbol_transaction_mosaics '
+			'WHERE height = 2 AND role = %s',
+			('restriction',))
+		self.assertEqual(0, cursor.fetchone()[0])
+		sync_state = self.puller.symbol_db.get_sync_state()
+		self.assertEqual('healthy', sync_state['status'])
+		self.assertEqual(3, sync_state['last_synced_height'])
+		self.assertEqual(bytes.fromhex(f'{3:064X}'), bytes(sync_state['last_synced_block_hash']))
+
+	def test_sync_block_headers_uses_only_global_position_zero_when_both_mosaic_aliases_resolve(self):
```
Jaguar0625: how is test verifying that only position zero is used? db contains both

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		self.assertEqual([(MOSAIC_ID, 0)], cursor.fetchall())
+		self.assertEqual(1, connector.paths.count(path))
+
+	def test_sync_block_headers_rejects_missing_mosaic_resolution_before_restriction_fetch(self):
+		# Arrange:
+		transaction = self._restriction_transaction(
+			entry_type=TransactionType.MOSAIC_GLOBAL_RESTRICTION.value, mosaic_id=ALIAS_MOSAIC_ID)
+		connector = self._create_alias_connector(transaction, '', None)
+
+		self._assert_sync_failure_has_no_write_state(
```
Jaguar0625, line 736: you are missing `# Act + Assert:`

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+			connector, ValueError, f'Missing Symbol mosaic resolution.*{ALIAS_MOSAIC_ID}')
+
+	def test_sync_block_headers_rejects_inapplicable_mosaic_resolution_before_restriction_fetch(self):
```
Jaguar0625: inapplicable because resolution is at source > tx source ?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		self._sync_with_connector(connector)
+
+		# Assert:
+		cursor = self.puller.symbol_db.connection.cursor()
+		cursor.execute('SELECT entry_type, target_address FROM symbol_mosaic_restrictions')
+		self.assertEqual([('address', bytes.fromhex('98' + '11' * 23))], [
+			(entry_type, bytes(target_address)) for entry_type, target_address in cursor.fetchall()
+		])
+
+	def _seed_backend6_and_lock_state(self):
```
Jaguar0625: backend6 ?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
```diff
@@ -0,0 +1,983 @@
...
+		self.assertEqual(before_state, self.puller.symbol_db.get_sync_state())
+		self.assertEqual([1, 2, 3], self._fetch_block_heights(self.puller.symbol_db))
+		self.assertEqual(before_chain_state, self._fetch_table_state((
+			'symbol_blocks',
+			'symbol_transactions',
+			'symbol_transaction_mosaics',
+			'symbol_transaction_addresses')))
+		self.assertEqual(before_restriction_state, self._fetch_restriction_state())
+		self.assertEqual(before_lock_state, self._fetch_lock_state())
+		self.assertEqual(before_backend6_state, self._fetch_backend6_state())
```
Jaguar0625: isn't this slightly more general version of the tests that track self._fetch_complete_batch_state()?
why not just add locks and restrictions to the base and you should get more coverage?

Jaguar0625 (commented):
👍

explorer/puller/tests/facade/test_async_utils.py
daoka: Added a 23-item test with a chunk size of 10, which exercises three chunks. It verifies the original non-sorted result order and asserts that the maximum number of concurrent requests is exactly 10.

explorer/puller/puller/model/symbol/format.py
daoka: Refactored the duplicated Address and Hash decoding into `_decoded_bytes_from_hex`, with the SDK-specific decoder passed as a callable. The public wrappers and their validation behavior remain unchanged.

explorer/puller/puller/model/symbol/MosaicRestriction.py
daoka: Agreed. Renamed the node integer conversion maps and related symbols from `NODE` to `ENUM` throughout the implementation and tests. The naming now distinguishes the enum integer value from the database label.

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
daoka: Updated the alias mosaic ID rejection test to use `_assert_key_rejects`, matching the surrounding key validation tests.

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
daoka: Updated the applicable key validation test names to include either `global` or `address`. The invalid-entry-type test remains generic because it does not arrange a valid restriction entry type.

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
daoka: Combined each True/False pair into a single test using `subTest`. Applied the same change to all four boolean rejection pairs in this test module.

explorer/puller/puller/model/symbol/MosaicRestriction.py
daoka: Added the generic `is_exact_integer` predicate to `format.py` and applied it to all eight matching validation sites in Mosaic Restriction, Lock, `async_utils`, and `SymbolPuller`. Added focused coverage for the helper as well.

explorer/puller/puller/model/symbol/MosaicRestriction.py
daoka: Good point about avoiding the literal. I introduced `UINT64_MAX_DECIMAL` for generic uint64 validation and now define `MOSAIC_ADDRESS_RESTRICTION_REMOVAL_SENTINEL` from it. This avoids coupling generic uint64 validation to Address restriction removal semantics.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Renamed the test and added an Arrange comment explaining the invalid state. A persisted restriction transaction is used as a rollback key source and must have exactly one primary restriction mosaic relation, so a missing relation represents corrupt persisted data and is rejected.

explorer/puller/tests/model/symbol/test_MosaicRestriction.py
daoka: Removed this test. It only verified standard namedtuple behavior, while the domain constraints are already covered by the Mosaic Restriction key creation tests.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: You're right that the custom fixture key is not what makes the transaction invalid. Removed the explicit keys from both the missing-address-target and missing-global-mosaic fixtures. The tests now focus on the missing required relation or target.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Renamed the test to state the actual invalid condition: a Global restriction transaction with multiple primary restriction mosaic relations. This matches the implementation check that requires exactly one primary relation.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Removed the assignment-based immutability check and renamed the test to focus on zero-argument defaults. The assertions for the six-field schema, field order, and empty defaults remain because positional construction is used by existing callers.

explorer/puller/puller/facade/SymbolPuller.py
daoka: The validation intentionally remains. Requesting page one does not guarantee that the node returns matching pagination metadata. Treating a mismatched empty response as a valid result could incorrectly delete current restriction state. I strengthened the existing comment to make this destructive-failure risk explicit.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Removed all `object.__new__(SymbolPuller)` usage from these tests. The exact-fetch tests now use the initialized `self.puller` provided by `SymbolPullerTestBase`, with the connector injected through the existing `set_symbol_connector` pattern.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Agreed. Combined the Account Restriction and Transfer cases into one `test_collector_ignores_non_mosaic_restriction_transactions` test using `subTest`.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Position 0 is the transaction's primary restriction mosaic relation. For a Global restriction, position 1 is the optional reference mosaic relation. Renamed the affected tests and added comments using these domain terms.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Added a full replacement-entry assertion to the Address exact-fetch test, including the empty row set. The request history is also asserted as the exact expected sequence. Global and Address expected paths are now defined once as independent constants.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: No, the arguments were mutually exclusive. Collapsed them into a single `response` argument. The connector already raises the value when it is an Exception, so both success and failure cases use the same path.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: The test covers both phases. The initial sync persists the resolved child relations on the branch that later becomes orphaned, and rollback then uses those persisted mosaic and address relations to build the restriction request. I renamed the test, clarified the phase-specific Arrange comments, and kept an exact rollback request-path assertion.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Changed the request assertion to compare the complete filtered restriction-path sequence with `[path]`. This proves that only the primary restriction mosaic generated a request and that no request was made for the position-one reference mosaic. Applied the same complete-history pattern to the related sync and rollback tests.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Added `# Act + Assert:` to all four related missing/after-source resolution rejection tests for both mosaic and address resolutions.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Yes. The transaction source has `primaryId = 1`, while the resolution entry has `primaryId = 2`, so that entry cannot apply to the transaction. Renamed both mosaic and address tests to `resolution_entry_after_transaction_source` and documented the IDs in the Arrange comments.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Removed the `backend6` terminology. The remaining seed helper is now named `_seed_existing_current_state`, while the state-fetch helper was removed as part of the shared complete-state snapshot consolidation.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Agreed. Moved `_fetch_table_state` and `_fetch_complete_batch_state` to `SymbolPullerTestBase` and expanded the complete snapshot to all 16 relevant tables, including `symbol_sync_state`, both lock tables, and `symbol_mosaic_restrictions`. Removed the duplicate helpers and updated the lock and restriction rollback tests to use the shared snapshot.

explorer/puller/puller/db/SymbolDatabase.py
daoka: Thanks for the suggestion. I kept the per-row execution for now. The production puller path performs an exact-key lookup and rejects responses with more than one entry, so it currently passes zero or one row. This is a property of that caller, not an invariant enforced by the DB method.

explorer/puller/puller/db/SymbolDatabase.py
daoka: The condition limits the delete to the one logical key being replaced; the rows below reinsert only that key, not every sibling. IS NOT DISTINCT FROM is required because a Global restriction has target_address = NULL, which would never match with =. Existing DB tests cover different-target siblings, different-entry-type siblings, and Global NULL replacement.

explorer/puller/tests/db/test_SymbolDatabase.py
Jaguar0625: so, the condition making test invalid is `mosaic_rows=[]`, correct ?

explorer/puller/puller/facade/SymbolPuller.py
Jaguar0625: > Treating a mismatched empty response as a valid result could incorrectly delete current restriction state

ok, but how is this different from node returning any other invalid data? isn't assumption node is owned/trusted by explorer operator?

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
Jaguar0625: i like that you renamed position 0 => primary restriction mosaic relation. is much clearer now 👍

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
Jaguar0625: > The test covers both phases

 can you create two tests, one for each phase? 
 
 > rollback then uses those persisted mosaic and address relations to build the restriction request
 
unless i am misreading, the aliases appear to be in block 2, so after the rollback i don't think there are any aliases still in the chain?

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: You're right that this validation is stricter than the existing paginated fetches, including Secret Locks.

This is not because pagination metadata is considered less trustworthy, or because the operator-owned node is treated as hostile. The reason is narrower: a successful empty result from this exact-key fetch is translated into deleting the current local restriction. Before performing that destructive action, this path fails fast if the returned `pageNumber` and `pageSize` do not match the requested page 1 and page size 100.

This check only detects an observable mismatch between the request and the returned metadata. It cannot detect a response that looks structurally valid but contains semantically incorrect data. When an entry is returned, the implementation separately validates its cardinality and logical key.

explorer/puller/tests/facade/symbol/test_PullerMosaicRestrictions.py
daoka: Done — I split this into separate sync and rollback tests.  
  
The sync test verifies that the unresolved aliases remain in the transaction body while the resolved target address and primary restriction mosaic relation are persisted in the transaction child rows.  
  
The rollback test runs the initial sync only as its Arrange. That initial sync receives a successful empty restriction response, so no current-state restriction row can provide the rollback key. The persisted resolved target address and position-0 mosaic relation from the orphaned transaction are therefore the only key source. `_repair_from_height()` reads them, builds the concrete restriction key, and fetches canonical state before applying the destructive rollback database transaction.  
  
After rollback, height 2 contains the canonical Transfer transaction with no `mosaicId`; the orphaned alias-bearing restriction transaction and its restriction child relation no longer remain.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes — mosaic_rows=[] is the invalidating condition in this test. The persisted transaction then has no primary restriction mosaic relation, so position_zero_rows is empty and the restriction key cannot be reconstructed. target_address=None is valid for a Global restriction and is normalized to None.

I also renamed the test to test_get_confirmed_mosaic_restriction_keys_since_rejects_persisted_global_transaction_without_primary_mosaic_relation so the invalid state is explicit.

Jaguar0625 (approved):
👍

PR #2481, opened 2026-08-05
[explorer/puller] feat: Capture rollback data

AnthonyLaw:
Problem: Fork-height detection was previously implemented, but the data required to safely repair an orphaned chain was not captured before rollback.

Solution:
- Capture orphan block and transaction data above the confirmed fork height before deleting anything.
- Include block beneficiaries and signers, direct and inner transaction accounts, remote-link accounts, fee and levy recipients.
- Extract affected namespace roots and mosaic identifiers from transaction payloads for later projection reconstruction.
- Read account creation heights to separate orphan-created accounts from surviving affected accounts.
- Preserve surviving affected accounts for later node-state refresh and remove orphan-created accounts during repair.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -20,7 +20,7 @@ NEM_MAX_ROLLBACK_DEPTH = 360
 
 
 class NemRollbackError(RuntimeError):
-	"""Raised when a safe NEM rollback point cannot be found."""
+	"""Raised when a NEM rollback cannot safely proceed."""
 
 
 BlockRecord = namedtuple('BlockRecord', [
@@ -92,6 +92,25 @@ TransactionRecord = namedtuple('TransactionRecord', [
 	'payload',
 	'size',
 	'version'
+])
+RollbackPayloadAccounts = namedtuple('RollbackPayloadAccounts', [
+	'accounts',
+	'remote_link_accounts'
+])
+RollbackProjectionIdentifiers = namedtuple('RollbackProjectionIdentifiers', [
+	'namespace_roots',
+	'mosaic_names'
+])
+NemRollbackImpact = namedtuple('NemRollbackImpact', [
+	'fork_height',
+	'affected_accounts',
+	'account_creation_heights',
+	'orphan_created_accounts',
+	'surviving_affected_accounts',
+	'orphan_beneficiaries',
+	'affected_remote_link_accounts',
+	'affected_namespace_roots',
+	'affected_mosaic_names'
 ])
 DatabaseConfig = namedtuple('DatabaseConfig', ['database', 'user', 'password', 'host', 'port'])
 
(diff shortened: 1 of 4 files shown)


explorer/puller/tests/db/test_NemDatabase.py
```diff
@@ -1141,3 +1141,83 @@ def test_insert_transaction_mosaic(self):
...
+			orphan_blocks, orphan_transactions, orphan_transfer_levy_recipients = nem_database.get_orphan_chain_records(1)
+			account_heights = nem_database.get_account_creation_heights({
+				surviving_account.address,
+				orphan_account.address
+			})
+
+		# Assert:
+		self.assertEqual(1, len(orphan_blocks))
+		self.assertEqual(str(BLOCKS[1].beneficiary), str(orphan_blocks[0].beneficiary))
+		self.assertEqual(str(BLOCKS[1].signer), str(orphan_blocks[0].signer))
```
Jaguar0625: you should be able to compare Address directly

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -174,6 +193,124 @@ async def detect_rollback(self, db_height, chain_height):
...
+				self._convert_public_key_to_address(transaction.payload['remote_account'])
+			)
+		elif TransactionType.MULTISIG_ACCOUNT_MODIFICATION.value == transaction.transaction_type:
+			for modification in transaction.payload['modifications']:
+				affected_accounts.add(
+					self._convert_public_key_to_address(modification['cosignatory_account'])
+				)
+		elif TransactionType.MULTISIG.value == transaction.transaction_type:
+			for signature in transaction.payload['signatures']:
+				affected_accounts.add(Address(signature['other_account']))
```
Jaguar0625: won't this always be the same (multisig) account ?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -174,6 +193,124 @@ async def detect_rollback(self, db_height, chain_height):
...
+			parent = transaction.payload['parent']
+			if parent:
+				affected_namespace_roots.add(parent.split('.')[0])
+			else:
+				affected_namespace_roots.add(transaction.payload['namespace'])
+		elif transaction.transaction_type in (
+			TransactionType.MOSAIC_DEFINITION.value,
+			TransactionType.MOSAIC_SUPPLY_CHANGE.value
+		):
+			affected_mosaic_names.add(transaction.payload['namespace_name'])
```
Jaguar0625, line 243: does 'namespace_name' include the mosaic name?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -174,6 +193,124 @@ async def detect_rollback(self, db_height, chain_height):
...
+		orphan_beneficiaries = set()
+		affected_remote_link_accounts = set()
+		affected_namespace_roots = set()
+		affected_mosaic_names = set()
+
+		orphan_records = self.nem_db.get_orphan_chain_records(fork_height)
+
+		for block in orphan_records.blocks:
+			orphan_beneficiaries.add(block.beneficiary)
+			affected_accounts.add(block.beneficiary)
```
Jaguar0625, line 275: isn't the block beneficiary always the signer in NEM or are you using it to denote the main account (vs remote account)?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -174,6 +193,124 @@ async def detect_rollback(self, db_height, chain_height):
...
+		orphan_created_accounts = {
+			address
+			for address, height in account_creation_heights.items()
+			if height > fork_height
+		}
+		surviving_affected_accounts = {
+			address
+			for address, height in account_creation_heights.items()
+			if height <= fork_height
+		}
```
Jaguar0625: you can optimize this into a single loop over account_creation_heights

explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -1610,3 +1612,290 @@ def test_no_matching_block_hash_within_360_blocks_requires_manual_investigation(
...
+		# Assert:
+		self.assertEqual(fork_height, capture.fork_height)
+		self.assertEqual(expected_accounts, capture.affected_accounts)
+		self.assertEqual({recipient}, capture.orphan_created_accounts)
+		self.assertEqual(expected_accounts - {recipient}, capture.surviving_affected_accounts)
+		self.assertEqual({beneficiary}, capture.orphan_beneficiaries)
+		self.assertEqual({sender}, capture.affected_remote_link_accounts)
+		self.assertEqual({'root'}, capture.affected_namespace_roots)
+		self.assertEqual({'root.token', 'root.supply'}, capture.affected_mosaic_names)
+		self.assertEqual(expected_accounts, set(capture.account_creation_heights))
```
Jaguar0625: do you want to validate values too ?

explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -1610,3 +1612,290 @@ def test_no_matching_block_hash_within_360_blocks_requires_manual_investigation(
...
+			transaction_type,
+			is_inner,
+			sender_address,
+			recipient_address,
+			payload,
+			100,
+			1
+		)
+
+	def test_can_capture_all_data_affected_by_rollback(self):  # pylint: disable=too-many-locals
```
Jaguar0625, line 1687: ok for now, but in future, it's better to have multiple unit tests - e.g. each testing each tx type - vs single integration test like this

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: Yes. namespace_name includes the mosaic name, for example root.token, and matches the existing mosaics.namespace_name field.

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: You are right.
`other_account` can be extracted once from the first signature, and the loop only collects each signature sender.

explorer/puller/puller/facade/NemPuller.py
AnthonyLaw: They differ when an account has an active remote link for delegated harvesting. The beneficiary is the main account that receives the harvested fees. The address derived from the signer is either the main account for local harvesting or the linked remote account for delegated harvesting.

explorer/puller/tests/db/test_NemDatabase.py
AnthonyLaw: Makes sense 👍🏼 

Previously, the database value was wrapped as Address(memoryview), while the expected value was Address(bytes), but after normalizing the database buffer with bytes() before constructing Address. It should compare Address directly instead of str

Jaguar0625 (approved):
👍

PR #2482, opened 2026-08-06
[explorer/frontend] fix: make header logo clickable

cryptoBeliever:
Problem:
The logo in the top-left corner of the header was a plain <div> with no navigation behavior, so clicking it did nothing on every page.

Solution:
Wrap the logo in a Next.js Link pointing to the home route, matching the existing menu items which already use createPageHref('home').


Jaguar0625 (approved):
👍

OlegMakarenko (approved):

AnthonyLaw (approved):
👍🏼

PR #2483, opened 2026-08-06
[explorer/frontend] fix: use runtime configuration on error pages

cryptoBeliever:
Fixes #2469

## Problem

Using the header search bar on a 404 page sends requests to a malformed URL:

```
❌ https://nem.fyi/undefined/block/B411ECEB...
✅ https://nem.fyi/api/nem/block/B411ECEB...
```

The requests fail and the search bar silently returns no results.

## Root cause

Next.js requires `pages/404` and `pages/500` to be statically prerendered - it explicitly rejects `getServerSideProps` and `getInitialProps` on them. Their HTML, including the `window.appConfig` script that `_app` inlines, is therefore generated once, during `next build`.

`PUBLIC_*` variables are runtime configuration: the `Dockerfile` only passes `NEXT_PUBLIC_EXPLORER_VARIANT` as a build arg. At build time `process.env.PUBLIC_API_BASE_URL` is `undefined`, and `JSON.stringify` drops keys with `undefined` values entirely - so the key is simply missing from `window.appConfig` on those pages.

`createApiUrl` then interpolates it into a template literal:

```js
  `${undefined}/block/HASH`  // → "undefined/block/HASH"
```

Every other page is server-rendered per request, which is why they are unaffected.

## Solution

Serve the configuration from a route that is never statically optimized, and load it from a script tag that runs before the application does.

The inline script is deliberately retained: if `/runtime-config.js` ever fails to load, server-rendered pages keep working with their freshly inlined values instead of losing configuration entirely.

Because the fix works below the level of any individual request, it covers the search bar, the backend health check, and any request added in the future - no per-caller handling required.

### DIFF
--- a/explorer/frontend/config/index.js
+++ b/explorer/frontend/config/index.js
@@ -35,4 +35,4 @@ export const publicAppConfig = Object.fromEntries(Object.entries(serverAppConfig
 
 const isClientSide = typeof window !== 'undefined';
 
-export default isClientSide ? (window.appConfig ?? publicAppConfig) : serverAppConfig;
+export default isClientSide ? window.appConfig : serverAppConfig;
--- a/explorer/frontend/next.config.js
+++ b/explorer/frontend/next.config.js
@@ -20,6 +20,9 @@ const variantAlias = `./variants/${VARIANT}`;
 module.exports = {
 	output: 'standalone',
 	reactStrictMode: true,
+	// `/api/*` on a deployed origin is proxied to the REST backend, so the runtime config script is
+	// exposed outside that namespace.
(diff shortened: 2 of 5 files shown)


Jaguar0625 (approved):
👍

OlegMakarenko (commented):
Good job! Can you please remove this fallback as well: https://github.com/symbol/product/blob/explorer/fix-runtime-config-on-error-pages/explorer/frontend/config/index.js#L38

Jaguar0625 (approved):
👍

OlegMakarenko (approved):

PR #2484, opened 2026-08-06
[explorer/nodewatch]: minor fixes 

Jaguar0625:
[explorer/nodewatch]: suppress zenlog.warn deprecation warnings
    
     problem: warnings are emitted during pytest
    solution: supress in pytest.ini

    [explorer/nodewatch]: replace None.__ne__ with lambda expression
    
     problem: None.__ne__ does not behave as expected in some python versions
              (see also https://github.com/python/cpython/issues/112125)
    solution: replace None.__ne__ with lambda expression
              (operator.is_not_none would work but is only available in python 3.14+)

    [explorer/nodewatch]: update plotly library
    
     problem: using old version of plotly from 2021
    solution: update to latest


AnthonyLaw (approved):
👍🏼

PR #2486, opened 2026-08-07
[explorer/puller] fix: skip account refresh for mosaic restriction ta…

daoka:

The Symbol puller collects every address relation from confirmed transactions as a dirty account and requests its current state from the Node `/accounts` endpoint.

This includes the `target` address relation of a Mosaic Address Restriction transaction. The account refresh path requires the Node response to contain one account item for every requested address.

## What's the issue?

A Mosaic Address Restriction target can have restriction state without having corresponding account state.

When the Node omits such an address from the `/accounts` batch response, the puller raises `Missing Symbol accounts batch item` and fails the block sync even though the Mosaic Address Restriction transaction itself is valid.

The dirty-account collector now excludes an address relation only when both of the following apply:

- the transaction type is `MOSAIC_ADDRESS_RESTRICTION`;
- the address relation role is `target`.

The transaction address relation is still persisted for transaction search and Mosaic Restriction state synchronization.

The address remains eligible for account refresh when it is independently observed as another account-bearing role, including:

- a block beneficiary;
- another transaction participant;
- a balance-change or balance-transfer receipt participant.

Non-Mosaic-Restriction transaction targets retain their existing account-refresh behavior.

### DIFF
--- a/explorer/puller/tests/facade/symbol/test_PullerAccounts.py
+++ b/explorer/puller/tests/facade/symbol/test_PullerAccounts.py
@@ -12,6 +12,7 @@ from puller.db.SymbolDatabase import RollbackRefreshEntries, SymbolDatabase
 from puller.facade.SymbolPuller import ACCOUNT_BATCH_FETCH_SIZE
 from puller.model.symbol.Account import create_account_row
 from puller.model.symbol.Block import create_block_row
+from tests.test.SymbolTestConstants import SIGNER_ADDRESS
 
 from .puller_test_utils import (
 	BENEFICIARY_ADDRESS,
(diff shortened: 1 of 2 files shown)


explorer/puller/tests/facade/symbol/test_PullerAccounts.py
```diff
@@ -290,6 +315,169 @@ def test_refresh_dirty_accounts_for_batch_deduplicates_repeated_transaction_part
...
+
+		# Act:
+		self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual([{
+			'addresses': [self._address_text(RECIPIENT_ADDRESS), self._address_text(BENEFICIARY_ADDRESS)]
+		}], connector.post_payloads)
+
+	def test_refresh_dirty_accounts_for_batch_keeps_mosaic_restriction_target_when_it_is_beneficiary(self):
```
Jaguar0625: how is this test different from previous?

only difference is targetAddress, but in both cases isn't RECIPIENT_ADDRESS picked up from `beneficiaryAddress=RECIPIENT_ADDRESS` ?

Jaguar0625 (commented):
👍

explorer/puller/tests/facade/symbol/test_PullerAccounts.py
daoka: Yes, `RECIPIENT_ADDRESS` was collected from the block beneficiary in both tests.

The intended distinction is which independent source shares the restriction target address:

- signer case: the target address is also the transaction signer;
- beneficiary case: the target address is also the block beneficiary.

I updated the fixtures to make this explicit. The signer case now uses `SIGNER_ADDRESS` as the target and a distinct block beneficiary, while the beneficiary case uses `RECIPIENT_ADDRESS` as both the target and block beneficiary. Their complete account request payloads now differ, and the Arrange comments explain why each target address remains dirty.

Jaguar0625 (approved):
👍

PR #2492, opened 2026-08-10
[explorer/rest] fix: cache statistics endpoints

cryptoBeliever:
## Problem

The three statistics endpoints recompute a summary of the entire chain on every request, and two
of them take no parameters at all - so every visitor pays for the same result.

Measured against a copy of production (11.5M transactions):

  | endpoint | buffer pages | median |
  |---|---|---|
  | `transaction/statistics` | 175 332 (~1.4 GB) | 843.5 ms |
  | `account/statistics` | 29 192 | 258.7 ms |
  | `transaction/statistics/range` (48 months) | - | 304.2 ms |
  | `transaction/statistics/range` (90 days) | - | 17.5 ms |

  On nem.fyi `transaction/statistics` takes **2.36 s**, because production reads those pages from
  disk rather than from a warm page cache. Concurrent callers make it worse: they queue on the same
  buffers (`IPC/BufferIo`) instead of sharing a result.

  The frontend calls these endpoints to render the home page, so this cost is on the first screen
  every visitor sees.

  ## Why caching is the right fix here

  These are the rare queries where a cache is not papering over a bad plan. The result is a handful
  of aggregates over the whole table, so there is no page to narrow down to and no index that makes
  a full aggregation cheap. The cost is inherent to the question.

  A 15 minute lifetime means a statistic can lag by at most one quarter hour - invisible against a
  figure that moves a couple of dozen times a day - and removes essentially all of the work.

### DIFF
--- a/explorer/rest/rest/facade/NemRestFacade.py
+++ b/explorer/rest/rest/facade/NemRestFacade.py
@@ -6,12 +6,7 @@ from symbollightapi.model.Exceptions import NodeException
 
 from rest.db.NemDatabase import NemDatabase
 from rest.facade.TtlCache import TtlCache
-from rest.model.common import (
-	STATISTICS_CACHE_MAX_ENTRIES,
-	STATISTICS_CACHE_TTL_SECONDS,
-	STATISTICS_RANGE_CACHE_MAX_ENTRIES,
-	Pagination
-)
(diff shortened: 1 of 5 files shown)


Jaguar0625 (commented):
👍

Jaguar0625 (commented):
👍

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2495, opened 2026-08-11
[explorer] fix: resolve account transactions through an aggregate hash column

cryptoBeliever:
## The problem

`/api/nem/transactions?address=…` times out on nem.fyi. Measured against a copy of production
(11.5M transactions), a single page of ten rows costs:

| account | buffer pages | time |
|---|---|---|
| dormant, 2 transactions | 10 262 395 | 80 s cold, 7–10 s warm |
| multisig, 339k transactions | 124 778 | 1.5 s |
| heaviest, 2.25M transactions | 2 031 867 | 2.3 s |

The dormant account being the worst looks backwards until you see the mechanism. `LIMIT 10` can
only stop a scan once ten rows have been found. An account with two transactions never finds them,
so the whole table has to be read to prove no more exist. **86.7% of accounts hold fewer than ten
transactions** (834 649 of 962 477), which means most of the address space falls into the full scan
rather than a rare corner of it.

Sample address: `NAAANAABDBG526XNAVTXD5I4IF3N6EIVOHIB3U7E`.

### The query before

```sql
SELECT
    t.transaction_hash, t.transaction_type, t.sender_address AS from_address, t.fee, t.height,
    t.timestamp, t.deadline, t.signature, t.recipient_address AS to_address, t.payload,
    COALESCE(m.mosaics, '[]'::json) AS mosaics, t.size, t.version
FROM transactions t
LEFT JOIN LATERAL (
    SELECT json_agg(json_build_object(
        'namespace_name', tm.namespace_name,
        'quantity',       tm.quantity,
        'divisibility',   mo.divisibility)) AS mosaics
    FROM transactions_mosaic tm
    LEFT JOIN mosaics mo ON mo.namespace_name = tm.namespace_name
    WHERE tm.transaction_id = t.id
) m ON true
WHERE t.is_inner = False
  AND ( (t.transaction_type != 4100 AND t.sender_address = $1)
     OR t.recipient_address = $1
     OR ( t.transaction_type = 4100 AND EXISTS (
            SELECT 1 FROM transactions it
             WHERE it.is_inner = true
               AND it.transaction_hash = decode(t.payload->>'inner_hash', 'hex')
               AND (it.sender_address = $1 OR it.recipient_address = $1) ) ) )
ORDER BY t.height DESC
LIMIT 10 OFFSET 0
```

### Why it was slow

Two independent reasons, and removing only one of them does not help.

**The correlated subquery.** The third arm reaches multisig inner transactions through
`decode(t.payload->>'inner_hash','hex')`. Being correlated, it is evaluated per candidate row. That
rules out any bitmap plan and leaves a walk of `transactions_height_is_inner_false_idx` with a
filter attached:

```
Index Scan using transactions_height_is_inner_false_idx on transactions t
  Rows Removed by Filter: 11 062 748
```

**The disjunction.** Even with the correlated arm gone, `sender_address = X OR recipient_address =
X` remains. A btree is sorted along one axis, so no single index can return that combined set in
height order. PostgreSQL either uses `BitmapOr`, which yields rows in heap order and destroys the
ordering, or abandons the indexes entirely. Either way every match must be sorted before `LIMIT`
can take ten.

Measured on identical data and indexes, changing only `OR` into `UNION ALL`:

| form | pages | plan | sort |
|---|---|---|---|
| `OR` | 1 144 026 | `Parallel Seq Scan`, 2 247 738 rows | `external merge` 19 MB × 3 workers |
| `UNION ALL` | 25 | 2 × `Index Only Scan` | `quicksort` 25 kB in memory |

### `aggregate_hash` on inner rows

Inner rows now carry `aggregate_hash`, pointing at the transaction that owns them, mirroring
`symbol_transactions.aggregate_hash` (`SymbolDatabase.py:409`).

The direction is what matters. NEM's existing link lives in the wrapper's payload and points
*down*; this one points *up*, from child to parent. That lets the column ride along in the address
index as an `INCLUDE` payload, so a branch never visits the heap to learn the owner, and the join
runs against the existing `transaction_hash` unique index.

The puller sets it when inserting the inner row, where the wrapper's hash is already in scope
(`NemPuller.py:582`).

### Indexes

```sql
CREATE INDEX transactions_sender_address_height_idx
    ON transactions (sender_address, height DESC, id DESC)
    INCLUDE (aggregate_hash)
    WHERE NOT (is_inner = false AND transaction_type = 4100);
```

and the same for `recipient_address`.

The partial predicate is not only about size. It proves to the planner that the condition already
holds for every entry, so it disappears from the plan. Without it the condition returns as a
runtime `Filter` needing `is_inner` and `transaction_type`, which are not in the index, so every
candidate visits the heap and the scan stops being index-only.

### The query after

```sql
SELECT
    t.transaction_hash, t.transaction_type, t.sender_address AS from_address, t.fee, t.height,
    t.timestamp, t.deadline, t.signature, t.recipient_address AS to_address, t.payload,
    COALESCE(m.mosaics, '[]'::json) AS mosaics, t.size, t.version
FROM transactions t
LEFT JOIN LATERAL (
    SELECT json_agg(json_build_object(
        'namespace_name', tm.namespace_name,
        'quantity',       tm.quantity,
        'divisibility',   mo.divisibility)) AS mosaics
    FROM transactions_mosaic tm
    LEFT JOIN mosaics mo ON mo.namespace_name = tm.namespace_name
    WHERE tm.transaction_id = t.id
) m ON true
WHERE t.is_inner = false AND t.id IN (
    SELECT id FROM (
        SELECT DISTINCT id, height FROM (
            (SELECT COALESCE(o.id, t.id) AS id, t.height
               FROM transactions t
               LEFT JOIN transactions o ON o.transaction_hash = t.aggregate_hash
              WHERE t.sender_address = $1
                AND t.transaction_type <> 4100
              ORDER BY t.height DESC, t.id DESC
              LIMIT 10)
          UNION ALL
            (SELECT COALESCE(o.id, t.id) AS id, t.height
               FROM transactions t
               LEFT JOIN transactions o ON o.transaction_hash = t.aggregate_hash
              WHERE t.recipient_address = $1
                AND t.transaction_type <> 4100
              ORDER BY t.height DESC, t.id DESC
              LIMIT 10)
        ) page
        ORDER BY height DESC, id DESC
        LIMIT 10 OFFSET 0
    ) page_ids
)
ORDER BY t.height DESC, t.id DESC
```

Reading it in order:

- Each branch matches the address **directly on inner rows** and resolves the owner by joining up,
  so the correlated subquery is gone.
- Each branch walks its own index in the order it already stores, and stops after `limit + offset`
  entries. Truncating is safe: the ten newest rows of the union must each be among the ten newest
  of the branch they came from, so nothing that belongs on the page can be discarded.
- The `LATERAL` runs only for the ten rows that survived, not for every match.
- `DISTINCT` is required because a self-transfer matches both branches.

`address` uses both branches, `senderAddress` the first, `recipientAddress` the second. A sender
and recipient pair narrows a **single** branch instead of producing two, because that combination
is a conjunction — unioning it would return everything sent by one plus everything received by the
other. This is not an edge case: the frontend rewrites `address` into the opposite role whenever
From or To is used on an account page (`transactions.js:29-35`).

Correctness rests on inner and outer rows sharing a height, so ordering by the inner row's height
is sound. Verified: 509 834 / 509 834 pairs.

### Results

| account | page | before | after |
|---|---|---|---|
| dormant, 2 tx | 10/0 | 10 262 395 | **35** |
| dormant, 2 tx | 50/500 | 10 262 395 | **13** |
| multisig, 339k | 10/0 | 124 778 | **39** |
| multisig, 339k | 50/500 | 140 290 | **183** |
| heaviest, 2.25M | 10/0 | 2 031 867 | **138** |
| heaviest, 2.25M | 50/500 | 2 036 316 | **1 652** |

Cost no longer tracks account activity. Through the REST code against the full database every
address variant answers in 1.6–9.9 ms.

Development timings understate the gain: the table fits in this machine's page cache, so the
"before" figures are served from RAM. Production reads the same pages from disk, which is why the
first measurement above took 80 s cold and 7 s once warm — the page counts are the honest metric.

## Deployment

The column and its backfill must be in place **before** this code runs:

```sql
ALTER TABLE transactions ADD COLUMN IF NOT EXISTS aggregate_hash bytea;

UPDATE transactions i
   SET aggregate_hash = o.transaction_hash
  FROM transactions o
 WHERE o.is_inner = false
   AND o.transaction_type = 4100
   AND decode(o.payload->>'inner_hash', 'hex') = i.transaction_hash
   AND i.is_inner
   AND i.aggregate_hash IS NULL;

CREATE INDEX CONCURRENTLY IF NOT EXISTS transactions_sender_address_height_idx
    ON transactions (sender_address, height DESC, id DESC)
    INCLUDE (aggregate_hash)
    WHERE transaction_type <> 4100;

CREATE INDEX CONCURRENTLY IF NOT EXISTS transactions_recipient_address_height_idx
    ON transactions (recipient_address, height DESC, id DESC)
    INCLUDE (aggregate_hash)
    WHERE transaction_type <> 4100;

```

Then both indexes `CONCURRENTLY`, then `VACUUM (ANALYZE) transactions` — the backfill rewrites
509 834 rows, and without refreshing the visibility map the branches make heap fetches they do not
need. Allow about 1.7 GB of permanent growth plus headroom during the build.

Order: SQL, then the puller, then **rerun the backfill** for anything that arrived in between (the
statement is idempotent), then this gate, then the REST service.

```sql
SELECT count(*) FILTER (WHERE is_inner) AS inner_rows,
       count(*) FILTER (WHERE is_inner AND aggregate_hash IS NOT NULL) AS linked
FROM transactions;
```

Both numbers must match.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -139,7 +139,7 @@ class NemDatabase(DatabaseConnection):
 			CREATE INDEX IF NOT EXISTS transactions_sender_address_height_idx
 				ON transactions(sender_address, height DESC, id DESC)
 				INCLUDE (aggregate_hash)
-				WHERE NOT (is_inner = false AND transaction_type = {TransactionType.MULTISIG.value})
+				WHERE transaction_type <> {TransactionType.MULTISIG.value}
 			'''
 		)
 
@@ -148,7 +148,7 @@ class NemDatabase(DatabaseConnection):
 			CREATE INDEX IF NOT EXISTS transactions_recipient_address_height_idx
 				ON transactions(recipient_address, height DESC, id DESC)
 				INCLUDE (aggregate_hash)
-				WHERE NOT (is_inner = false AND transaction_type = {TransactionType.MULTISIG.value})
+				WHERE transaction_type <> {TransactionType.MULTISIG.value}
 			'''
 		)
 
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -2,6 +2,7 @@ import json
 from binascii import unhexlify
 from collections import namedtuple
 
+from symbolchain.nc import TransactionType
 from symbolchain.nem.Network import Address
 
 from .DatabaseConnection import DatabaseConnection
@@ -133,6 +134,24 @@ class NemDatabase(DatabaseConnection):
 			'''
 		)
 
+		cursor.execute(
+			f'''
+			CREATE INDEX IF NOT EXISTS transactions_sender_address_height_idx
+				ON transactions(sender_address, height DESC, id DESC)
+				INCLUDE (aggregate_hash)
+				WHERE NOT (is_inner = false AND transaction_type = {TransactionType.MULTISIG.value})
+			'''
+		)
+
+		cursor.execute(
+			f'''
+			CREATE INDEX IF NOT EXISTS transactions_recipient_address_height_idx
+				ON transactions(recipient_address, height DESC, id DESC)
+				INCLUDE (aggregate_hash)
+				WHERE NOT (is_inner = false AND transaction_type = {TransactionType.MULTISIG.value})
+			'''
+		)
+
 		# Create indexes for transactions_mosaic table
 		cursor.execute(
 			'''
@@ -250,6 +269,7 @@ class NemDatabase(DatabaseConnection):
 				deadline timestamp NOT NULL,
 				signature bytea,
 				is_inner boolean NOT NULL,
+				aggregate_hash bytea,
 				payload jsonb,
 				size int NOT NULL,
 				version int NOT NULL
@@ -681,11 +701,12 @@ class NemDatabase(DatabaseConnection):
 				deadline,
 				signature,
 				is_inner,
+				aggregate_hash,
 				payload,
 				size,
 				version
 			)
-			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
+			VALUES (%s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s, %s)
 			RETURNING id
 			''',
 			(
@@ -700,6 +721,7 @@ class NemDatabase(DatabaseConnection):
 				transaction.deadline,
 				unhexlify(transaction.signature) if transaction.signature else None,
 				transaction.is_inner,
+				unhexlify(transaction.aggregate_hash) if transaction.aggregate_hash else None,
 				json.dumps(transaction.payload) if transaction.payload else None,
 				transaction.size,
 				transaction.version
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -85,8 +85,9 @@ TransactionRecord = namedtuple('TransactionRecord', [
 	'recipient_address',
 	'payload',
 	'size',
-	'version'
-])
+	'version',
+	'aggregate_hash'
+], defaults=[None])
 DatabaseConfig = namedtuple('DatabaseConfig', ['database', 'user', 'password', 'host', 'port'])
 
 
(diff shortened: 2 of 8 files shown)


Jaguar0625 (commented):
👍

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2497, opened 2026-08-12
[explorer/rest] fix: select the listing page before joining

cryptoBeliever:
## Problem
Three listings did their expensive work for the whole table and only then applied `ORDER BY` and `LIMIT`:

  ```sql
  FROM accounts a
  LEFT JOIN account_remark ar ON ar.address = a.address
  LEFT JOIN LATERAL ( ...jsonb_array_elements(a.mosaics)... ) am ON true
  {where_condition}
  {order_condition}
  {limit_condition}
  ```

For a page of ten rows at `offset=100000`, that `LATERAL` expands a jsonb array for 100 010 accounts
so that ten survive. `transactions` has the same shape for its mosaic aggregation, and `mosaics`
runs four joins — two of them into `blocks` — before the limit applies.

## Solution
The page is selected first and the joins run only for the rows that survive:

  ```sql
  FROM (
      SELECT * FROM accounts a
      {where_condition}
      {order_condition}
      {limit_condition}
  ) a
  LEFT JOIN account_remark ar ON ar.address = a.address
  LEFT JOIN LATERAL ( ... ) am ON true
  {order_condition}
  ```

  The same rewrite applies to `_generate_transaction_sql_query` and `_generate_mosaic_query`. It is
  the shape `_generate_namespace_query` and the account transaction query already use, so this brings
  the remaining three into line rather than introducing a new pattern.

  Every `WHERE` these generators receive refers only to the base table — `accounts`, `transactions` or
  `mosaics` — so nothing needed rewriting to move inside.

### DIFF
--- a/explorer/rest/rest/db/NemDatabase.py
+++ b/explorer/rest/rest/db/NemDatabase.py
@@ -322,7 +322,12 @@ class NemDatabase(DatabaseConnectionPool):
 				cosignatory_of,
 				cosignatories,
 				ar.remark
-			FROM accounts a
+			FROM (
+				SELECT * FROM accounts a
+				{where_condition}
+				{order_condition}
+				{limit_condition}
+			) a
 			LEFT JOIN account_remark ar
 				ON ar.address = a.address
 			LEFT JOIN LATERAL (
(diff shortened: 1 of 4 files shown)


Jaguar0625 (approved):
👍

PR #2498, opened 2026-08-12
[explorer/puller] feat: repair database during rollback

AnthonyLaw:
Problem: Database rollback repair was not implemented or integrated into the sync workflow.

Solution:

- Implemented fork detection in the sync workflow.
- Before each sync, the workflow validates the chain height and checks for a fork.
- When a fork is detected, it captures the rollback impact and prefetches the current state of surviving accounts.
- Once all required data is available, it repairs the database to the confirmed fork height.
- Account state is prefetched from the node to ensure the repair uses clean, authoritative data rather than potentially stale database records.

**Note**: To keep this PR small, namespace and mosaic reconstruction, along with harvested-fee recalculation in `repair_rollback`, will be implemented in a next PR.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -627,7 +627,6 @@ class NemDatabase(DatabaseConnection):
 			for record in results
 		]
 
-
 	def get_accounts_for_refresh(self, limit, last_account_id):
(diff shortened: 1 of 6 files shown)


Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #2499, opened 2026-08-13
[explorer/puller] fix: write account remote links after their accounts exist

cryptoBeliever:
## Problem

Delegated harvesting links (`accounts.remote_address`) were silently dropped during sync.

`_process_account_key_link` issued an immediate `UPDATE accounts SET remote_address = … WHERE address = …`
from `_process_transactions`, which runs per block. Account rows, however, are inserted by
`_flush_pending_account_state` only at the batch boundary, every 50 blocks. When a main account first
appears in the same batch as its own key link transaction, the `UPDATE` matches zero rows.

Nothing recovers afterwards: the flush inserts the account with `remote_address = NULL`
(hardcoded in `_create_account_record`), and the `DO UPDATE` clause of `upsert_account` does not
list that column.

## Solution

The link change is recorded into an accumulator and written in `_flush_pending_account_state`,
after `_process_account_batch` has upserted the accounts — so the row is guaranteed to exist.
This mirrors the existing handling of `accounts_harvested_fee`, which is already applied after
the account upsert for the same reason.

Keying the accumulator by main address gives last-write-wins within a batch, which matches block
order when an account links and unlinks inside one window.

## Backfill

```sql

BEGIN;

  WITH latest_link AS (
        -- the newest key link of each main account is the one in force
        SELECT DISTINCT ON (t.sender_address)
                t.sender_address,
                t.payload->>'mode'           AS mode,
                t.payload->>'remote_account' AS remote_account
        FROM transactions t
        WHERE t.transaction_type = 2049 -- ACCOUNT_KEY_LINK
        ORDER BY t.sender_address, t.height DESC, t.id DESC
  ),
  resolved AS (
        SELECT
                l.sender_address,
                CASE WHEN '1' = l.mode THEN r.address END AS remote_address -- mode 1 activates, 2 deactivates
        FROM latest_link l
        LEFT JOIN accounts r
                ON r.public_key = decode(lower(l.remote_account), 'hex')
        -- an activation whose remote account is unknown would clear the link instead of setting it
        WHERE '1' <> l.mode OR r.address IS NOT NULL
  )
  UPDATE accounts a
  SET remote_address = resolved.remote_address
  FROM resolved
  WHERE a.address = resolved.sender_address AND a.remote_address IS DISTINCT FROM resolved.remote_address;

  COMMIT;

```

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -335,8 +335,11 @@ class NemPuller:
 				last_height
 			)
 
-	def _flush_pending_account_state(self, cursor, pending_addresses, accounts_harvested_fee):
-		"""Flushes accumulated account and harvested fee state, clearing both accumulators."""
+	def _flush_pending_account_state(self, cursor, pending_addresses, accounts_harvested_fee, pending_remote_links):
+		"""Flushes accumulated account state, clearing every accumulator.
+
+		Accounts are upserted first, so the updates that follow always find their row.
+		"""
 
 		if pending_addresses:
 			asyncio.run(self._process_account_batch(cursor, pending_addresses))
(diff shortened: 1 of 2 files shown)


Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

PR #2500, opened 2026-08-13
[explorer] feat: show the account behind a delegated harvesting key

cryptoBeliever:
## Problem

With delegated harvesting, a block is signed by the remote account but credits its fees to the
main one. The block page shows the signer as the harvester, so clicking it lands on the remote
account - and that page is a dead end:

- balance 0
- Linked Keys: *No linked keys*

Nothing tells the user this is a delegated key, or which account it harvests for.
The cause is that the link is stored in one direction only. `accounts.remote_address` on the main
account points at the remote one; the remote account has no column pointing back.

## Solution

The single-account endpoint resolves the reverse direction and returns it as `mainAddress`. The
account page renders it as **Main Account** inside Linked Keys and marks the account with a 
**Linked** label.

### Listings are untouched

The lookup runs only in `_get_account`, never in `get_accounts`, so a page of up to 250 accounts
still costs one query. To keep that honest in the response shape, `mainAddress` is returned by `to_detail_dict()` rather
than `to_dict()` - the same split the Symbol block views already use (`to_dict` for the list,
`to_detail_dict` for the single resource).


Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2501, opened 2026-08-15
[explorer/puller] feat: restore namespace, mosaics and harvested fees from rollback

AnthonyLaw:
Here is part 2.

Problem: 

- Rollback repair did not fully restore the derived blockchain state. A rollback could leave namespace and mosaic inconsistent, while harvested fees and the latest harvested height could still include orphaned blocks.
- Namespace re-registration after the grace period could also retain stale sub-namespaces.

Solution:

- Rebuild affected namespaces by replaying surviving namespace registration transactions up to the fork height.
- Rebuild affected mosaics by replaying surviving mosaic definition and supply change transactions.
- Preserve the seeded `nem.xem` mosaic during restoration.
- Recalculate harvested fees and the latest harvested height from surviving blocks.
- Preserve sub-namespaces for valid same-owner renewals within the grace period.
- Clear stale sub-namespaces when a root namespace is registered after the grace period.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -1083,7 +1083,7 @@ class NemDatabase(DatabaseConnection):  # pylint: disable=too-many-public-method
 	def rollback_account_harvesting(cursor, orphan_harvested_fees_map, fork_height):
 		"""Subtracts orphan block fees and restores the latest surviving harvest height."""
 
-		for beneficiary in sorted(orphan_harvested_fees_map, key=str):
+		for beneficiary, orphan_harvested_fees in orphan_harvested_fees_map.items():
 			cursor.execute(
 				'''
 				UPDATE accounts
@@ -1098,7 +1098,7 @@ class NemDatabase(DatabaseConnection):  # pylint: disable=too-many-public-method
 				WHERE address = %s
 				''',
 				(
-					orphan_harvested_fees_map[beneficiary],
+					orphan_harvested_fees,
 					beneficiary.bytes,
 					fork_height,
 					beneficiary.bytes
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -8,6 +8,8 @@ from symbolchain.nem.Network import Address
 
 from .DatabaseConnection import DatabaseConnection
 
+NEM_NAMESPACE_GRACE_PERIOD = 30 * 1440
+
 AccountRefreshRecord = namedtuple('AccountRefreshRecord', ['id', 'address'])
 RollbackBlockRecord = namedtuple('RollbackBlockRecord', ['beneficiary', 'signer'])
 RollbackTransactionRecord = namedtuple(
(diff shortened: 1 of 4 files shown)


explorer/puller/puller/db/NemDatabase.py
```diff
@@ -950,3 +1077,27 @@ def insert_transaction_mosaic(cursor, transaction_id, mosaic):
 				mosaic.quantity
 			)
 		)
+
+	@staticmethod
+	def recalculate_account_harvesting(cursor, beneficiaries):
```
Jaguar0625: this could become an expensive operation for a beneficiary that has harvested a lot of blocks. is it too difficult to calculate the differences from the rolled-back blocks and subtract?

explorer/puller/tests/db/test_NemDatabase.py
```diff
@@ -1455,3 +1641,187 @@ def test_can_get_latest_surviving_account_key_link(self):
...
+
+			return self._fetch_account_from_db(cursor, account.address)
+
+	def test_can_recalculate_harvesting_from_surviving_blocks(self):
+		# Act:
+		result = self._run_recalculate_account_harvesting_test(ACCOUNTS[0], [BLOCKS[0]])
+
+		# Assert:
+		self.assertEqual(BLOCKS[0].total_fee, result[8])
+		self.assertEqual(BLOCKS[0].height, result[11])
```
Jaguar0625: would be clearer to return named tuple instead of index [8] and [11] in these two tests

Jaguar0625 (approved):
👍

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: I think `SUM(total_fee)` could make this query expensive. I can capture the harvested fees from the rolled-back blocks and subtract them instead. 
However, the of `last_harvested_height` will remain unchanged.

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -950,3 +1078,29 @@ def insert_transaction_mosaic(cursor, transaction_id, mosaic):
 				mosaic.quantity
 			)
 		)
+
+	@staticmethod
+	def rollback_account_harvesting(cursor, orphan_harvested_fees, fork_height):
```
Jaguar0625: good job on the change 👍 

i'd suggest renaming orphan_harvested_fees to indicate that its a map, e.g.

orphan_harvested_fees_map
beneficiary_harvested_fees_map

explorer/puller/puller/db/NemDatabase.py
```diff
@@ -950,3 +1078,29 @@ def insert_transaction_mosaic(cursor, transaction_id, mosaic):
...
+		"""Subtracts orphan block fees and restores the latest surviving harvest height."""
+
+		for beneficiary in sorted(orphan_harvested_fees, key=str):
```
Jaguar0625: i am not sure if there is a reason you need to use sorted vs just iterating over natural ordering via items?

Jaguar0625 (approved):
👍

explorer/puller/puller/db/NemDatabase.py
AnthonyLaw: It makes sense. 
`orphan_harvested_fees_map` sounds better, as it's used in the rollback method.

Jaguar0625 (approved):
👍

PR #2505, opened 2026-08-16
[explorer/frontend] fix: search only the entity type a filter needs

cryptoBeliever:
## Problem

The searchable filters each queried endpoints they have no use for. Typing into one sent a
request for every other entity type as well, and only the matching one was rendered:

| Filter | Endpoint it uses | Endpoints it queried needlessly |
| --- | --- | --- |
| From / To - 40-char address | `account?address=` | `block`, `mosaic`, `namespace` |
| From / To - 64-char public key | `account?publicKey=` | `block`, `transaction`, `mosaic`, `namespace` |
| From / To - partial text | none | `block`, `mosaic`, `namespace` (+ parent namespace) |
| Transferred mosaic | `mosaic` | `block`, `namespace` (+ parent namespace) |
| Block | `block` | `mosaic`, `namespace` |

Two causes:

- The filter modal calls the shared global search, which looks up every entity type by
  design, and then discards all but one (`Filter.jsx`, `searchResult?.[type]`).
- The lookups inside `search` were chained with `await`, so the discarded requests were not
  just extra - they ran one after another ahead of the useful one.

Each miss came back as 400/404 and was swallowed by `createTryFetchInfoFunction`, so the
only visible symptom was the wait after every debounce tick.

## Solution

- `FilterModal` passes its type: `onSearchRequest(text, type)`.
- `search(text, type)` resolves a single handler for that type, so a filter queries exactly
  the endpoint it renders. An unknown type yields an empty result rather than an error.
- Called without a type - the global search bar in `Header.jsx`, which does need every type
  - the full set still runs, now through `Promise.all` rather than a chain of awaits.
- For the account filter partial text now produces no request at all: the account lookup
  was already guarded by text length, and the block, mosaic and namespace lookups that used
  to fire alongside it are gone.

### DIFF
--- a/explorer/frontend/__tests__/api/search.test.js
+++ b/explorer/frontend/__tests__/api/search.test.js
@@ -14,8 +14,7 @@ jest.mock('@/app/utils/server', () => {
 	};
 });
 
-const runSearchTest = async (searchQuery, responseMap, expectedResult) => {
-	// Arrange:
+const mockMakeRequest = responseMap => {
 	const spy = jest.spyOn(utils, 'makeRequest');
 
 	spy.mockImplementation(url => {
@@ -27,11 +26,30 @@ const runSearchTest = async (searchQuery, responseMap, expectedResult) => {
 			return Promise.reject(error404Response);
 	});
 
+	return spy;
+};
+
+const runSearchTest = async (searchQuery, responseMap, expectedResult) => {
+	// Arrange:
+	mockMakeRequest(responseMap);
+
 	// Act:
 	const result = await search(searchQuery);
 
 	// Assert:
 	expect(result).toStrictEqual(expectedResult);
+};
+
+const runScopedSearchTest = async (searchQuery, type, responseMap, expectedResult, expectedRequestedURLs) => {
+	// Arrange:
+	const spy = mockMakeRequest(responseMap);
+
+	// Act:
+	const result = await search(searchQuery, type);
+
+	// Assert:
+	expect(result).toStrictEqual(expectedResult);
+	expect(spy.mock.calls.map(call => call[0])).toStrictEqual(expectedRequestedURLs);
 };
 
 describe('api/search', () => {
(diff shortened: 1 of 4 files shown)


explorer/frontend/__tests__/api/search.test.js
```diff
@@ -161,4 +179,62 @@ describe('api/search', () => {
...
+			'https://explorer.backend/namespace/1': namespaceInfoResponse
+		};
+		const expectedResult = {
+			block: blockInfoResult
+		};
+
+		// Act + Assert:
+		await runScopedSearchTest(searchQuery, 'block', responseMap, expectedResult, ['https://explorer.backend/block/1']);
+	});
+
```
Jaguar0625: do you want to add tests targeting scoped 'transaction' and 'namespace' searches too?

explorer/frontend/__tests__/api/search.test.js
```diff
@@ -161,4 +179,62 @@ describe('api/search', () => {
...
+			'https://explorer.backend/mosaic/namespace-name.sub-name'
+		]);
+	});
+
+	it('requests only the block when scoped to block', async () => {
+		// Arrange:
+		const searchQuery = '1';
+		const responseMap = {
+			'https://explorer.backend/block/1': blockInfoResponse,
+			'https://explorer.backend/namespace/1': namespaceInfoResponse
```
Jaguar0625: is this second mapping required? if not, you could just assert keys(responseMap) instead of last parameter to runScopedSearchTest.

Jaguar0625 (commented):
👍

explorer/frontend/__tests__/api/search.test.js
cryptoBeliever: You're right, it wasn't required. Dropped and moved the assertion to the helper method.

explorer/frontend/__tests__/api/search.test.js
cryptoBeliever: Added scoped tests for namespace and transaction, plus a "requests nothing" case for transaction, so all five handlers are covered.

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2506, opened 2026-08-17
[explorer] fix: list multisig transfers when filtering by mosaic

cryptoBeliever:
## Problem

Filtering the transaction listing by mosaic dropped every transfer wrapped in a multisig
transaction. The puller attaches `transactions_mosaic` rows to TRANSFER transactions, and
for a multisig that is the inner row - but the filter asked the top level row about the
mosaic and additionally forced `transaction_type = TRANSFER`, so the wrapper failed on both
counts.

# Solution

Select the page from the mosaic side and map each link to the transaction the listing shows,
the resolution the address branches already use:

```sql
WHERE t.is_inner = false AND t.id IN (
    SELECT COALESCE(o.id, t.id)                      -- inner transfer → its wrapper
    FROM transactions_mosaic tm
    JOIN transactions t ON t.id = tm.transaction_id
    LEFT JOIN transactions o ON o.transaction_hash = t.aggregate_hash
    WHERE tm.namespace_name = %s
    ORDER BY tm.transaction_id DESC
    LIMIT %s OFFSET %s)
```

LIMIT now applies to the driving index, so the listing reads ten rows instead of 591 838:
9 ms and roughly fifty buffers with the new transactions_mosaic(namespace_name, transaction_id) index.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -175,6 +175,13 @@ class NemDatabase(DatabaseConnection):
 			'''
 		)
 
+		cursor.execute(
+			'''
+			CREATE INDEX IF NOT EXISTS transactions_mosaic_namespace_name_transaction_id_idx
+				ON transactions_mosaic(namespace_name, transaction_id)
+			'''
+		)
+
 	def create_tables(self):
 		"""Creates database tables."""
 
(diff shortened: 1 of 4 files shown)


Jaguar0625 (commented):
👍

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2508, opened 2026-08-18
[explorer/frontend] fix: keep 50 blocks in the home visualisation after refresh

cryptoBeliever:
## Problem

The recent blocks visualisation on the home page shows 50 blocks when the page
loads and shrinks to 10 roughly a minute later, then stays that way.

`getServerSideProps` preloads the strip with `fetchBlockPage({ pageSize: 50 })`,
but the 60 s refresh went through `useAsyncCall(fetchBlockPage, preloadedBlocks,
DATA_REFRESH_INTERVAL)`. `useAsyncCall` invokes its callback with no arguments,
so `createSearchCriteria` fell back to its default of 10 rows per page
(`utils/server.js`). The first poll therefore replaced the preloaded 50 blocks
with 10.

## Solution

- `RECENT_BLOCK_COUNT` is now a single constant used by both `getServerSideProps`
  and the polling call.
- The polling call is wrapped in an arrow function, so it keeps passing
  `{ pageSize: RECENT_BLOCK_COUNT }` on every refresh - matching how the latest
  and pending transaction sections on the same page are already written.

### DIFF
--- a/explorer/frontend/__tests__/components/BlockPreview.test.js
+++ b/explorer/frontend/__tests__/components/BlockPreview.test.js
@@ -1,0 +1,175 @@
+import '@testing-library/jest-dom';
+import BlockPreview from '@/app/components/BlockPreview';
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import TimeAgo from 'javascript-time-ago';
+import en from 'javascript-time-ago/locale/en.json';
+
+TimeAgo.addDefaultLocale(en);
+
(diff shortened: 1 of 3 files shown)


Jaguar0625 (commented):
is it possible to add a regression test for this behavior?

Jaguar0625 (approved):
👍

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2509, opened 2026-08-18
[explorer/rest] fix: cap listing page size

cryptoBeliever:
## Problem

The NEM listing routes validated only that `limit` was not negative, so a single request could
ask the database for any number of rows - `/api/nem/transactions?limit=1000000` was a valid
request that the database had to serve in full. The same check was copy-pasted into six routes.

## Solution
 
Added `MAX_PAGE_LIMIT = 250` and a shared `_validate_pagination(limit, offset)` helper in
`rest/routes/nem.py`. Every paginated NEM route (`blocks`, `accounts`, `namespaces`, `mosaics`,
`mosaic/rich/list`, `transactions`) now rejects a `limit` outside `1..250` and a negative
`offset` with HTTP 400, replacing all six copies of the old inline check.

Related to https://github.com/symbol/product/pull/2510


Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2510, opened 2026-08-18
[explorer/frontend] fix: stop requesting a whole block in a single call

cryptoBeliever:
## Problem
The block page asked for every transaction of the block at once - `fetchTransactionPage` was called with `pageSize: blockInfo.transactionCount` - and paged the result on the client. Nothing was lost, because `/api/nem/transactions` does not bound `limit`. The fee treemap made it worse by needing the same full set. Above its own guard of 400 data points it rendered nothing at all - not a chart, not a message, just an empty field - so the cost was paid for a block whose chart was never shown.

## Solution
The table now uses ordinary server-side pagination, 50 transactions per page, through `usePagination`. The treemap keeps its own single request, capped at `MAX_TRANSACTION_SQUARES`, and is skipped entirely for blocks above that size - no request is sent and the chart is replaced with `message_tooManyTransactionsToVisualize` rather than a partial picture. The chart no longer decides from the length of the data it received, but from the block's `transactionCount`, which both call sites pass, and the preview fetch was aligned with the same cap so its data is complete whenever the chart is drawn. `useClientSidePagination` had no remaining users and was removed.

Related to https://github.com/symbol/product/pull/2509

### DIFF
--- a/explorer/frontend/__tests__/pages/BlockInfo.test.js
+++ b/explorer/frontend/__tests__/pages/BlockInfo.test.js
@@ -1,9 +1,14 @@
 import '@testing-library/jest-dom';
 import { blockInfoResult } from '../test-utils/blocks';
+import { transactionPageResult } from '../test-utils/transactions';
 import * as BlockService from '@/app/api/blocks';
+import * as TransactionService from '@/app/api/transactions';
+import { MAX_TRANSACTION_SQUARES } from '@/app/components/ValueTransactionSquares';
 import BlockInfo, { getServerSideProps } from '@/app/pages/blocks/[height]';
 import * as utils from '@/app/utils';
-import { render, screen } from '@testing-library/react';
+import { act, render, screen } from '@testing-library/react';
+/* eslint-disable import/no-unresolved */
+import { mockAllIsIntersecting } from 'react-intersection-observer/test-utils';
 
 jest.mock('@/app/utils', () => {
 	return {
@@ -16,6 +21,13 @@ jest.mock('@/app/api/blocks', () => {
 	return {
 		__esModule: true,
 		...jest.requireActual('@/app/api/blocks')
+	};
+});
+
+jest.mock('@/app/api/transactions', () => {
+	return {
+		__esModule: true,
+		...jest.requireActual('@/app/api/transactions')
 	};
 });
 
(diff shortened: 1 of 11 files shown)


explorer/frontend/__tests__/pages/BlockInfo.test.js
```diff
@@ -115,4 +127,95 @@ describe('BlockInfo', () => {
...
+
+			// Assert:
+			expect(fetchTransactionPage).toHaveBeenCalledWith({
+				height: blockInfoResult.height,
+				pageSize: MAX_TRANSACTION_SQUARES
+			});
+			expect(screen.queryByText('message_tooManyTransactionsToVisualize')).not.toBeInTheDocument();
+		});
+
+		it('requests the fee visualisation for a block sitting exactly on the cap', async () => {
```
Jaguar0625: how is this validating the fee visualization is being requested?

explorer/frontend/__tests__/pages/BlockInfo.test.js
```diff
@@ -115,4 +127,95 @@ describe('BlockInfo', () => {
...
+
+			// Assert:
+			expect(fetchTransactionPage).toHaveBeenCalledWith({
+				pageNumber: 2,
+				height: blockInfoResult.height,
+				pageSize: TRANSACTION_PAGE_SIZE
+			});
+		});
+
+		it('requests all the transactions of the block for the fee visualisation', async () => {
```
Jaguar0625: isn't the action same as 'requests the first transaction page of the block'. only difference is you're checking last (vs first) page?

Jaguar0625 (approved):
👍

explorer/frontend/__tests__/pages/BlockInfo.test.js
cryptoBeliever: Yes, the action is the same, the two tests asserted different requests: the table's first page, and the single request the fee visualisation data comes from. Merging them still makes sense - done, as `requests the first transaction page and the fee visualisation data`.

explorer/frontend/__tests__/pages/BlockInfo.test.js
cryptoBeliever: The name was misleading. It checks that a block exactly on the cap is still visualised: the treemap request is sent, and the cap message is not shown. Renamed to `keeps the fee visualisation when the block sits exactly on the cap`.

Jaguar0625 (approved):
👍

AnthonyLaw (approved):
👍🏼

PR #2511, opened 2026-08-18
[explorer/puller] fix: refresh accounts stuck in a transient remote status

cryptoBeliever:
## Problem

`accounts.remote_status` mirrors the NIS `meta.remoteStatus` field. Two of its
values are transient - `ACTIVATING` and `DEACTIVATING` - and settle into
`ACTIVE` / `INACTIVE` roughly six hours after a delegated harvesting key link.

The puller writes `remote_status` in exactly one place, `_process_account_batch`,
which runs only when a block touches the account. An account whose status is
captured mid-transition keeps that stale value indefinitely, until some later
transaction happens to touch it again.

## Solution

- `get_accounts_for_refresh` additionally selects accounts whose `remote_status`
  is `ACTIVATING` or `DEACTIVATING`.
- The refresh `UPDATE` now carries `remote_status` from the node response.

### DIFF
--- a/explorer/puller/puller/db/NemDatabase.py
+++ b/explorer/puller/puller/db/NemDatabase.py
@@ -448,7 +448,12 @@ class NemDatabase(DatabaseConnection):
 			SELECT id, address
 			FROM accounts
 			WHERE id > %s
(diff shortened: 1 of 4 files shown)


Jaguar0625 (approved):
👍

PR #2513, opened 2026-08-19
[Explorer/puller] fix: symbol effective fee overflow

daoka:

Symbol Puller calculates a transaction's effective fee with:

```text
min(max_fee, transaction_size * block_fee_multiplier)
```

Although `symbol_transactions.effective_fee` is a PostgreSQL `bigint`, both operands of the multiplication are `integer` values. PostgreSQL therefore evaluates the multiplication as a 32-bit integer before storing the result.

## What's the issue?

Live testnet synchronization failed at height 47,470 with:

```text
psycopg2.errors.NumericValueOutOfRange: integer out of range
```

The affected aggregate transaction had the following values:

```text
transaction size:     110568
block fee multiplier: 45221
calculated fee:       4999995528
max fee:              5000000000
```

The calculated fee exceeds PostgreSQL's 32-bit integer limit even though it fits in the existing `bigint` destination column. This prevents transaction persistence and stops block synchronization.

- Cast the transaction size to `bigint` before multiplying it by the block fee multiplier.
- Preserve the existing effective fee behavior:

  ```text
  effective_fee = min(max_fee, transaction_size * block_fee_multiplier)
  ```

- Add PostgreSQL integration tests covering:
  - a calculated fee above the 32-bit integer range;
  - a calculated fee above the 32-bit range that is capped by `max_fee`.
- Preserve the existing `NULL` effective fee behavior for embedded transactions.

This change does not require a schema migration because `effective_fee` is already stored as `bigint`.

This is not a breaking change.

### DIFF
--- a/explorer/puller/puller/db/SymbolDatabase.py
+++ b/explorer/puller/puller/db/SymbolDatabase.py
@@ -1870,7 +1870,7 @@ class SymbolDatabase(DatabaseConnection):  # pylint: disable=too-many-public-met
 				%(deadline)s,
 				%(network_deadline)s,
 				%(max_fee)s,
-				LEAST(%(max_fee)s, %(size)s * (SELECT fee_multiplier FROM symbol_blocks WHERE height = %(height)s)),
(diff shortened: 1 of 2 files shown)


Jaguar0625 (approved):
👍

PR #2514, opened 2026-08-19
[explorer/puller] fix: validate Symbol block parent hash continuity

daoka:

The Symbol Puller validates that fetched block heights are sequential, but it does not verify that each block's `previousBlockHash` matches the preceding block hash.

## What's the issue?

If a Symbol node switches branches while multiple block pages are being fetched, the Puller can receive sequential block heights from different branches. Without parent hash validation, this discontinuous sequence could be persisted to the Explorer database.

Refs #2353

The Puller now validates parent hash continuity for newly fetched blocks across:

- the last stored block and the first newly fetched block
- blocks within the same page
- consecutive pages
- consecutive page-fetch batches

Validation happens before fetching related transaction and receipt data or writing the current batch. A mismatch raises a deterministic `ValueError` and prevents the invalid block batch from being persisted.

The validation uses already fetched block data. It adds no Symbol node requests and only adds one database hash lookup when synchronization starts above height 1.

This is not a breaking API change.

This change does not scan or repair inconsistencies already stored in historical data.


explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -1339,3 +1341,20 @@ def _validate_block_page(rows, expected_start_height):
 			expected_height = expected_start_height + index
 			if row['height'] != expected_height:
 				raise ValueError(f'Unexpected Symbol block height {row["height"]}; expected {expected_height}')
+
+	@staticmethod
+	def _validate_block_chain(rows, previous_block_hash):
+		for row in rows:
+			if 1 < row['height']:
+				expected_hash = bytes(previous_block_hash) if previous_block_hash is not None else None
+				actual_previous_hash = bytes(row['previous_hash'])
```
Jaguar0625: there is a slight inconsistency here where you allow previous_block_hash to be None but not  row['previous_hash'].

given this and handling for nemesis block above, not sure if you need to filter out nemesis block vs verifying its parent hash is empty/None

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -765,6 +765,182 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		# Act:
+		block_heights, sync_state = self._sync_with_connector(connector)
+
+		# Assert:
+		self.assertEqual([1, 2, 3], block_heights)
+		self.assertEqual(3, sync_state['last_synced_height'])
+
+	def test_sync_block_headers_accepts_height_one_without_stored_parent_hash(self):
+		# Arrange:
+		connector = FakeConnector(1, {0: [create_node_block(1, previous_hash='A' * 64)]})
```
Jaguar0625: test is named without_stored_parent_hash, but you seem to be sending nemesis block with parent hash (which probably should be rejected)?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -765,6 +765,182 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		# Act / Assert:
+		with self.assertRaisesRegex(
+			ValueError,
+			f'Symbol block chain mismatch at height {chain_height}: '
+			f'expected previous hash {chain_height - 1:064X}, got {"A" * 64}'
+		):
+			asyncio.run(self.puller.sync_block_headers())
+
+		# Assert:
+		self.assertEqual(list(range(1, chain_height)), self._fetch_block_heights(self.puller.symbol_db))
```
Jaguar0625: do you intend to keep blocks in database despite failure?

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/SymbolPuller.py
daoka: Thanks. Symbol's nemesis block uses a 32-byte zero previousBlockHash, as confirmed against mainnet /blocks/1 and the Symbol core implementation. I now validate height 1 against bytes(32), while height 2 and later must match the preceding block hash. The unreachable None allowance and its error-message branch have been removed.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Agreed. The fixture was invalid. I removed the nonzero parent-hash override and renamed the test to verify acceptance of the canonical zero previous hash. I also added a separate test that rejects a height-1 block with a nonzero previous hash. The valid parent-hash chain test for height 2 and later remains unchanged.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Yes. Per-batch persistence predates this PR and is intentionally retained; the batch containing the mismatch is not written, while earlier committed batches remain. However, the parent-hash validation exposed a recovery gap because those rows can remain above the unchanged watermark. This change adds an atomic write-ahead dirty_state_from_height marker immediately before the first database write of each batch. On restart, the marked range is repaired before forward synchronization, including derived namespace, mosaic, and metadata state, and the marker is cleared only after terminal success. Caught failures attempt to mark the state unhealthy without masking the primary exception. An explicit max_height does not modify data outside its cap; capped dirty repairs stop with the marker retained, except when the node's raw chain height proves the upper tail orphaned by chain shortening. The safety guarantee is durable restart convergence, not immediate fail-closed behavior. Overlapping runs remain prohibited by the external scheduler contract.

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -3061,6 +3063,83 @@ def test_get_sync_state_returns_upserted_value(self):
...
+		database = self._create_database()
+		database.mark_sync_write_intent(10)
+		database.mark_sync_write_intent(20)
+		database.mark_sync_write_intent(5)
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT dirty_state_from_height, updated_at FROM symbol_sync_state WHERE id = 1')
+		before = cursor.fetchone()
+
+		# Act:
+		database.mark_sync_write_intent(5)
```
Jaguar0625: while test is correct, i'd sugest not calling `mark_sync_write_intent` with 5 twice in a row

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -3061,6 +3063,83 @@ def test_get_sync_state_returns_upserted_value(self):
...
+	def test_mark_sync_unhealthy_rolls_back_update_failure_and_keeps_connection_usable(self):
+		# Arrange:
+		database = self._create_database()
+		database.upsert_sync_state(_create_sync_state(dirty_state_from_height=3))
+		cursor = database.connection.cursor()
+		cursor.execute('DROP TABLE symbol_sync_state')
+
+		# Act:
+		with self.assertRaises(PsycopgError):
+			database.mark_sync_unhealthy()
```
Jaguar0625: this is failing because table was dropped?
if so, how does subsequent call succeed?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -3860,6 +3939,24 @@ def test_rejects_duplicate_block_hash_for_different_heights(self):
...
+		database = self._create_database()
+		database.upsert_blocks([_create_block(1)])
+		invalid_block = {**_create_block(3), 'hash': None}
+
+		# Act:
+		with self.assertRaises(PsycopgError):
+			database.upsert_blocks([_create_block(2), invalid_block])
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT height FROM symbol_blocks ORDER BY height')
+		self.assertEqual([(1,)], cursor.fetchall())
```
Jaguar0625: either mark with // Assert: or // Sanity:

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -3886,6 +3983,31 @@ def test_rejects_invalid_receipt_group(self):
...
+			database.upsert_receipts_for_height(
+				1,
+				[_create_receipt(1, amount=75), _create_receipt(1, receipt_group='invalid')],
+				75)
+		cursor = database.connection.cursor()
+		cursor.execute('SELECT amount FROM symbol_receipts')
+		self.assertEqual([(50,)], cursor.fetchall())
+		cursor.execute('SELECT block_reward FROM symbol_blocks WHERE height = 1')
+		self.assertEqual((50,), cursor.fetchone())
+		database.upsert_receipts_for_height(1, [_create_receipt(1, amount=75)], 75)
```
Jaguar0625: similarly, please clearly mark the act / assert two stage test like this

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4287,6 +4430,110 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+			finalized_height=1,
+			last_synced_height=1,
+			last_synced_block_hash=b'hash 1',
+			dirty_state_from_height=1))
+		before_blocks = fetch_full_block_state(database)
+		before_sync_state = fetch_normalized_sync_state(database)
+
+		# Act / Assert:
+		with self.assertRaises(KeyError):
+			database.repair_rollback_from_height(
```
Jaguar0625: what is causing the error violation?

explorer/puller/tests/db/test_SymbolDatabase.py
```diff
@@ -4287,6 +4430,110 @@ def test_can_repair_rollback_from_height_in_one_transaction(self):
...
+		self.assertEqual(
+			[(replacement_namespace['namespace_id'], 'replacement', 0)],
+			cursor.fetchall())
+		self.assertEqual([create_persisted_mosaic_state(replacement_mosaic, [])], fetch_mosaic_state(database))
+		self.assertEqual([replacement_metadata], fetch_metadata_rows(database))
+		sync_state = database.get_sync_state()
+		self.assertEqual('repairing', sync_state['status'])
+		self.assertIsNone(sync_state['last_synced_height'])
+		self.assertIsNone(sync_state['last_synced_block_hash'])
+		self.assertEqual(1, sync_state['dirty_state_from_height'])
```
Jaguar0625: test is fine, but do you ever expect this situation to occur - no blocks or transactions but state present (e.g. namespaces)

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -190,32 +191,76 @@ async def post_symbol_node(self, url_path, request_payload, property_name=None,
...
+		raw_sync_state = self.symbol_db.get_sync_state()
+		if is_finalization_capped and raw_sync_state:
+			raw_last_synced_height = raw_sync_state['last_synced_height']
+			if raw_last_synced_height and raw_last_synced_height >= finalized_height:
+				finalized_hash = self.symbol_db.get_block_hash(finalized_height)
+
+		dirty_state_from_height = raw_sync_state['dirty_state_from_height'] if raw_sync_state else None
+		node_chain_shorter_than_dirty = (
+			dirty_state_from_height is not None and node_chain_height < dirty_state_from_height)
+		dirty_range_outside_cap = False
```
Jaguar0625: since this is boolean, please prefix variable with "is_"

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -190,32 +191,76 @@ async def post_symbol_node(self, url_path, request_payload, property_name=None,
...
+		start_height = await self._repair_unfinalized_rollback(
+			raw_sync_state,
+			finalized_height,
+			finalized_hash,
+			chain_height,
+			dirty_state_from_height=None if dirty_range_outside_cap else dirty_state_from_height,
+			node_chain_height=node_chain_height,
+			verify_unfinalized=not dirty_range_outside_cap)
+		if dirty_range_outside_cap:
+			self.symbol_db.mark_sync_unhealthy()
```
Jaguar0625: why is this unhealthy condition?

explorer/puller/puller/facade/SymbolPuller.py
```diff
@@ -269,9 +317,23 @@ def _get_bounded_sync_state(self, sync_state, chain_height):
...
-	async def _repair_unfinalized_rollback(self, sync_state, finalized_height, finalized_hash):
+	async def _repair_unfinalized_rollback(  # pylint: disable=too-many-arguments,too-many-positional-arguments,too-many-locals
+		self,
+		sync_state,
+		finalized_height,
+		finalized_hash,
+		chain_height,
+		dirty_state_from_height=None,
+		node_chain_height=None,
+		verify_unfinalized=True
```
Jaguar0625: would rename to should_*

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -764,6 +796,563 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+				self._sync_with_connector(second_connector)
+		finally:
+			self.puller.symbol_db = database
+
+		# Assert:
+		sync_state = database.get_sync_state()
+		self.assertEqual('unhealthy', sync_state['status'])
+		self.assertEqual(1, sync_state['dirty_state_from_height'])
+
+	def test_sync_block_headers_keeps_dirty_tail_outside_explicit_cap_untouched(self):
```
Jaguar0625: test looks ok, but could you explain how we would get into scenario where dirty_state_from_height is greater than chain height. also, why wouldn't a successful resync reset it?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -764,6 +796,563 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		# Act:
+		asyncio.run(self.puller.sync_block_headers(max_height=4))
+
+		# Assert:
+		self.assertEqual([1, 2, 3, 4, 5], self._fetch_block_heights(database))
+		self.assertEqual(0, connector.paths.count('blocks/4'))
+		self.assertEqual([], [path for path in connector.paths if path.startswith('blocks?page')])
+		sync_state = database.get_sync_state()
+		self.assertEqual('unhealthy', sync_state['status'])
+		self.assertEqual(3, sync_state['dirty_state_from_height'])
```
Jaguar0625: in this case, you have just synced through block 4, so shouldn't dirty_state start at next block (5) from original bad sync?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -764,6 +796,563 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		detail_indices = [
+			index for index, path in enumerate(connector.paths)
+			if path.startswith(('namespaces/', 'mosaics', 'metadata?'))
+		]
+		page_indices = [
+			index for index, path in enumerate(connector.paths)
+			if path.startswith('blocks?page')
+		]
+		self.assertTrue(detail_indices)
+		self.assertTrue(page_indices)
```
Jaguar0625: could you check length of these instead of just non-empty?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -764,6 +796,563 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		asyncio.run(self.puller.sync_block_headers(max_height=2))
+
+		# Assert:
+		self.assertEqual([1, 2, 3], self._fetch_block_heights(database))
+		self.assertEqual([], [path for path in connector.paths if path.startswith('blocks?page')])
+		sync_state = database.get_sync_state()
+		self.assertEqual('healthy', sync_state['status'])
+		self.assertEqual(2, sync_state['chain_height'])
+		self.assertEqual(2, sync_state['last_synced_height'])
+		self.assertIsNone(sync_state['dirty_state_from_height'])
```
Jaguar0625: yes, this behavior is what i expected in test_sync_block_headers_keeps_partially_capped_dirty_tail_untouched. can you explain how these tests are different?

explorer/puller/tests/facade/symbol/test_PullerSync.py
```diff
@@ -764,6 +796,563 @@ def test_sync_block_headers_persists_synced_block_watermark(self):
...
+		self.assertEqual(1, sync_state['dirty_state_from_height'])
+		block_paths = [path for path in connector.paths if path.startswith('blocks?pageSize=100&offset=')]
+		transaction_paths = [path for path in connector.paths if path.startswith('transactions/confirmed?')]
+		statement_paths = [path for path in connector.paths if path.startswith('statements/transaction?')]
+		self.assertEqual(BLOCK_PAGE_FETCH_CONCURRENCY + 1, len(block_paths))
+		self.assertEqual([transaction_path(1, chain_height - 1)], transaction_paths)
+		self.assertEqual([statement_path(1, chain_height - 1)], statement_paths)
+
+		# Act: restart with the valid node chain so the persisted dirty tail is repaired before forward sync.
+		repaired_pages = {
```
Jaguar0625: isn't this just pages without last element?

Jaguar0625 (commented):
👍

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Agreed. I removed the repeated call at height 5 and narrowed this test to minimum-height convergence. The neighboring test already covers the no-op behavior and `updated_at` preservation for an equal or higher height.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes. The first call fails because `symbol_sync_state` was dropped in the current uncommitted transaction. `mark_sync_unhealthy()` rolls the transaction back on failure, and PostgreSQL transactional DDL restores the dropped table. The second call then succeeds on the same reusable connection. I added an explicit table-existence assertion between the two calls.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Done. I separated the failure expectation, rollback-state assertion, connection-reuse action, and final successful-state assertion into explicit stages.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Done. I split the receipt test into explicit failure, rollback verification, retry, and successful-state stages. I applied the same structure to the four transaction-boundary tests added by this PR without changing pre-existing tests outside the diff.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: The injected mosaic entry intentionally omits the required `raw_payload` field. This raises `KeyError('raw_payload')` in `_execute_upsert_mosaic()` after the height-scoped deletes have executed, allowing the test to verify transaction-wide rollback. I renamed the fixture and now assert the exact key error.

explorer/puller/tests/db/test_SymbolDatabase.py
daoka: Yes, as an intermediate rollback-repair state. The repair transaction deletes height-scoped chain rows and applies namespace, mosaic, and metadata state re-fetched from the node in the same transaction. If execution stops before the following forward sync, node-confirmed current state can temporarily remain without its local block rows. The facade test verifies that the next run completes forward sync and returns to a terminal healthy state.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Done. I renamed the variables to `is_node_chain_shorter_than_dirty` and `is_dirty_repair_blocked_by_cap`. The latter also describes the actual condition more directly than the previous range-oriented name.

explorer/puller/puller/facade/SymbolPuller.py
daoka: The dirty marker records only the lower bound of the potentially dirty tail. When an explicit cap is below the raw node height, the puller cannot repair that tail without potentially modifying rows outside the operator-requested scope. It therefore retains the marker, skips dirty-tail repair and unfinalized verification, marks the state unhealthy, and stops before forward sync. Finalized safety checks still run.

explorer/puller/puller/facade/SymbolPuller.py
daoka: Done. I renamed the argument to `should_verify_unfinalized` and updated all call sites. I also added the boolean-prefix rule to the local coding guidelines for future changes.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: In this test, the lower chain height is the effective height produced by `max_height`, not the raw node height. A previous run started writing at height 5, leaving watermark 4 and dirty marker 5. The node is still at height 5, but a later run is capped at height 2. A successful uncapped resync would clear the marker; this capped run cannot repair height 5 without crossing the cap, so it stops unhealthy with the marker retained. I corrected the fixture to model that sequence explicitly.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Agreed; the original fixture was inconsistent. This test is intended to cover a dirty tail that partially overlaps the cap, so I changed the previous watermark to 2 and the dirty start to 3. The potentially dirty range is now 3–5 while the operator cap is 4. The separate fully-outside-cap test uses watermark 4 and dirty start 5.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: Done. I tightened the filter to exclude the startup native-mosaic request and now assert the complete ordered sequence: one namespace request, one mosaic request, one metadata request, followed by one block-page request. This verifies both the exact counts and the repair-before-forward-sync ordering.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: The distinction is whether the raw node height proves the upper tail is orphaned. In the partially capped case, the node is at height 5, the dirty tail starts at 3, and the cap is 4; repairing could modify valid row 5 outside the cap, so synchronization stops unhealthy. In the node-shortening case, the node is at height 3 while the dirty marker is 5; rows 4–5 are therefore proven orphaned and can be removed before applying the cap. I aligned the fixtures so this difference is explicit.

explorer/puller/tests/facade/symbol/test_PullerSync.py
daoka: It is not removed; the final page is replaced. The original `pages` mapping contains a last page with an invalid previous hash. The restart fixture keeps every earlier page and overrides only `last_page_offset` with the valid final block. I simplified the construction to make that replacement explicit.

Jaguar0625 (approved):
👍

PR #2517, opened 2026-08-19
[explorer/puller] fix: add timestamps to NEM workflow logs

cryptoBeliever:
Problem:
zenlog configures its handler with colorlog.ColoredFormatter and a format that carries no timestamp. ColoredFormatter emits ANSI escapes unconditionally, so it never checks whether the stream is a terminal. Both NEM workflows run from cron with output redirected to a file, which leaves puller.log full of literal
"ESC[32m    i    ESC[0m | ESC[32m" prefixes and no way to tell when any line was
written - the log shows repeated sync cycles that cannot be correlated with a
block height or an incident.

Solution:
configure_logging replaces the formatter on the zenlog handler with a plain logging.Formatter that prefixes every record with an ISO-like timestamp and drops the color escapes. sync_nem_block and refresh_nem_accounts call it first thing in main.

### DIFF
--- a/explorer/puller/puller/workflows/log_utils.py
+++ b/explorer/puller/puller/workflows/log_utils.py
@@ -1,0 +1,12 @@
+import logging
+
+from zenlog import log
+
+LOG_FORMAT = '  %(asctime)s | %(styledname)-8s | %(message)s'
+LOG_DATE_FORMAT = '%Y-%m-%d %H:%M:%S'
+
+
+def configure_logging():
+	"""Replaces the colorized zenlog output format with a timestamped plain text one."""
+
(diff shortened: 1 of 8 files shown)


Jaguar0625 (approved):
👍

Jaguar0625:
not sure if you want to update symbol scripts too in symbol.py:main?

cryptoBeliever:
@Jaguar0625 Added it for Symbol as well (CC: @daoka). I also added nem_workflow_utils to de-duplicate the common setup in both NEM workflows.

Jaguar0625 (approved):
👍

PR #2520, opened 2026-08-20
[explorer/rest] fix: give the NEM transaction listing a consistent order

cryptoBeliever:
Problem:
The unfiltered listing ordered only by height. Heights repeat within a block, and a tie the database is free to break any way it likes makes a page depend on the plan: under a parallel scan six identical queries returned six different pages, with no row in common between the first and the last.

Solution:
Move the id tiebreaker from the two branch paths into the shared order_condition so it covers the unfiltered path too.

Migration:

```

CREATE INDEX CONCURRENTLY IF NOT EXISTS transactions_height_id_is_inner_false_idx
	ON transactions(height DESC, id DESC)
	WHERE is_inner = false;

DROP INDEX CONCURRENTLY IF EXISTS transactions_height_is_inner_false_idx;

```


Jaguar0625 (approved):
👍

PR #2525, opened 2026-08-25
[explorer/puller] feat: handle verify current block hash and previous hash

AnthonyLaw:
Problem: Fork detection checked only the current block hash. To make detection more reliable, it should also verify the previous block hash.
Solution: Added previous block hash verification.

### DIFF
--- a/explorer/puller/puller/facade/NemPuller.py
+++ b/explorer/puller/puller/facade/NemPuller.py
@@ -187,9 +187,22 @@ class NemPuller:
 		for height in range(comparison_height, minimum_height - 1, -1):
 			db_block_hash = self.nem_db.get_block_hash(height)
 			node_block = await self._retry_get_block(height)
+			current_hash_matches = db_block_hash.lower() == node_block.block_hash.lower()
 
-			if db_block_hash.lower() == node_block.block_hash.lower():
-				return None if height == db_height else height
+			if 1 == height:
+				if current_hash_matches:
+					return None if height == db_height else height
+
+				continue
+
+			if current_hash_matches:
+				db_previous_block_hash = self.nem_db.get_block_hash(height - 1)
+				previous_hash_matches = db_previous_block_hash.lower() == node_block.previous_block_hash.lower()
+
+				if previous_hash_matches:
+					return None if height == db_height else height
+
+				continue
 
 		raise NemRollbackError(
 			f'No matching NEM block hash found within {NEM_MAX_ROLLBACK_DEPTH} blocks '
--- a/explorer/puller/tests/facade/test_NemPuller.py
+++ b/explorer/puller/tests/facade/test_NemPuller.py
@@ -3,6 +3,7 @@ import asyncio
 import datetime
 import tempfile
 import unittest
+from collections import namedtuple
 from unittest.mock import AsyncMock, Mock, call, patch
 
 import testing.postgresql
@@ -46,6 +47,8 @@ from puller.facade.NemPuller import (
 )
 
 # region test data
+
+NodeBlockHashes = namedtuple('NodeBlockHashes', ['block_hash', 'previous_block_hash'])
 
 NEM_CONNECTOR_RESPONSE_BLOCKS = [
 	Block(
@@ -1599,12 +1602,23 @@ class NemPullerTest(unittest.TestCase):  # pylint: disable=too-many-public-metho
 			'pending_remote_links': pending_remote_links
 		})
 
-	def _run_detect_rollback_test(self, database_hashes, node_hashes, db_height, chain_height):
+	def _run_detect_rollback_test(
+		self,
+		database_hashes_map,
+		node_block_hashes_map,
+		db_height,
+		chain_height
+	):
 		self.puller.nem_db.get_block_hash = Mock(
-			side_effect=lambda height: database_hashes.get(height, 'db-hash')
+			side_effect=lambda height: database_hashes_map.get(height, 'db-hash')
 		)
 		self.puller._retry_get_block = AsyncMock(  # pylint: disable=protected-access
-			side_effect=lambda height: Mock(block_hash=node_hashes.get(height, 'chain-hash'))
+			side_effect=lambda height: Mock(
+				**node_block_hashes_map.get(
+					height,
+					NodeBlockHashes('chain-hash', 'chain-previous-hash')
+				)._asdict()
+			)
 		)
 
 		# Act:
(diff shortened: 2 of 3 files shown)


explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -1599,34 +1602,110 @@ def test_can_process_transactions_inner_outer(self, mock_process_transaction):
...
+			)
 		)
 
 		# Act:
 		return asyncio.run(self.puller.detect_rollback(db_height, chain_height))
 
 	def test_returns_none_when_comparison_hash_matches(self):
 		# Act + Assert: no exception
-		self._run_detect_rollback_test({3: 'ABCDEF'}, {3: 'abcdef'}, 3, 3)
+		self._run_detect_rollback_test(
```
Jaguar0625: should you check it returns None on non-fork cases like this ?

explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -1599,34 +1602,110 @@ def test_can_process_transactions_inner_outer(self, mock_process_transaction):
...
+
+	def test_returns_none_when_nemesis_hash_matches(self):
+		# Act + Assert: no exception
+		self._run_detect_rollback_test(
+			{1: 'common'},
+			{1: NodeBlockHashes('COMMON', 'unused')},
+			1,
+			1
+		)
+		self.puller.nem_db.get_block_hash.assert_called_once_with(1)
```
Jaguar0625: is there reason you're checking puller._retry_get_block sometimes and self.puller.nem_db.get_block_hash other times ?

might make sense to check both in helper ?

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -187,9 +187,22 @@ async def detect_rollback(self, db_height, chain_height):
...
 			node_block = await self._retry_get_block(height)
+			current_hash_matches = db_block_hash.lower() == node_block.block_hash.lower()
 
-			if db_block_hash.lower() == node_block.block_hash.lower():
-				return None if height == db_height else height
+			if 1 == height:
+				if current_hash_matches:
+					return None if height == db_height else height
+
+				continue
```
Jaguar0625: consider adding test where nemesis hash does not match

Jaguar0625 (commented):
👍

explorer/puller/puller/facade/NemPuller.py
```diff
@@ -187,9 +187,22 @@ async def detect_rollback(self, db_height, chain_height):
 		for height in range(comparison_height, minimum_height - 1, -1):
 			db_block_hash = self.nem_db.get_block_hash(height)
 			node_block = await self._retry_get_block(height)
+			current_hash_matches = db_block_hash.lower() == node_block.block_hash.lower()
```
cryptoBeliever: @AnthonyLaw 

My expectation reading the change was that the parent-hash check would land at the point where blocks are fetched, the way it works on the Symbol side, rather than inside detect_rollback.

Symbol's equivalent, for reference (added in 64a33546, "validate Symbol block parent hash continuity"):
- explorer/puller/puller/facade/SymbolPuller.py:1424 - _validate_block_chain(rows, previous_block_hash), which walks every fetched row and requires row['previous_hash'] to equal the
running hash of the preceding block.
- explorer/puller/puller/facade/SymbolPuller.py:441 - the chain is seeded from the database: previous_block_hash = self.symbol_db.get_block_hash(start_height - 1) if start_height > 1 else 
None.
- explorer/puller/puller/facade/SymbolPuller.py:461 - it's called inside the page loop, so the seed carries across pages and across concurrent fetch batches, and crucially it runs before
_sync_block_batch_with_dirty_state, so a discontinuous batch is never written.

On the NEM side, sync_blocks (explorer/puller/puller/facade/NemPuller.py:1053) has no equivalent:

```
while chain_height > db_height:
blocks = await self._retry_get_blocks_after(db_height)
for block in blocks:
  block_queue.put(block)
db_height = blocks[-1].height
```

Blocks go straight from the queue into insert_block; the only constraint is blocks.height UNIQUE. Nothing compares a fetched block's previousBlockHash against the preceding block, or
against the current DB tip.

The scenario I'm worried about: page 1 ends at height H on branch A and gets committed (the writer commits every 50 blocks). The node reorgs to H-5 and rebuilds on branch B. The next
get_blocks_after(H) returns H+1… from branch B. Heights stay contiguous, so nothing objects, and the database ends up holding a splice of two branches with the discontinuity buried at the
H/H+1 boundary - orphaned transactions, harvested fees, namespaces, mosaics and remote links from branch A all left behind.

The next run doesn't catch it. detect_rollback compares at the tip, H+37 say: the block hash matches the node, and its parent H+36 also matches - both are branch B - so it returns None and
sync continues as if nothing happened. Symbol's _validate_block_chain would have raised a ValueError on the first block of the second page and written nothing.

Jaguar0625: this is a good point. i had viewed this change more as identifying common block during a rollback, and it seems mostly correct for that purpose. but, having equivalent to Symbol of validating chain links during sync / before db insertion would also be valuable. 👍

explorer/puller/tests/facade/test_NemPuller.py
```diff
@@ -1599,41 +1602,153 @@ def test_can_process_transactions_inner_outer(self, mock_process_transaction):
...
+		# Arrange + Act:
+		fork_height = self._run_detect_rollback_test(
+			{1: 'parent', 2: 'common', 3: 'same-current'},
+			{
+				2: NodeBlockHashes('common', 'parent'),
+				3: NodeBlockHashes('SAME-CURRENT', 'different-parent')
+			},
+			3,
+			3,
+			[3, 2, 2, 1],
```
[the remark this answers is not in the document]
AnthonyLaw: When checking block 3’s parent, it queries block 2 from the database. 
The hashes don’t match, so it steps back to block 2.

The loop then queries block 2 again to check the block itself.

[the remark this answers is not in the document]
Jaguar0625: can you not reuse the parent block lookup instead of looking up same block twice?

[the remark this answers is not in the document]
AnthonyLaw: I made an improvement, which cached the parent hash, so it reuses it when needed.

Jaguar0625 (approved):
👍

PR #2526, opened 2026-08-25
[bridge] feat: expose bridge account balances on a /metrics endpoint

cryptoBeliever:
## Problem

We have no monitoring on the bridge. If a bridge account runs out of funds, or a node stops
answering, we find out when payouts stop - or when someone notices by hand.

## Solution

A Prometheus `/metrics` endpoint on the API we already run, reporting the balances of the
bridge accounts on both legs.

| Metric | Labels | Meaning |
|---|---|---|
| `bridge_node_up` | `network`, `endpoint` | the node named in configuration answered |
| `bridge_balance` | `network`, `token`, `address` | bridge account balance of the configured token |
| `bridge_gas_balance` | `network`, `address` | native currency balance available for fees |

`network` is `native` or `wrapped`. Sample output from a live instance against public testnet nodes:

```
bridge_node_up{endpoint="http://libertalia.nemtest.net:7890",network="native"} 1.0
bridge_node_up{endpoint="https://201-sai-dual.symboltest.net:3001",network="wrapped"} 1.0
bridge_balance{address="TAIOKTRMJLDEVAT2BBUDNENR6LS3E6VCZI2FDCTV",network="native",token="nem:xem"} 1e+07
bridge_balance{address="TBQAAM...RDWZ4FCQ",network="wrapped",token="77F94E840313C8EA"} 0.0
bridge_gas_balance{address="TBQAAM...RDWZ4FCQ",network="wrapped"} 7.57821053281e+011
```


bridge/bridge/api/metrics/MetricsLoader.py
```diff
@@ -0,0 +1,7 @@
+from .ChainCollector import ChainCollector
+
+
+def load_collectors(config, context, timeout_seconds):
+	"""Loads every metric collector that applies to a bridge configuration."""
+
+	return [ChainCollector(config, context, timeout_seconds)]
```
Jaguar0625: are you planning on adding more connectors in future?

bridge/bridge/api/metrics/ChainCollector.py
```diff
@@ -0,0 +1,96 @@
...
+
+		try:
+			await self.context.load()
+		except Exception:  # pylint: disable=broad-except
+			# the facades never got built, so the endpoint has to come from configuration instead
+			network_configs = (self.config.native_network, self.config.wrapped_network)
+			return [
+				NetworkReading(role, network_config.endpoint, None, None, None, None)
+				for (role, network_config) in zip(NETWORK_ROLES, network_configs)
+			]
```
Jaguar0625: is there reason you want to gracefully fallback to config here?

config.load failure is fatal. all the other routes that depend on it will return 5xx. i don't see compelling reason for this route to behave differently ?

bridge/bridge/api/metrics/ChainCollector.py
```diff
@@ -0,0 +1,96 @@
...
+	if balance is None:
+		return NetworkReading(role, facade.config.endpoint, address, token, None, None)
+
+	# fees are always paid in the chain's native currency; a mosaic id of None means the bridge moves
+	# that currency itself, so the balance above already covers them and there is nothing separate to report
+	gas_balance = None
+	if facade.extract_mosaic_id().id:
+		gas_balance = await _try_read(facade.read_native_currency_balance(connector))
+
+	return NetworkReading(role, facade.config.endpoint, address, token, balance, gas_balance)
```
Jaguar0625: given usage of NetworkReading i think it would make more sense to have list of (token, balance) tuples or map of {token: balance}. then upstream doesn't need to know about gas_balance.

bridge/bridge/api/metrics/ChainCollector.py
```diff
@@ -0,0 +1,96 @@
...
+# what a single scrape learned about one network; balance is None exactly when the node could not be read
+NetworkReading = namedtuple('NetworkReading', ['role', 'endpoint', 'address', 'token', 'balance', 'gas_balance'])
+
+
+async def _try_read(awaitable):
+	"""Awaits a node read, returning None instead of raising, so one bad node cannot fail the scrape."""
+
+	try:
+		return await awaitable
+	except Exception:  # pylint: disable=broad-except
```
Jaguar0625: i think this is a bit too broad. if you want to catch *everything* from a node use NodeException, but i think for this purpose, HttpException would be sufficient.

bridge/bridge/api/metrics/ChainCollector.py
```diff
@@ -0,0 +1,96 @@
...
+			'bridge_node_up',
+			'node configured for the bridge is reachable',
+			['network', 'endpoint'],
+			registry=registry)
+		balance_gauge = Gauge('bridge_balance', 'bridge account balance', ['network', 'token', 'address'], registry=registry)
+		gas_gauge = Gauge(
+			'bridge_gas_balance',
+			'native currency balance available for fees',
+			['network', 'address'],
+			registry=registry)
```
Jaguar0625: is there reason you want two separate gauges? i think you only need single (with varying token).

right now native currencies can get logged to `bridge_balance` sometimes and `bridge_gas_balance` other times, which seems undesirable.

bridge/bridge/api/metrics/MetricsCollector.py
```diff
@@ -0,0 +1,32 @@
...
+
+		registry = CollectorRegistry()
+		await asyncio.gather(*[self._collect_safely(collector, registry) for collector in self._collectors])
+		return registry
+
+	async def _collect_safely(self, collector, registry):
+		try:
+			await collector.collect(registry)
+		except Exception:  # pylint: disable=broad-except
+			self._logger.exception('collector %s failed', type(collector).__name__)
```
Jaguar0625: unless there is good reason otherwise, i'd also remove this error handling. since it should be getting handled by collect(). or if you want to log it, reraise the error.

bridge/tests/test/MockEthereumServer.py
```diff
@@ -64,6 +67,10 @@ async def rpc_main(self, request):  # pylint: disable = too-many-return-statemen
 
 			raise ValueError(f'unknown ETH RPC method: {method}')
 
+		async def _handle_eth_get_balance(self, request, account_address):
+			balance = address_to_balance_map.get(account_address, 0) if address_to_balance_map else 0
+			return await self._process(request, {'result': hex(balance)})
```
Jaguar0625: please keep helper functions in same order as branches, so after _handle_eth_gas_price

bridge/bridge/ethereum/EthereumConnector.py
```diff
@@ -144,6 +144,20 @@ async def nonce(self, account_address, block_identifier='latest'):
 
 	# endregion
 
+	# region native_balance
+
+	async def native_balance(self, account_address, block_identifier='latest'):
```
Jaguar0625: since this is identical to nonce aside from function name, create helper function that just takes difference (function name) that both call

bridge/tests/ethereum/test_EthereumConnector.py
```diff
@@ -159,6 +162,48 @@ async def test_can_query_balance_with_custom_block_identifier(server):  # pylint
 # endregion
 
 
+# region native_balance
```
Jaguar0625: please match source ordering (so this should be after nonce)

bridge/tests/ethereum/test_EthereumConnector.py
```diff
@@ -159,6 +162,48 @@ async def test_can_query_balance_with_custom_block_identifier(server):  # pylint
...
+	await _assert_can_query_native_balance(server, None, 'latest')
+
+
+async def test_can_query_native_balance_with_custom_block_identifier(server):  # pylint: disable=redefined-outer-name
+	await _assert_can_query_native_balance(server, 0xAABBCC, '0xAABBCC')
+
+
+async def test_native_balance_of_unknown_account_is_zero(server):  # pylint: disable=redefined-outer-name
+	# Act:
+	balance = await EthereumConnector(server.make_url('')).native_balance(
```
Jaguar0625: correct, but i'd suggest creating connector separately in Arrange

bridge/bridge/ethereum/EthereumNetworkFacade.py
```diff
@@ -54,6 +54,21 @@ def extract_mosaic_id(self):
...
+		if not mosaic_id.id:
+			# an empty mosaic id means the bridge moves ETH itself rather than an ERC-20
+			return await connector.native_balance(self.bridge_address)
+
+		return await connector.balance(self.bridge_address, mosaic_id.id)
+
+	async def read_native_currency_balance(self, connector):
+		"""Reads the balance of the chain's native currency, which is what transaction fees are paid in."""
+
+		return await connector.native_balance(self.bridge_address)
```
Jaguar0625: since read_bridge_balance can already read network currency, i don't think this function is needed as it doesn't seem to provide anything new.

bridge/tests/nem/test_NemNetworkFacade.py
```diff
@@ -514,3 +515,40 @@ async def test_cannot_calculate_transfer_transaction_fee_without_mosaic_fee_info
 		facade.calculate_transfer_transaction_fee(_create_sample_balance_transfer('', amount=88887_000), ('foo', 'baz'))
 
 # endregion
+
+
+# region read_bridge_balance, read_native_currency_balance
+
+async def test_can_read_bridge_balance_of_native_currency(server):  # pylint: disable=redefined-outer-name
```
Jaguar0625: would prefer you keep ordering of tests across different facades consistent i.e.
(read native, read token) or (read token, read native)

bridge/tests/symbol/test_SymbolNetworkFacade.py
```diff
@@ -422,3 +426,31 @@ def test_can_calculate_transfer_transaction_fee_with_message():
...
+	# Arrange:
+	facade = SymbolNetworkFacade(_create_config(server, 'FAF0EBED913FA202'))
+	await facade.init()
+
+	# Act:
+	balance = await facade.read_bridge_balance(facade.create_connector())
+
+	# Assert:
+	assert 778899 == balance
+
```
Jaguar0625: you seem to be missing read_bridge_balance variant that reads currency balance

bridge/tests/api/metrics/test_MetricsCollector.py
```diff
@@ -0,0 +1,76 @@
...
+def _patch_collectors(monkeypatch, collectors):
+	monkeypatch.setattr('bridge.api.metrics.MetricsCollector.load_collectors', lambda _config, _context, _timeout_seconds: collectors)
+
+
+async def test_metrics_of_every_collector_land_in_one_registry(monkeypatch):
+	# Arrange:
+	_patch_collectors(monkeypatch, [_ProbeCollector('bridge_probe_one', 1), _ProbeCollector('bridge_probe_two', 2)])
+
+	# Act:
+	registry = await MetricsCollector(None, None).collect()
```
Jaguar0625: it feels more natural to pass collectors into MetricsCollector, so app would call load_collectors and pass down result

bridge/tests/api/metrics/test_MetricsCollector.py
```diff
@@ -0,0 +1,76 @@
...
+
+	# Assert:
+	assert 1 == registry.get_sample_value('bridge_probe_one')
+	assert 2 == registry.get_sample_value('bridge_probe_two')
+
+
+async def test_each_scrape_starts_from_an_empty_registry(monkeypatch):
```
Jaguar0625: i think naming is slightly misleading as i think each scrape _overwrites_ not resets 

i.e. it doesn't recreate gauges, just updates values

Jaguar0625 (commented):
👍

